PR #1, opened 2014-12-19
Unescape the less than and greater than inequalities

peterjc:
This change seems sensible for both reading the input markdown, and the rendering as HTML.

Assuming there is not something in the template which deals with this (and my change is not needed), then I would guess this fix will be needed on all the other recently migrated lessons (and the migration script may need tweaking).

Update: Looking at a few of the other lessons with R and Python (by eye), I did not find any more unescaped inequalities.

### DIFF
--- a/01-numpy.html
+++ b/01-numpy.html
@@ -28,7 +28,7 @@
           <h1 class="title">Programming with Python</h1>
           <h2 class="subtitle">Analyzing Patient Data</h2>
 <div id="learning-objectives" class="objectives objectives">
-<h2>Learning Objectives</h2>
+<h2 id="learning-objectives" class="objectives objectives">Learning Objectives</h2>
 <ul>
 <li>Explain what a library is, and what libraries are used for.</li>
 <li>Load a Python library and use the things it contains.</li>
@@ -43,7 +43,7 @@
 <p>In order to load our inflammation data, we need to <strong>import</strong> a library called NumPy. In general you should use this library if you want to do fancy things with numbers, especially if you have matrices. We can load NumPy using:</p>
 <pre class="sourceCode python"><code class="sourceCode python"><span class="ch">import</span> numpy</code></pre>
 <p>Importing a library is like getting a piece of lab equipment out of a storage locker and setting it up on the bench. Once it's done, we can ask the library to read our data file for us:</p>
-<pre class="sourceCode python"><code class="sourceCode python">numpy.loadtxt(fname=&amp;<span class="co">#39;inflammation-01.csv&amp;#39;, delimiter=&amp;#39;,&amp;#39;)</span></code></pre>
+<pre class="sourceCode python"><code class="sourceCode python">numpy.loadtxt(fname=<span class="st">&#39;inflammation-01.csv&#39;</span>, delimiter=<span class="st">&#39;,&#39;</span>)</code></pre>
 <pre class="output"><code>array([[ 0.,  0.,  1., ...,  3.,  0.,  0.],
        [ 0.,  1.,  2., ...,  1.,  0.,  1.],
        [ 0.,  1.,  1., ...,  2.,  1.,  1.],
--- a/01-numpy.html
+++ b/01-numpy.html
@@ -93,7 +93,7 @@
  [ 0.  0.  1. ...,  1.  1.  0.]]</code></pre>
 <p>Now that our data is in memory, we can start doing things with it. First, let's ask what <strong>type</strong> of thing <code>data</code> refers to:</p>
 <pre class="sourceCode python"><code class="sourceCode python"><span class="dt">print</span> <span class="dt">type</span>(data)</code></pre>
-<pre class="output"><code>&amp;lt;type &#39;numpy.ndarray&#39;&amp;gt;</code></pre>
+<pre class="output"><code>&lt;type &#39;numpy.ndarray&#39;&gt;</code></pre>
 <p>The output tells us that <code>data</code> currently refers to an N-dimensional array created by the NumPy library. We can see what its <strong>shape</strong> is like this:</p>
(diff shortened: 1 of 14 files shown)


gvwilson:
I haven't had a chance to review - could you please?

abostroem:
@gvwilson not sure how to test this. When I run 
make preview 
it only updates the README.html...

peterjc:
You can look at the README.md file in your favourite markdown aware viewer, or just browse the files as of my commit right here on GitHub using GitHub's rendering, e.g.:
https://github.com/peterjc/python-novice-inflammation/blob/a4009e55195018f3a1e5b04e2fcc7f3b278170de/01-numpy.md

aaren:
@peterjc you missed one:

```
$ ack '&#' -G .*md
01-numpy.md
39:numpy.loadtxt(fname=&#39;inflammation-01.csv&#39;, delimiter=&#39;,&#39;)
```

aaren:
@abostroem you can force make to update everything by doing `make preview -B`

aaren:
@gvwilson any idea why this is like this in the first place?

gvwilson:
@aaren Apologies for losing the thread - why what is like what?

peterjc:
@gvwilson I think the question was how did these over-escaped HTML entities get into the markdown source?

@aaren Good catch on the HTML escaped quotes, give my a moment and I'll fix that in another commit...

aaren:
@gvwilson yes, why are these html things (`&#39;`, etc.) in the markdown at all? Is there a systemic reason?

peterjc:
Rebased and the quote fixed.

peterjc:
I haven't down a grep across all the lessons (I only have a handful cloned locally), but it looks like something went wrong on this lesson only?

aaren:
Here is the full (?) list of html entities still in the markdown:

```
$ ack '&' -G '.*md$'
05-defensive.md
71:<ipython-input-19-33d87ea29ae4> in <module&gt;()

04-cond.md
412:    That means that on a 40&times;60 data set, we are performing the same calculation 2400 times.

01-numpy.md
39:numpy.loadtxt(fname=&#39;inflammation-01.csv&#39;, delimiter=&#39;,&#39;)
238:if we have an M&times;N array in Python,
468:The expression `(40,)` tells us we have an N&times;1 vector,

```

@peterjc can you fix the defensive programming one as well?

As for the others (`&times`), I'm not sure what to do with these. If we had mathjax we could replace with latex, but I don't think that we do so either leave them as is or replace with `x`.

peterjc:
I thought I did all the less-than and greater-than signs, thanks again @aaren.

I'm not sure what is best for the times signs - they should look nice in the HTML output at least.

gvwilson:
It's almost certainly a side effect of an automatic translator (probably 
the IPython Notebook to Markdown translation).  There's no reason to 
keep the escapes.  I'm happy to replace them all with "natural" characters.

01-numpy.html
```diff
@@ -28,7 +28,7 @@
           <h1 class="title">Programming with Python</h1>
           <h2 class="subtitle">Analyzing Patient Data</h2>
 <div id="learning-objectives" class="objectives objectives">
-<h2>Learning Objectives</h2>
+<h2 id="learning-objectives" class="objectives objectives">Learning Objectives</h2>
```
aaren: Does this need to change with this PR? I know it is happening automatically, but could you move it to a different commit?

05-defensive.html
```diff
@@ -28,7 +28,7 @@
           <h1 class="title">Programming with Python</h1>
           <h2 class="subtitle">Defensive Programming</h2>
 <div id="learning-objectives" class="objectives objectives">
-<h2>Learning Objectives</h2>
+<h2 id="learning-objectives" class="objectives objectives">Learning Objectives</h2>
```
aaren: also here.

abostroem:
@aaren and @peterjc does the current form render correctly and this just fixes the code to be all html or does the current version (pre-PR) render incorrectly?

Please let me know when you are ready to merge

aaren:
@abostroem the pre-PR version renders incorrectly. Look at `01-numpy.html` in your browser. This PR modifies the markdown input to not use html escaped characters and then updates the html to reflect that.

peterjc:
I deliberately committed in the changes to ``*.html`` as generated by ``make preview`` as per my understanding of the instructions.

(Personally I don't understand why generated ``*.html`` files are under source code control, but that's a different issue.)

gvwilson:
> I deliberately committed in the changes to |*.html| as generated by 
> |make preview| as per my understanding of the instructions.
>

Thanks.

> (Personally I don't understand why generated |*.html| files are under 
> source code control, but that's a different issue.)
>
From https://github.com/swcarpentry/lesson-template/blob/gh-pages/DESIGN.md:

> Using Pandoc instead of Jekyll means that we have to compile our 
> Markdown into HTML on our own machines and commit it to the |gh-pages| 
> branch of the lesson's GitHub repository.
>
> It's considered bad practice to put computer-generated files under 
> version control, but the HTML pages put into the lesson's root 
> directory by Pandoc /must/ be committed to version control in order 
> for the lesson to be displayed properly on GitHub.
>
> It's also considered bad practice to put generated files in the same 
> directory as source files. We do it because some authors may use 
> auxiliary source files such as R Markdown and IPython Notebook files 
> which are executable, and contain paths to things like images and data 
> files. If we put our source files in a sub-directory, those paths 
> won't work (or conversely, if we put our generated files in a 
> different directory, the paths won't work there).
>

-- 
Dr. Greg Wilson    | gvwilson@software-carpentry.org
Software Carpentry | http://software-carpentry.org

aaren:
This looks good to me. I'll leave it to @abostroem to merge.

PR #3, opened 2014-12-28
challenge titles, glossary entries, grid images

PBarmby:
Adding glossary entries and FIXME challenge titles.


03-loop.md
```diff
@@ -362,7 +362,7 @@ Sure enough,
 the maxima of these data sets show exactly the same ramp as the first,
 and their minima show the same staircase structure.
 
-> FIXME {.challenge}
+> From 1 to N {.challenge}
```
tbekolay: I think these challenge titles should be prepended with `##`, like they are in 02-func.md?

tbekolay:
This looks great! The challenge titles in `03-loop.md` might need to be made into `<h2>`s, but that can be done in another PR. Merging!

03-loop.md
gvwilson: Yes, challenge titles should be level-2 headings.  The validation tool 
should catch this...

gvwilson:
I'd prefer them to be fixed pre-merge if it's not too late.

tbekolay:
Fair enough! Holding off for now.

I also noticed that `README.md` was rendered out into an HTML document, which I don't think it should be, so @PBarmby if you could remove that from this branch that'd be great also.

PBarmby:
Think it's good to go now.

PR #7, opened 2015-01-19
point all the image src URIs to use the fig/ path

zmughal:


abostroem:
I'd like a second opinion on this from someone more familiar than I am with how the websites build and the history of this project- why were we pointing to src and is fig the right place to be pointing.
@gvwilson ?

wking:
On Mon, Jan 19, 2015 at 11:22:52AM -0800, Azalee Bostroem wrote:
> … is fig the right place to be pointing.

swcarpentry/lesson-template#43 may help you here.  I'm not sure why
it's currently pointing somewhere else, but I haven't looked into the
history at all.

r-gaia-cs:
This looks good. +1 to merge.

> why were we pointing to src

Because this is how HTML works, check http://www.w3.org/TR/html5/embedded-content-0.html#the-img-element. Sorry if I don't have a better answer.

@zmughal Thanks for this pull request. Do you mind replace the HTML with Markdown? E.g. replacing

~~~
<img src="fig/python-sticky-note-variables-01.svg" alt="Variables as Sticky Notes" />
~~~

with

~~~
![Variables as Sticky Notes](fig/python-sticky-note-variables-01.svg)
~~~

If you do that, you can update this pull request using

~~~
$ git add -u
$ git commit --amend
$ git push -f origin fix/fig-src
~~~

where `origin` is the name of your local repository.

zmughal:
@r-gaia-cs Yeah, I can do that right now.

zmughal:
Done! I added a couple captions where they were missing too. I can remove them if they distract from the flow of the text.

The HTML will need to be rebuilt.

01-numpy.md
```diff
@@ -500,7 +500,7 @@ pyplot.imshow(data)
 pyplot.show()
 ~~~
 
-<img src="../../novice/python/01-numpy_files/novice/python/01-numpy_71_0.png">
+![Heatmap of the Data](fig/01-numpy_71_0.png)
```
zmughal: added caption here

01-numpy.md
```diff
@@ -513,7 +513,7 @@ pyplot.plot(ave_inflammation)
 pyplot.show()
 ~~~
 
-<img src="../../novice/python/01-numpy_files/novice/python/01-numpy_73_0.png">
+![Average Inflammation Over Time](fig/01-numpy_73_0.png)
```
zmughal: added caption here

01-numpy.md
```diff
@@ -529,14 +529,14 @@ pyplot.plot(data.max(axis=0))
 pyplot.show()
 ~~~
 
-<img src="../../novice/python/01-numpy_files/novice/python/01-numpy_75_1.png">
+![Maximum Value Along The First Axis](fig/01-numpy_75_1.png)
```
zmughal: added caption here

01-numpy.md
```diff
@@ -529,14 +529,14 @@ pyplot.plot(data.max(axis=0))
...
 
 ~~~ {.python}
 pyplot.plot(data.min(axis=0))
 pyplot.show()
 ~~~
 
-<img src="../../novice/python/01-numpy_files/novice/python/01-numpy_75_3.png">
+![Minimum Value Along The First Axis](fig/01-numpy_75_3.png)
```
zmughal: added caption here

01-numpy.md
```diff
@@ -572,7 +572,7 @@ plt.tight_layout()
 plt.show()
 ~~~
 
-<img src="../../novice/python/01-numpy_files/novice/python/01-numpy_80_0.png">
+![The Previous Plots as Subplots](fig/01-numpy_80_0.png)
```
zmughal: added caption here

03-loop.md
```diff
@@ -46,7 +46,7 @@ def analyze(filename):
 analyze('inflammation-01.csv')
 ~~~
 
-<img src="../../novice/python/03-loop_files/novice/python/03-loop_2_0.png">
+![](fig/03-loop_2_0.png)
```
zmughal: These do not have captions since they are explained in the text directly.

r-gaia-cs:
> Done! I added a couple captions where they were missing too.

@zmughal Thanks.

> I can remove them if they distract from the flow of the text.

What you call "caption" is the alternative text that is render if the web browser doesn't support images or the image can't be retrieved. They don't distract from the flow of the text.

> The HTML will need to be rebuilt.

Could you do that?

~~~
$ make preview
~~~

Sorry to forget it in the first time. After you rebuild the HTML could you add it to this pull request using the same commands from my last comments?

01-numpy.html
```diff
@@ -68,17 +68,23 @@ <h2 id="learning-objectives" class="objectives objectives">Learning Objectives</
 <pre class="output"><code>weight in kilograms is now: 57.5</code></pre>
 <p>As the example above shows, we can print several things at once by separating them with commas.</p>
 <p>If we imagine the variable as a sticky note with a name written on it, assignment is like putting the sticky note on a particular value:</p>
-<p><img src="img/python-sticky-note-variables-01.svg" alt="Variables as Sticky Notes" /></p>
+<div class="figure">
+<img src="fig/python-sticky-note-variables-01.svg" alt="Variables as Sticky Notes" /><p class="caption">Variables as Sticky Notes</p>
```
zmughal: @r-gaia-cs, this is what I mean by the caption. It adds a `<p class="caption">` to the HTML in addition to the `alt` attribute.

zmughal:
Done!

wking:
On Mon, Jan 19, 2015 at 04:09:57PM -0800, Raniere Silva wrote:
> > The HTML will need to be rebuilt.
> 
> Could you do that?
> 
> ~~~
> $ make preview
> ~~~
> 
> Sorry to forget it in the first time. After you rebuild the HTML
> could you add it to this pull request using the same commands from
> my last comments?

I'd recommend the compiled HTML land as a separate commit, to avoid
distracting from the human-generated changes.  To avoid churn due to
installed tools, it makes sense to me to have the maintainers product
that commit with their local toolset.  But I'm not a maintainer here,
so that's just my two cents ;).  Handle it however you like.

01-numpy.html
r-gaia-cs: Sorry that I didn't understand it correctly.

This is due one of the Pandoc's default extensions to Markdown. If you want we can disable this extension

~~~
$ pandoc -t html -f markdown-implicit_figures<<EOF 
![foo](bar.jpg) 
EOF
<p><img src="bar.jpg" alt="foo" /></p>
~~~

or do it for each case when need:

~~~
$ pandoc -t html <<EOF
![foo](bar.jpg)\  
EOF
<p><img src="bar.jpg" alt="foo" /> </p>
~~~

There is one white space right after the `\`.

gvwilson: Is there a way to automatically number figures?  If so, I'm +1 on 
titling them (but only if the number is part of the title, e.g., "Figure 
3: Thorium Analysis").  If they can't be numbered, or if the number 
cannot be included in the title, I'm -1 on titling.

zmughal: Looks like it can't be done with pandoc yet. <https://github.com/jgm/pandoc/pull/1298>

wking: On Mon, Jan 19, 2015 at 04:43:05PM -0800, Greg Wilson wrote:
> Is there a way to automatically number figures?

Maybe not yet?  jgm/pandoc#615

It looks like there's an implementation in jgm/pandoc#1298 that's
blocking on a cleaner solution [1].

[1]: https://github.com/jgm/pandoc/pull/1298#issuecomment-52658301

r-gaia-cs: > Is there a way to automatically number figures?

We can use CSS.

~~~
body {
  counter-reset: figure;
}
div.figure p::before {
  content: "Figure " counter(figure) ": ";
  counter-increment: figure;
}
~~~

Another possible solution is using another filter based on

~~~
$ pandoc -t json <<EOF
![foo](bar.jpg) 
EOF
[{"unMeta":{}},[{"t":"Para","c":[{"t":"Image","c":[[{"t":"Str","c":"foo"}],["bar.jpg","fig:"]]}]}]]
$ pandoc -t json -f markdown-implicit_figures<<EOF
![foo](bar.jpg)
EOF
[{"unMeta":{}},[{"t":"Para","c":[{"t":"Image","c":[[{"t":"Str","c":"foo"}],["bar.jpg",""]]}]}]]
~~~

zmughal:
I backed out the HTML in the commit for now and used the `<backslash><white-space>` escaping on some of the figures so that all images will at least have an `alt` attribute. Let me know if that works out.

r-gaia-cs:
@zmughal Sorry for the delay.

> I (...) used the `<backslash><white-space>` escaping on some of the figures

You only used it on the figures at `03-loop.md`. This is correctly?

zmughal:
@r-gaia-cs , yes, this is correct. I can do so on the remaining figures if you'd like. In fact, maybe just make those changes and decide how to enumerate the figures later in another issue?

(The reason I only did that on just the ones in `03-loop.md` is because the caption didn't add anything that the surrounding text didn't explain already.)

r-gaia-cs:
> I can do so on the remaining figures if you'd like.

@zmughal Please. When you update this pull request, please send me a note (can be on IRC or email to avoid spam) and I will merge this.

> In fact, maybe just make those changes and decide how to enumerate the figures later in another issue?

:+1:

PR #23, opened 2015-02-01
Post-UBC Refactor

BillMills:
After studying the Python material closely for a few weeks and teaching it at UBC, I have a few ideas for these lessons. Apologies for the large PR; a couple of issues are tied up together.

## Things I'd like to address:
 - **Length.** There is a lot of great content in this lesson, but at the speed novices can follow it at, it could fill an entire day; we got through the first 3 sections in a morning at UBC, even after omitting parts of them. The four most useful things I would like learners to go home with after this lesson are:
  - Numpy & matplotlib 
  - Loops
  - Conditionals
  - Functions

  If these are our goals in a reverse instructional design sense, there are a number of secondary things that can be put off, to the end of the lesson if there's time, or more likely to a second or intermediate Python lesson.

 - **Complexity.** The cognitive load is quite heavy in some places, particularly with the half hour (or more) of ipython blocks material before getting to conditionals, and in detours through details like default parameter values and aliasing. Again, pulling the key concepts forward will alleviate this.

 - **Ordering.** The ordering of the first four sections (what I would hope to get through in one novice session) could use a second look. As @gvwilson pointed out, functions would be better motivated if they came a bit later when the example analysis script starts getting unwieldy, and as @ttimbers observed at UBC, prioritizing getting to conditionals is good if they don't see them in the shell - comments on my approach here to these discussions very welcome.

## Changes in this PR:
 - **The first four sections are now in the order:**
    - **Numpy demo**: essentially unchanged.
    - **Loops**: motivated by wanting to draw the plots from the numpy demo for all 12 inflammation datasets.
    - **Conditionals**: after looping in the last lesson, a couple of different features in the data are revealed. Instead of ipython blocks material, write a loop that checks each file for these features, and generates different messages depending on what it finds in the data. This removes the cognitive load of the ipython blocks material, and connects to the story about inflammation better.
    - **Functions**: now that we're doing a number of different things in our analysis script, show how they can be packaged up in functions to make things easier to read, debug and reuse.

 - **Move some material later**: material on the call stack is too abstract for most novices; the ipython blocks content is also well written, but not as useful as other things we could be focusing on. These sections were very well written, so I've moved them off to a supplemental lesson at the end, for instructors to incorporate as they see fit. Also, teaching aliasing in a novice lesson is a missed opportunity to let learners get used to the brand new things we just showed them, by changing the syntax almost immediately; let's leave this for another day.

 - **Modify inflammation data**: in order to create a couple of different things to look for in the data, `inflammation-03.csv`, `-08.csv` and `-11.csv` now have different structure: random maxima per day, but a zero minimum per day (apparently an uninflamed person snuck into those studies) - this way, the conditionals lesson can check for a few different things in an if / elif / else construct. 

Thanks, and comments, requests and alternate solutions always welcome.


01-numpy.md
```diff
@@ -544,38 +544,34 @@ Neither result seems particularly likely,
 so either there's a mistake in our calculations
 or something is wrong with our data.
 
-It's very common to create an **alias** for a library when importing it
-in order to reduce the amount of typing we have to do.
-Here are our three plots side by side using aliases for `numpy` and `pyplot`:
+Here are our three plots side by side:
```
tbekolay: I appreciate that the concept of making aliases is a bit unnecessary and unrelated to other pedagogical goals at this point, but it's not a point that has to be understood deeply (or even has any real deep meaning); it's enough to say that you can make shortcuts to save on typing. Perhaps we should just import with an alias from the very beginning, rather than waiting until now to do it?

Briefly, my justification for wanting to keep `np` and `plt` in general is that this is the standard way to write numpy / matplotlib code, which means our code will look like what students see when they google for answers to their specific questions. Saving typing is a non-trivial gain as well, as we don't want to make students type so much that they resort to `%pylab inline` or `from numpy import *`, which is sometimes recommended for the purpose of saving typing.

02-loop.md
```diff
@@ -337,76 +302,86 @@ filenames = glob.glob('*.csv')
...
+> Write a loop that takes a string,
+> and produces a new string with the characters in reverse order,
+> so `'Newton'` becomes `'notweN'`.
 
-> FIXME {.challenge}
-> 
-> Write a function called `total` that calculates the sum of the values in a list.
+> ## FIXME {.challenge}
+>
+> Write a loop calculates the sum of the values in a list.
```
tbekolay: "loop that calculates"

03-cond.md
```diff
@@ -0,0 +1,213 @@
...
+
+Now that we've seen how conditionals work,
+we can use them to check for the suspicious features we saw in our inflammation data.
+In the first couple of plots, the maximum inflammation per day
+seemed to rise like a straight line, one unit per day.
+We can check for this inside the `for` loop we wrote with the following conditional:
+
+~~~ {.python}
+if data.max(axis=0)[0] == 0 and data.max(axis=0)[20] == 20:
+  print 'Suspicious looking maxima!'
```
tbekolay: Please indent with 4 spaces for consistency

03-cond.md
```diff
@@ -0,0 +1,213 @@
...
+~~~
+
+We also saw a different problem in the third dataset;
+the minima per day were all zero (looks like a healthy person snuck into our study).
+We can also check for this with an `elif` condition:
+
+~~~{.python}
+elif data.min(axis=0).sum() == 0:
+  print 'Minima add up to zero!'
```
tbekolay: Please indent with 4 spaces for consistency

03-cond.md
```diff
@@ -0,0 +1,213 @@
...
+~~~
+
+And if neither of these conditions are true, we can use `else` to give the all-clear:
+
+~~~ {.python}
+else:
+  print 'Seems OK!'
```
tbekolay: Please indent with 4 spaces for consistency

04-func.md
```diff
@@ -217,138 +223,68 @@ Real-life functions will usually be larger than the ones shown here --- typicall
...
-Why go to all this trouble?
-Well,
-here's a function called `span` that calculates the difference between
-the mininum and maximum values in an array:
+and another function called `detectProblems` that checks for those systematics
+we noticed:
 
 ~~~ {.python}
-import numpy
+def detectProblems(filename):
```
tbekolay: Please indent with 4 spaces for consistency

04-func.md
```diff
@@ -217,138 +223,68 @@ Real-life functions will usually be larger than the ones shown here --- typicall
...
-Notice that `span` assigns a value to a variable called `diff`.
-We might very well use a variable with the same name to hold data:
+Notice that rather than jumbling this code together in one giant `for` loop,
+we can now read and reuse both ideas separately.
+We can reproduce the previous analysis with a much simpler `for` loop:
 
 ~~~ {.python}
-diff = numpy.loadtxt(fname='inflammation-01.csv', delimiter=',')
-print 'span of data:', span(diff)
+for f in filenames:
```
tbekolay: Please indent with 4 spaces for consistency

tbekolay:
I'm a big fan of these changes! I usually skip the call stack and IPython blocks material in the interest of time, so moving it to a supplementary section makes a lot of sense to me. Introducing functions is always much more difficult than it seems, so moving it later also seems like a big improvement.

The only thing I'm not 100% on board with is getting rid of the `np` and `plt` aliases. I agree that the place that we introduce them is somewhat arbitrary, and that should change, but once the students are let loose into the real world, I would want them used to reading and writing code like `import numpy as np; np.arange(4)` and not `import numpy; numpy.arange(4)` or `from numpy import *; arange(4)`.

Also, a minor nitpick that I pointed out in a few places: I think all of our examples use 4-spaces for indentation. We should keep this consistent.

Let me know what you think @BillMills! The sooner we get this merged in the better!

gvwilson:
> The only thing I'm not 100% on board with is getting rid of the |np| 
> and |plt| aliases. I agree that the place that we introduce them is 
> somewhat arbitrary, and that should change, but once the students are 
> let loose into the real world, I would want them used to reading and 
> writing code like |import numpy as np; np.arange(4)| and not |import 
> numpy; numpy.arange(4)| or |from numpy import *; arange(4)|.
>

How would you feel about doing it the long way first, then having a 
call-out box saying "this will be easier to read if..." and then doing 
it the short (aliased) way thereafter.

> Let me know what you think @BillMills <https://github.com/BillMills>! 
> The sooner we get this merged in the better!
+1 from me as well.

tbekolay:
> How would you feel about doing it the long way first, then having a
> call-out box saying "this will be easier to read if..." and then doing
> it the short (aliased) way thereafter.

That's essentially what we do now, minus the call-out box (we just introduce it in a paragraph at the end of the lesson). I'm fine with doing it this way, but this PR introduces a change such that the aliased versions are never used, which I'm not sold on yet (but could be persuaded).

gvwilson:
I believe the aliased versions should be used.

BillMills:
Okay, sounds good - I'll put the aliasing material back in right now, and make all the other smaller corrections you suggest. Update inbound shortly, thanks for having a look!

tbekolay:
Great :rainbow: :smile_cat:

BillMills:
Okay I think that's got it per all your requests - thanks!

01-numpy.md
```diff
@@ -558,24 +558,22 @@ plt.figure(figsize=(10.0, 3.0))
...
 
 plt.tight_layout()
 plt.show()
 ~~~
 
-![The Previous Plots as Subplots](fig/01-numpy_80_0.png)\ 
+![The Previous Plots as Subplots](fig/01-numpy_80_0.png)\
 
-The first two lines re-load our libraries as `np` and `plt`,
-which are the aliases most Python programmers use.
```
kynan: Wasn't the suggestion to make this *more prominent* rather than remove it?

kynan:
Great set of changes. Agree with moving the call stack and blocks to a supplementary section.

Even with these changes I think 30min per lesson are ambitious.

01-numpy.md
tbekolay: The aliases are described above; I think it's okay to remove this text.

PR #30, opened 2015-02-12
Run whole command line lesson from command line

DamienIrving:
This was demo in instructor training - more changes to come


DamienIrving:
@abostroem Thanks for merging that. I created this PR as a demonstration in front of the class during the recent instructor training course in Melbourne. There are a couple of other places in the lesson that need to be updated so it is clear that the entire lesson is run from the command line - I'll submit a new PR in the coming weeks with those minor changes.

PR #31, opened 2015-02-12
Adjusted phrase to clarify the learning objective

ndjones:
The phrasing was slightly confusing, as the second part might have referred to the program, rather than the assertions.

### DIFF
--- a/05-defensive.md
+++ b/05-defensive.md
@@ -7,7 +7,7 @@ minutes: 30
 > ## Learning Objectives {.objectives}
 >
 > *   Explain what an assertion is.
-> *   Add assertions to programs that correctly check the program's state.
+> *   Add assertions to programs, which correctly check the program's state.
 > *   Correctly add precondition and postcondition assertions to functions.
 > *   Explain what test-driven development is, and use it when creating new functions.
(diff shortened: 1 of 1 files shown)


kynan:
Makes sense to me.

ewanbarr:
This appears to be grammatically incorrect (mix of singular and plural and an incorrect use of the work "which") and should really just read:

Add assertions to a program that checks the program's state is correct.

ndjones:
Agree .. the further refinement of plural to singular is required, good catch - I hadn't noticed that.

The problem with the suggested phrasing is that the "that" refers to the noun that precedes it, in this case the word program. Changing it to ", which" does make the qualifier at the end non-restrictive so perhaps changes the intended meaning .. though it then reads correctly as simply a clause to provide further information on the first clause.

Perhaps either of the following would work..
- Add assertions to a program, which check the program's state is correct.
- Add assertions that check the program's state is correct.

ewanbarr:
Either of those work for me.

abostroem:
@ndjones Please update this with either statement so we can merge this PR

kynan:
I think this is good to go now.

PR #32, opened 2015-02-12
Make the challenge slightly more domain specific

ianedwardthomas:
Changed variables in examples to illustrate why age and weight might suddenly change.  Semantically the same as before, but at least a *little* more compelling.    If the astronaut travels at FTL, someone might try to work out what would happen to the age in that case :)


kynan:
Makes sense to me.

PR #33, opened 2015-02-12
Added second example to exponentiation

rhosking:


kynan:
Not quite sure what this demonstrates?

rhosking:
While only a small contribution, repetition helps, showing / typing another example could help generalise the idea.

I know I had to google the definition of exponentiation to double check I understood the notion.


On 13/02/2015, at 10:02 am, Florian Rathgeber <notifications@github.com<mailto:notifications@github.com>> wrote:


Not quite sure what this demonstrates?

abostroem:
I'm not convinced this is necessary - other comments?

timClicks:
Maybe we add another other example with `float`or possibly `decimal.Decimal` to show that `**` works with other date types?

kynan:
It doesn't add anything as is, but using a different data type as suggested by @timClicks would be useful.

gvwilson:
Thanks to @rhosking for the PR - I think exponentiation will be clear to our learners without the second example.

PR #35, opened 2015-02-12
Update 02-loop.md

raffademichelis:
may not be obvious to everyone that loops in python don't have any "END" keyword

### DIFF
--- a/02-loop.md
+++ b/02-loop.md
@@ -111,7 +111,8 @@ for variable in collection:
 
 We can call the **loop variable** anything we like,
 but there must be a colon at the end of the line starting the loop,
-and we must indent the body of the loop.
+and we must indent the body of the loop. Unlike many other codes, there is no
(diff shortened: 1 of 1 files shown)


02-loop.md
```diff
@@ -111,7 +111,8 @@ for variable in collection:
 
 We can call the **loop variable** anything we like,
 but there must be a colon at the end of the line starting the loop,
-and we must indent the body of the loop.
+and we must indent the body of the loop. Unlike many other codes, there is no
```
kynan: programming languages rather than codes

02-loop.md
```diff
@@ -111,7 +111,8 @@ for variable in collection:
...
+command to end a loop (e.g. end for); what is indented after the for line belongs to the loop.
```
kynan: for line -> for statement

02-loop.md
aussieraf: yes, agree

PR #39, opened 2015-02-12
Change plotting examples to use Python rather than MATLAB style

lonsbio:
Existing example plots use MATLAB style. Updated to use a more Python like style (i.e store the plots and axes in variables) but keeping as close as possible to exisiting code so the challengws etc should be unaffected.

### DIFF
--- a/01-numpy.md
+++ b/01-numpy.md
@@ -554,22 +554,23 @@ from matplotlib import pyplot as plt
 
 data = np.loadtxt(fname='inflammation-01.csv', delimiter=',')
 
-plt.figure(figsize=(10.0, 3.0))
+fig = plt.figure(figsize=(10.0, 3.0))
 
-plt.subplot(1, 3, 1)
-plt.ylabel('average')
-plt.plot(data.mean(axis=0))
+axes1 = fig.add_subplot(1, 3, 1)
+axes2 = fig.add_subplot(1, 3, 2)
+axes3 = fig.add_subplot(1, 3, 3)
 
-plt.subplot(1, 3, 2)
-plt.ylabel('max')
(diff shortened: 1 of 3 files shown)


abostroem:
I shy away from introducing object oriented plotting in the novice lesson. I find that if you are not used to object oriented programming the syntax can be difficult and that plotting lessons take a really long time to teach. In the interest of time I usually differ OO syntax to a separate lesson on plotting (if there is time). I'd love to hear other opinions.

jduckles:
We're already using OO syntax in the Numpy stuff at the very beginning. I personally think the syntax reads better with this PR and is a much better example of Python usage, so I'm :+1:  but I'm willing to hear other arguments like @abostroem.

lonsbio:
Thanks for the feedback. I'm closing the request since I noticed an Agg error with fig.show() on a clean IPython notebook - will resubmit with a fix that's more consistent with the plot introductions.

lonsbio:
Have updated so that plt.show(image) style that is consistent in the changes (which fixes a backend error too!).

gvwilson:
I'm gonna let @jiffyclub make the call on this one... :-)

jiffyclub:
I am 100% for this because these are best practices for working with matplotlib and I don't think the added complexity is all that bad. For plotting people are kind of in "trust us and type in what we show you" mode anyway, we might as well show them the right way. 

("Stop using `pyplot` functions and use `Figure`/`Axes` methods" is the first thing I tell people about matplotlib. The global state that happens with the `pyplot` interface leads to so many problems.)

gvwilson:
If @jiffyclub likes it, I like it.

PR #41, opened 2015-02-13
Created a new question for the python introduction function section.

theboocock:
Added a question in a slightly new format. Would really like to know if questions like this are useful.

Motivation: Just a suggestion to make the question more relatable.
Goal: What we will have or solved after completing all the smaller more manageable tasks.
Tasks: Set of challenges, which if completed should solve the goal.

In this case, I tried to combine the conditionals, loops,  and functions in a hopefully understandable way. Maybe
this is more of a mastery question.

### DIFF
--- a/04-func.md
+++ b/04-func.md
@@ -720,7 +720,6 @@ but *do* have to provide `delimiter=` for the second parameter.
 > Run the commands `help(numpy.arange)` and `help(numpy.linspace)`
 > to see how to use these functions to generate regularly-spaced values,
 > then use those values to test your `rescale` function.
-
 > ## Variables inside and outside functions {.challenge}
 >
 > What does the following piece of code display when run - and why?
@@ -739,3 +738,30 @@ but *do* have to provide `delimiter=` for the second parameter.
 >
 > print k
 > ~~~
+
+> ## Combining functions to use both loops, conditionals {.challenge}
+>
+> Motivation:
+> Imagine you have just returned from a week long holiday where you also totally blew your petrol budget. You realise it would be useful to know the
+> number of days where you went over your daily petrol usage allowance of $100. However, as you only recorded the miles travelled per day, and you know the
+> efficiency of your vehicle in litres/km, a few calculations are going to be needed. These new fangled python skills are going to come in handy.   
+>
+> Overall Goal:
+> Given a Python list of miles travelled per day, a measurement of litres per kilometre, and the price of fuel, calculate the number of days where you exceeded $100
+> 
+> ~~~ {.python}
+> trips = [80,50,170,180,80,50,200]
+> ~~~
+> Tasks:
+> 
+> 1. Use the above list of miles travelled per day in iPython and create a function that converts miles
+> to kilometres. 
+>
+> kilometres = miles / 1.61
+>
+> 2. Use a loop to display the results of calling the miles to kilometres function on each item in the trips list.
+> 
+> 3. Create another function that converts kilometres to the cost of that trip, given that the fuel efficiency of the car is 0.5 litres per kliometres, and fuel
+>     costs $2 a litre.  
+>
+> 4. Using a loop, a conditional, and the two functions created above, calculate the total number of days in which your total petrol bill exceeded $100. 
--- a/04-func.md
+++ b/04-func.md
@@ -739,7 +739,7 @@ but *do* have to provide `delimiter=` for the second parameter.
 > print k
 > ~~~
 
-> ## Combining functions to use both loops and conditionals {.challenge}
+> ## Combining functions that use both loops and conditionals to create a workflow {.challenge}
 >
 > Overall Goal:
 > Given a list of miles travelled per day, a measurement of litres per kilometre, and the price of fuel, calculate the number of days where you exceeded your daily petrol budget of $100.
--- a/04-func.md
+++ b/04-func.md
@@ -739,15 +739,10 @@ but *do* have to provide `delimiter=` for the second parameter.
 > print k
 > ~~~
 
-> ## Combining functions to use both loops, conditionals {.challenge}
->
-> Motivation:
-> Imagine you have just returned from a week long holiday where you also totally blew your petrol budget. You realise it would be useful to know the
-> number of days where you went over your daily petrol usage allowance of $100. However, as you only recorded the miles travelled per day, and you know the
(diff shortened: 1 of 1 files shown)


04-func.md
```diff
@@ -739,3 +738,30 @@ but *do* have to provide `delimiter=` for the second parameter.
 >
 > print k
 > ~~~
+
+> ## Combining functions to use both loops, conditionals {.challenge}
```
r-gaia-cs: I prefer "and" instead of ",":

~~~
> ## Combining functions to use both loops and conditionals {.challenge}
~~~

04-func.md
```diff
@@ -739,3 +738,30 @@ but *do* have to provide `delimiter=` for the second parameter.
...
+>
+> Overall Goal:
+> Given a Python list of miles travelled per day, a measurement of litres per kilometre, and the price of fuel, calculate the number of days where you exceeded $100
+> 
+> ~~~ {.python}
+> trips = [80,50,170,180,80,50,200]
+> ~~~
+> Tasks:
+> 
+> 1. Use the above list of miles travelled per day in iPython and create a function that converts miles
```
r-gaia-cs: What about reword as

~~~
1.  Create a function that receive one argument,
    the above list of miles traveled per day,
    and return a new list with the values converted to kilometers.
~~~

04-func.md
```diff
@@ -739,3 +738,30 @@ but *do* have to provide `delimiter=` for the second parameter.
...
+> to kilometres. 
+>
+> kilometres = miles / 1.61
+>
+> 2. Use a loop to display the results of calling the miles to kilometres function on each item in the trips list.
```
r-gaia-cs: This item will be cover by my suggestion of reword the first task.

04-func.md
```diff
@@ -739,3 +738,30 @@ but *do* have to provide `delimiter=` for the second parameter.
...
+> 
+> 3. Create another function that converts kilometres to the cost of that trip, given that the fuel efficiency of the car is 0.5 litres per kliometres, and fuel
```
r-gaia-cs: What about reword as

~~~
2.  Create another function that
    calculate the cost of that trip for each day
    and receive as arguments
    the list of kilometers traveled by day,
    the fuel efficiency of the car in liters per kilometers
    and the fuel cost per liter.

    To test your use 0.5 for the fuel efficiency and US$2 for the cost of fuel.
~~~

04-func.md
```diff
@@ -739,3 +738,30 @@ but *do* have to provide `delimiter=` for the second parameter.
...
+>     costs $2 a litre.  
+>
+> 4. Using a loop, a conditional, and the two functions created above, calculate the total number of days in which your total petrol bill exceeded $100.
```
r-gaia-cs: What about reword as

~~~
3.  Create another function that
    return the total number of days in which you total petrol bill exceeded your budget
    and receive as arguments
    the list of cost of trip per day and your budget.

    To test your function use US$100 as your budget.
~~~

04-func.md
```diff
@@ -739,3 +738,30 @@ but *do* have to provide `delimiter=` for the second parameter.
 >
 > print k
 > ~~~
+
+> ## Combining functions to use both loops, conditionals {.challenge}
+>
+> Motivation:
```
r-gaia-cs: This can be moved to `instructors.md`.

04-func.md
```diff
@@ -739,3 +738,30 @@ but *do* have to provide `delimiter=` for the second parameter.
...
+> Imagine you have just returned from a week long holiday where you also totally blew your petrol budget. You realise it would be useful to know the
+> number of days where you went over your daily petrol usage allowance of $100. However, as you only recorded the miles travelled per day, and you know the
+> efficiency of your vehicle in litres/km, a few calculations are going to be needed. These new fangled python skills are going to come in handy.   
+>
+> Overall Goal:
```
r-gaia-cs: This can be moved to `instructors.md`.

r-gaia-cs:
@smilefreak Thanks for this pull request. I like the challenge but when I'm teaching I always try to enforce that students apply good programming practices like creating functions in a reusable way, pipe functions to address one task, ... so I suggest one possible reword of the exercises. The motivation is great but I think that it belongs at `instructors.md`.

smilefreak:
@r-gaia-cs Great suggestions!
 I am updating the pull-request to incorporate them. However, I am having trouble locating ```instructors.md``` it is not located in the repo and on the website the link goes no where.
Do you have any suggestion of how to progress on this.

smilefreak:
@r-gaia-cs  I have updated the pull request to emphasise better practice, and removed the motivation as instructors should be able to ad-lib that.

04-func.md
```diff
@@ -739,3 +738,34 @@ but *do* have to provide `delimiter=` for the second parameter.
...
+> This function will recieve the following as arguments: the list of kilometres traveled per day, 
+> the fuel effieciency of the car in litres per kilometer, and the fuel cost per litre. The function should return a list that contains values
+> which represent the cost of the trip per day.
+>
+> Test you function using a fuel efficiency value of 0.5, and $2 for the cost of a litre of fuel.
+>
+> 3. Create another function that returns the total number of days in which 
+>    your petrol bill exceeded your daily budget. The function will recieve
+>    as arguments, a list with values representing the cost of your trip per day.
+> , and your budget.
```
kynan: A variant could be to return the days at which the budget is exceeded.

abostroem:
I like this example. How long do you anticipate it taking students (have you taught a similar example before)? 

Can you change all instances of petrol to fuel. In the US we don't use that term - it will make the lesson more universal.

kynan:
I think this is pretty good. The first function is way simpler than the 2nd and 3rd, so you could arguably fold it into the 2nd one. On the other hand it's a "quick win" to easy people into the challenge.

gvwilson:
Can/should this be broken into several challenges in order, rather than one large one?  And while it seems silly, can the justification be made more science-y?

smilefreak:
Good point, ill try make it more science-y. Could do a chemistry based one or something, those guys use volumes and prices. 

Yea could separate into a few challenges, would just like to keep them nearby on the website. Would want to do this because I think that having a theme of problems to solve, will help to drive home the practicality of a rather abstract concept.

tbekolay:
Since this PR hasn't progressed in over a year, I'm going to close it. I'm adding the revise-and-resubmit label because this PR would be welcome in this lesson, but not in its current state. If the PR is updated then feel free to reopen it! Or submit a new PR.

PR #42, opened 2015-02-15
Challenges in line with lesson + backslash fix

thomas-coudrat:
Challenges in line:
In order to prompt regular challenges along the teaching of this topic, I moved each challenge to directly follow the section it evaluates. This in line formatting of challenges was also pointed out in another pull request (https://github.com/swcarpentry/python-novice-inflammation/pull/40) by @ewanbarr. I believe it follows the layout guide for topics (https://github.com/swcarpentry/lesson-template/blob/gh-pages/LAYOUT.md#topics): this guide does not explicitly mention that the challenges should all be located at the end of the a topics' page.

Challenges merge:
I have also merged two challenges together and kept their common header "Check your understanding".

Layout style:
Additionally, the {.objectives}, {.challenge} and {.callout} styles have not been used in the current public page for this topic (http://swcarpentry.github.io/python-novice-inflammation/01-numpy.html). Using these styles will help highlight the various challenge  breaks (green boxes) throughout the topic.

Backslash fix:
There was a problem in the markdown to html conversion: the python code snippet between the min() and max() images was not displayed properly. (Find this string "~ {.python} pyplot.plot(data.min(axis=0)) pyplot.show() ~" on the page http://swcarpentry.github.io/python-novice-inflammation/01-numpy.html). This was due to a backslash used at the end of the image embedding markdown line. The backslash is an escape character in markdown, it was escaping '~' thus not displaying the code snipped properly. I removed all backslash characters in the image embedding lines as they did not seem to have a purpose (I could be wrong!).

### DIFF
--- a/01-numpy.md
+++ b/01-numpy.md
@@ -124,7 +124,7 @@ we can print several things at once by separating them with commas.
 If we imagine the variable as a sticky note with a name written on it,
 assignment is like putting the sticky note on a particular value:
 
-![Variables as Sticky Notes](fig/python-sticky-note-variables-01.svg)\
+![Variables as Sticky Notes](fig/python-sticky-note-variables-01.svg)
 
 This means that assigning a value to one variable does *not* change the values of other variables.
 For example,
@@ -138,7 +138,7 @@ print 'weight in kilograms:', weight_kg, 'and in pounds:', weight_lb
 weight in kilograms: 57.5 and in pounds: 126.5
 ~~~
 
-![Creating Another Variable](fig/python-sticky-note-variables-02.svg)\
+![Creating Another Variable](fig/python-sticky-note-variables-02.svg)
 
 and then change `weight_kg`:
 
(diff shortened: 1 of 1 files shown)


abostroem:
This is a great idea! I just merged a fix to the example of mass and age changing. Can you please incorporate that change and then we're ready to merge.

kynan:
Or we just cherry-pick this patch on top of current `gh-pages`.

kynan:
Done: 8d7a49f

thomas-coudrat:
Hi there,
My apologies for a this question as I am fairly new to this: do I need to do anything now? If I understand correctly, my changes were accepted that did not interfere with @abostroem 's fix merge.

kynan:
Nope, your changes are in. I basically did in one go what would have happened if you had merged `gh-pages` into your branch and then we had merged this pr (in effect, but the history is different).

If you like you can still do the merge on your branch as an exercise and then diff against 8d7a49f and see if you get the same result :)

PR #47, opened 2015-02-23
Add subsection on nose and unittest

vahtras:
This is a tranfser of suggested improvements that I did for the instructors course in the older lesson format. It introduces the nosetests tool and the unittest module in a new section. The main point I wish to make is to write test functions that do not have more than one assertion, and to encourage use of the good tools for testing that comes with Python.

### DIFF
--- a/05-defensive.md
+++ b/05-defensive.md
@@ -418,6 +418,167 @@ This violates another important rule of programming:
 "[always initialize from data](../../rules.html#always-initialize-from-data)".
 We'll leave it as an exercise to fix `range_overlap`.
 
+### Using a testing framework
+
+#### nose
+
+In the previous section we had a single test function that does different
+assertions on input data. This makes it easier to run many tests but there are
+better ways, and Python provides them for free. The recommended practise for
+test functions is to test one and one thing only. Let us split up the function
+and give each function a descriptive name. Good function names give a better
+overview of what is being tested, and it is easier to spot the corner cases we may
+have forgot about,  that should have been tested.
+
+~~~{.python}
+def test_nonoverlapping_returns_None():
+    assert range_overlap([ (0.0, 1.0), (5.0, 6.0) ]) == None
+
+def test_adjacent_intervals_returns_None():
+    assert range_overlap([ (0.0, 1.0), (1.0, 2.0) ]) == None
+
+def test_single_interval_returns_self():
+    assert range_overlap([ (0.0, 1.0) ]) == (0.0, 1.0)
+
+def test_two_partially_overlapping_returns_intersection():
+    assert range_overlap([ (2.0, 3.0), (2.0, 4.0) ]) == (2.0, 3.0)
+
+def test_three_partially_overlapping_returns_intersection():
+    assert range_overlap([ (0.0, 1.0), (0.0, 2.0), (-1.0, 1.0) ]) == (0.0, 1.0)
+~~~
+
+The third-party testing package nose (to be installed separately) detects modules and functions beginning with ``test``. If the function and tests are collected in a single file we execute the program ``nosetests`` from the shell
+
+~~~{.bash}
+$ nosetests
+~~~
+~~~{.error}
+FF.F.
+======================================================================
+FAIL: test_range_overlap.test_nonoverlapping_returns_None
+----------------------------------------------------------------------
+Traceback (most recent call last):
+  File "/usr/lib/python2.7/dist-packages/nose/case.py", line 197, in runTest
+    self.test(*self.arg)
+  File "/home/olav/courses/python-novice-inflammation/test_range_overlap.py", line 11, in test_nonoverlapping_returns_None
+    assert range_overlap([ (0.0, 1.0), (5.0, 6.0) ]) == None
+AssertionError
+
+======================================================================
+FAIL: test_range_overlap.test_adjacent_intervals_returns_None
+----------------------------------------------------------------------
+Traceback (most recent call last):
+  File "/usr/lib/python2.7/dist-packages/nose/case.py", line 197, in runTest
+    self.test(*self.arg)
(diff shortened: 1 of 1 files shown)


05-defensive.md
```diff
@@ -418,6 +418,167 @@ This violates another important rule of programming:
...
+assertions on input data. This makes it easier to run many tests but there are
+better ways, and Python provides them for free. The recommended practise for
+test functions is to test one and one thing only. Let us split up the function
+and give each function a descriptive name. Good function names give a better
+overview of what is being tested, and it is easier to spot the corner cases we may
+have forgot about,  that should have been tested.
+
+~~~{.python}
+def test_nonoverlapping_returns_None():
+    assert range_overlap([ (0.0, 1.0), (5.0, 6.0) ]) == None
```
kynan: Should this be using `is` instead of `==`?

05-defensive.md
```diff
@@ -418,6 +418,167 @@ This violates another important rule of programming:
...
+
+----------------------------------------------------------------------
+Ran 5 tests in 0.002s
+
+FAILED (failures=3)
+~~~
+
+Now all tests are run even if one fails, and we get a better picture of what
+is working and what is not. The first line of output contains a character for
+each test, a dot for pass and an ``R`` for fail. For the tests that fail further output is given. The goal is to have an output with a single line of dots!
```
kynan: Could also break this line.

05-defensive.md
```diff
@@ -418,6 +418,167 @@ This violates another important rule of programming:
...
+    assert range_overlap([ (0.0, 1.0) ]) == (0.0, 1.0)
+
+def test_two_partially_overlapping_returns_intersection():
+    assert range_overlap([ (2.0, 3.0), (2.0, 4.0) ]) == (2.0, 3.0)
+
+def test_three_partially_overlapping_returns_intersection():
+    assert range_overlap([ (0.0, 1.0), (0.0, 2.0), (-1.0, 1.0) ]) == (0.0, 1.0)
+~~~
+
+The third-party testing package nose (to be installed separately) detects modules and functions beginning with ``test``. If the function and tests are collected in a single file we execute the program ``nosetests`` from the shell
```
kynan: I think you want to use single ` instead of ``?

05-defensive.md
```diff
@@ -418,6 +418,167 @@ This violates another important rule of programming:
...
+is working and what is not. The first line of output contains a character for
+each test, a dot for pass and an ``R`` for fail. For the tests that fail further output is given. The goal is to have an output with a single line of dots!
+
+
+#### The unittest module
+
+Another option to testing is the ``unittest`` library. This leads to slightly
+more complex test code but it is part of the standard library and is available
+wherever Python is installed - there is not need to install additional
+programs.  To use the unittest module  requires a little familiarity with
```
kynan: programs -> packages

kynan:
I think this is pretty good modulo some minor niggles pointed out.

05-defensive.md
drj11: Opinions vary. Is this case I think you could reasonably argue that `== None` is correct because `range_overlap` is at liberty to return any value that is `== None` and not just the unique `None` object.

However, having said all that, most Python programmers would probably expect to see `is None` and would be surprised if `range_overlap` did in fact return an object that was `== None` but not `is None`.

abostroem:
I've been moving away from teaching testing in the novice python section and instead adding a 1 hour unit on code review (which I think could be its own repository). I think this should either be in supplementary.md or in the intermediate python. Opinions?

gvwilson:
It would be *great* to have this in the main lesson - is it something 
you can write up (both "is it write-uppable?" and "do you have time to 
do so?")
Thanks,
Greg

jiffyclub:
I think this lesson is definitely beyond the python-novice level, mainly because of the inclusion of subclasses in the section on `unittest`. And in general I believe we've been moving toward taking testing completely off the standard introductory curriculum.

If we include a section on testing with frameworks, either here or in python-intermediate, I think it should be its own lesson because of the number of topics required:

- rules of test discovery
- project structure for proper testing
- mentioning both nose and pytest (I'd leave unittest out of the picture)
- fixtures?
- comparing numbers?
- probably more stuff...

05-defensive.md
vahtras: ``is None`` is fine too, but as I copied these lines from the previous section we should have the same throughout the whole lesson

05-defensive.md
vahtras: Thanks for pointing that out

vahtras:
I can agree that the section on ``unittest`` module can be left out or handled at a more intermediate level. The first part is not that advanced though, mainly a split of the test function in the original text to several separate test functions for each test case. I feel strongly that testing  should not be left out completely at  the novice level. Comments?

jiffyclub:
I also would like to see testing somehow included, but see this blog post
and the ensuing discussion:
http://software-carpentry.org/blog/2014/10/why-we-dont-teach-testing.html

tbekolay:
I agree with what others have said here; let's drop the `unittest` section and keep the rest. Individual instructors can choose to spend more or less time on it, as desired. @vahtras, do you want to update this PR accordingly? If you're busy, I can make the updates instead.

vahtras:
I will have another review and update the PR

-Olav

claireverity:
An easier intro to tests might be doctest.testmod(). This would also tie in nicely with encouraging good docstrings and descriptive comments.

katrinleinweber:
Since I didn't find another mention of [Katy Huff's "Python Testing and Continuous Integration" lesson](https://katyhuff.github.io/python-testing/) in this lesson, I'll just put it here in case anyone reads until here, wondering whether or how pytest or unittest are used further.

Issue #48, opened 2015-02-28
Improve consistency of teaching flow


rgaiacs:
@moorepants and @zonca [reported](http://software-carpentry.org/blog/2015/02/instructor-debriefing-2015-02-24.html) that

> We tried the 5.3 Beta Python and it was a little rough around the edges. (...) it has big issues, like functions being introduced in lesson 4 but used in 2.

abostroem:
@r-gaia-cs are the other issues listed somewhere? Can you either point me to them or elaborate?

r-gaia-cs:
> are the other issues listed somewhere? Can you either point me to them or elaborate?

@moorepants and @zonca could you elaborate the others issues that you find? Thanks in advance.

zonca:
thank you @r-gaia-cs , @abostroem, other issue are mainly formatting:

* output in defining default: http://swcarpentry.github.io/python-novice-inflammation/06-func.html
* some FIXME at the end of: http://swcarpentry.github.io/python-novice-inflammation/03-lists.html

PR #49, opened 2015-02-28
Remove func from loop

abostroem:
In response to #48 removed functions from loop lesson. I'd appreciate an independent review to make sure this seems reasonable without the functions.


kynan:
Looks reasonable to me.

jdblischak:
+1

PR #56, opened 2015-03-12
Fixed small typo in lists topic

dotsdl:
I fixed a small typo in the lists topic, and also regenerated the HTML files from the lessons. The updates to the HTML files came from changes that weren't mine, but they're updated anyhow.

This is my first pull request to the lesson repository, mainly to fulfill the instructor training requirement. I realize it's not very substantial.


tbekolay:
Every little bit helps ;) Thanks!

PR #59, opened 2015-04-05
Moving towards compliance with template.

gvwilson:
1.  Moving material from `10-supplementary` into `discussion`.
2.  Adding styles.
3.  Making names consistent between links and files.


kynan:
What's the rationale for moving supplementary material to discussion?

The rest is uncontroversial.

tbekolay:
I'll second @kynan's comment; everything looks good to me, but I'm wondering if there's a precedent for putting additional material in the discussion section. I wonder if we should do what happened for the shell, and make a python-extras repo for this material instead?

kynan:
I think the supplementary material as it is now doesn't warrant a dedicated repo. How do we intend this material to be used? Not for discussion I would think but rather as optional content that can be used if appropriate.

abostroem:
I'll merge this later tonight. @gvwilson any final comments on the material moved to the discussion section?

gvwilson:
No final thoughts from me.

PR #61, opened 2015-04-10
Drop 'import ... as ...' syntax

wking:
It's not worth introducing extra syntax just because folks might see
this in the wild.  As Greg said in 181e4635 (Starting to write
instructor's guide, 2015-02-15):

> In order to teach that, we must teach people a little about the
> mechanics of manipulating data with lists and file I/O so that their
> functions can do things they actually care about.  Our teaching
> order tries to show practical uses of every idea as soon as it is
> introduced; instructors should resist the temptation to explain the
> "other 90%" of the language as well.

### DIFF
--- a/01-numpy.md
+++ b/01-numpy.md
@@ -599,17 +599,16 @@ Neither result seems particularly likely,
 so either there's a mistake in our calculations
 or something is wrong with our data.
 
-It's very common to create an [alias](reference.html#alias) for a library when importing it
-in order to reduce the amount of typing we have to do.
-Here are our three plots side by side using aliases for `numpy` and `pyplot`:
+You can groups similar plots in a single figure using subplots.
+Here are our three plots side by side:
 
 ~~~ {.python}
-import numpy as np
-from matplotlib import pyplot as plt
+import numpy
+import matplotlib.pyplot
 
-data = np.loadtxt(fname='inflammation-01.csv', delimiter=',')
+data = numpy.loadtxt(fname='inflammation-01.csv', delimiter=',')
 
-fig = plt.figure(figsize=(10.0, 3.0))
+fig = matplotlib.pyplot.figure(figsize=(10.0, 3.0))
 
 axes1 = fig.add_subplot(1, 3, 1)
 axes2 = fig.add_subplot(1, 3, 2)
@@ -626,7 +625,7 @@ axes3.plot(data.min(axis=0))
 
 fig.tight_layout()
 
-plt.show(fig)
+matplotlib.pyplot.show(fig)
 ~~~
 
 ![The Previous Plots as Subplots](fig/01-numpy_80_0.png)
--- a/08-cmdline.md
+++ b/08-cmdline.md
@@ -135,12 +135,12 @@ though we can call it whatever we want:
 
 ~~~ {.python}
 import sys
-import numpy as np
+import numpy
 
 def main():
     script = sys.argv[0]
     filename = sys.argv[1]
-    data = np.loadtxt(filename, delimiter=',')
+    data = numpy.loadtxt(filename, delimiter=',')
     for m in data.mean(axis=1):
         print m
 ~~~
@@ -160,12 +160,12 @@ Let's add a call to `main`:
 
 ~~~ {.python}
 import sys
-import numpy as np
+import numpy
 
 def main():
     script = sys.argv[0]
     filename = sys.argv[1]
-    data = np.loadtxt(filename, delimiter=',')
+    data = numpy.loadtxt(filename, delimiter=',')
     for m in data.mean(axis=1):
         print m
 
@@ -307,12 +307,12 @@ Here's our changed program
 
 ~~~ {.python}
 import sys
-import numpy as np
+import numpy
 
 def main():
     script = sys.argv[0]
     for filename in sys.argv[1:]:
-        data = np.loadtxt(filename, delimiter=',')
+        data = numpy.loadtxt(filename, delimiter=',')
         for m in data.mean(axis=1):
             print m
 
@@ -352,7 +352,7 @@ so we could just do this:
 
 ~~~ {.python}
 import sys
-import numpy as np
+import numpy
 
 def main():
     script = sys.argv[0]
@@ -360,7 +360,7 @@ def main():
     filenames = sys.argv[2:]
 
     for f in filenames:
-        data = np.loadtxt(f, delimiter=',')
+        data = numpy.loadtxt(f, delimiter=',')
 
         if action == '--min':
             values = data.min(axis=1)
@@ -402,7 +402,7 @@ so that the program fails fast:
 
 ~~~ {.python}
 import sys
-import numpy as np
+import numpy
 
 def main():
     script = sys.argv[0]
@@ -414,7 +414,7 @@ def main():
         process(f, action)
 
 def process(filename, action):
-    data = np.loadtxt(filename, delimiter=',')
+    data = numpy.loadtxt(filename, delimiter=',')
 
     if action == '--min':
         values = data.min(axis=1)
--- a/code/readings-01.py
+++ b/code/readings-01.py
@@ -1,9 +1,9 @@
 import sys
-import numpy as np
+import numpy
 
 def main():
     script = sys.argv[0]
     filename = sys.argv[1]
-    data = np.loadtxt(filename, delimiter=',')
+    data = numpy.loadtxt(filename, delimiter=',')
     for m in data.mean(axis=1):
         print m
--- a/code/readings-02.py
+++ b/code/readings-02.py
@@ -1,10 +1,10 @@
 import sys
-import numpy as np
+import numpy
 
 def main():
     script = sys.argv[0]
     filename = sys.argv[1]
-    data = np.loadtxt(filename, delimiter=',')
+    data = numpy.loadtxt(filename, delimiter=',')
     for m in data.mean(axis=1):
         print m
 
--- a/code/readings-03.py
+++ b/code/readings-03.py
@@ -1,10 +1,10 @@
 import sys
-import numpy as np
+import numpy
 
 def main():
     script = sys.argv[0]
     for filename in sys.argv[1:]:
-        data = np.loadtxt(filename, delimiter=',')
+        data = numpy.loadtxt(filename, delimiter=',')
         for m in data.mean(axis=1):
             print m
 
--- a/code/readings-04.py
+++ b/code/readings-04.py
@@ -1,5 +1,5 @@
 import sys
-import numpy as np
+import numpy
 
 def main():
     script = sys.argv[0]
@@ -7,7 +7,7 @@ def main():
     filenames = sys.argv[2:]
 
     for f in filenames:
-        data = np.loadtxt(f, delimiter=',')
+        data = numpy.loadtxt(f, delimiter=',')
 
         if action == '--min':
             values = data.min(axis=1)
--- a/code/readings-05.py
+++ b/code/readings-05.py
@@ -1,5 +1,5 @@
 import sys
-import numpy as np
+import numpy
 
 def main():
     script = sys.argv[0]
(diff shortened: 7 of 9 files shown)


gvwilson:
Mmm... the other reason for introducing the aliases is that instructors and learners are typing this stuff as they go along, and some of them won't have editors that tab-complete.

r-gaia-cs:
Since this only about Python
and I expect that people use IPython Notebook
**that has tab-complete**
I'm +1 to drop `import ... as ...`
**for our novice lesson**.

gvwilson:
A large number of instructors *don't* use the Notebook (at least not 
initially).

wking:
On Fri, Apr 10, 2015 at 04:26:57AM -0700, Greg Wilson wrote:
> A large number of instructors *don't* use the Notebook (at least not
> initially).

Do they use IPython's command line interface?  Or do they switch
between an editor and a terminal for running scripts?

wking:
On Fri, Apr 10, 2015 at 02:59:22AM -0700, Greg Wilson wrote:
> Mmm... the other reason for introducing the aliases is that
> instructors and learners are typing this stuff as they go along, and
> some of them won't have editors that tab-complete.

And this commit only requires:

* Six ‘np.’ → ‘numpy.’ edits
* Two ‘plt.’ → ‘matplotlib.pyplot.’ edits

in the core topics (not counting the code/readings-*.py, since they're
readings ;).  And five of the ‘np.’ → ‘numpy.’ edits are in
08-cmdline.md, where folks are most likely editing a single
cell/script and won't need to retype ‘numpy.’ at every place it shows
up in the notes.

abostroem:
I agree with @gvwilson  on this one. When we tab complete what we are doing is invisible to our students

wking:
On Fri, Apr 10, 2015 at 09:13:16AM -0700, Azalee Bostroem wrote:
> I agree with @gvwilson on this one. When we tab complete what we are
> doing is invisible to our students

And when we introduce new syntax we're loading our learner's short
term memory ;).  I'm not saying there are no negative effects to this
change, I'm just saying that I think the benefit of covering only a
single import syntax outweighs the cost of typing out ‘numpy.’ twice
and ‘matplotlib.pyplot.’ once.  There's definitely room for discussion
in there, though.  Do you think the extra typing is a large cost?  Is
explaining tab-completion a large cost (we'd already covered that in
the shell lesson by the time we hit this content at UW)?  Do you think
covering the different posible syntaxes for imports is a small cost?

r-gaia-cs:
> Is explaining tab-completion a large cost?

The problem isn't explaining tab-completion
but that students aren't used to it.

BernhardKonrad:
I think a handful of additional characters is a really good deal for not having to introduce a concept we would rather avoid having to talk about in the first exposure to Python.
How about whoever teaches next tries this version and reports back if the students did, in fact, have troubles typing the longer version?

tbekolay:
This issue seems controversial and with good arguments for either side. I think the only fair way to settle this is with a good old fashioned vote. Register your +1 for either A) dropping import aliases or B) keeping import aliases. We'll give it several days, or until there's an obvious consensus.

I'm +1 for keeping them.

wking:
On Mon, Apr 20, 2015 at 06:33:30AM -0700, Trevor Bekolay wrote:
> I'm +1 for keeping them.

I'm +1 for dropping them (obviously ;).

To help me understand your preference for dropping them, do you think
there's something I'm missing in my cost/benefit list [1]?  Or do you
just disagree with my weighting of the various terms?  Both are valid
options (I'm certainly not omniscient, and my weightings are
subjective), but so far I haven't seen any suggestions for
alternatives.

[1]: https://github.com/swcarpentry/python-novice-inflammation/pull/61#issuecomment-91622898

tbekolay:
Just the weighting; I think the "folks will see this in the wild" and "my fingers hurt from typing" outweigh the short term memory demand. Many novices that I have taught are slow at editing text; they type at a leisurely pace, and when they make a mistake at the start of a line, they may not realize it for a while. Once they realize it, they'll backspace back to where they made the error, and start typing from that point. Not that big a deal for `numpy`, but `matplotlib.pyplot` is super easy to confuse ("matplotlib" has a lot of hard consonants that sound similar when said quickly). I don't think novices are really thinking that deeply about the syntax in class either; they're primarily copying out what we present to them, and will inspect them deeply later. Making a mistake in typing can set them back twenty seconds or so, which we have to wait for; if we continue pressing forward and students fall behind, they quickly get discouraged.

gvwilson:
+1 for keeping.

wking:
On Mon, Apr 20, 2015 at 09:56:21AM -0700, Trevor Bekolay wrote:
> Many novices that I have taught are slow at editing text…

Agreed.  More typing isn't free.

> … but `matplotlib.pyplot` is super easy to confuse …

Also agreed, but with this change they only have to type it out twice
more.  If that takes them a minute or two more, it's still less time
than we spent at UW explaining the ‘import … as …’ syntax and the
magical namespacing you get out of the ‘%pylab inline’ magic we used
(I prefer ‘%matplotlib inline’ [1] to avoid flooding the local
namespace).  It looks like the official lessons dropped the inline
call back in ee835927 (Extracting grids and reformatting lesson,
2014-12-03), so I'm not sure how figures are displayed now, or how
much time folks are taking to explain imports with the current lesson.
But I'd be surprised if you can add the extra syntax in much under the
minute or two of typing you're saving ;).

As an aside on the inline business, UW was mostly taught from an older
version of this lesson.  Now that we've dropped the inline magic, how
are figures displayed?  Is everyone launching an external viewer?

[1]: http://nbviewer.ipython.org/github/ipython/ipython/blob/1.x/examples/notebooks/Part%203%20-%20Plotting%20with%20Matplotlib.ipynb#matplotlib-mode

jttkim:
I vote +1 for dropping "import as".

I think that `import numpy as np` sets a problematic example in the respect that `np` is (or appears like) an ad-hoc identifier and I'd argue that the practice of using reasonably readable variable names should also be applied to import aliases. Using a few terse aliases is convenient initially but can become a problematic habit in more extensive "real life" work (e.g. merging / harmonising / reusing code where different aliases have been used for the same package is definitely not a nice job). From a didactic perspective, possibly having different students using variations of aliases (e.g. `import numpy as np`, `import numpy as npy`, `import numpy as num`) can well confuse students as well as instructors.

On a related issue, I'd also favour `import matplotlib.pyplot` to `from matplotlib import pyplot`, regardless of whether it's followed by an `as` clause, in order to minimise the amount of syntax variations.

Personally I think that `import ... as` should be used sparingly (if at all), and `from ... import` even more so. While I have sympathy for the idea of designing materials to be realistic by incorporating language elements "encountered in the wild", I think the best examples to show are not just commonly encountered, but also examples of good practice -- and with import aliasing and `from ... import` I'm not convinced the latter is the case.

abostroem:
I'm +1 for using the following:

import numpy as np
from matplotlib import pyplot 
(optional as plt)
%matplotlib
(optional in line)

These are all very common syntaxes in on line documentation and I think exposure to it is important for our students

BernhardKonrad:
I agree that it is important to show the `import ... as.` syntax *at some point* in the workshop. But I don't think it should be the first python command that they see. Hence I vote for not using `import ... as` in novice-inflammation.
I do vote for introducing `import ... as` in the second python lecture / day 2.
But since this vote is only about novice-inflammation, again, my +1 is for dropping the syntax.

I would also like to reiterate that it would be great if we found a way for a more data-driven discussion (i.e. experiment with either model), instead of only relying on (very valuable, but nevertheless) anecdotal evidence and personal opinions.

iglpdc:
I agree with @abostroem (`import ... as` is common in documentation, almost the default, see [1]) and with @tbekolay (novices tend to type it wrong). In my opinion, the first is more important than the second.

The problem with tab completion is that no matter how many times you say they should use it, some people don't do it and prefer trust their typing. I guess non-programmers never have to use tab, so it's a really strange key for them.

[1] http://www.scipy.org/getting-started.html

wking:
On Mon, Apr 20, 2015 at 12:13:33PM -0700, Bernhard Konrad wrote:
> I do vote for introducing `import ... as` in the second python
> lecture / day 2.

What lesson do you teach for the second Python lecture?

wking:
On Mon, Apr 20, 2015 at 12:16:33PM -0700, Ivan Gonzalez wrote:
> (`import ... as` is common in documentation, almost the default, see [1])

I think this is just so frozen into the numpy community that it's hard
for them to transition out of it ;).  But I hardly ever see ‘import …
as …’ outside of the numpy ecosystem.  And going back to @gvwilson's
initial “other 90%” rule (quoted in the initial comment here), I
personally think getting to practical use cases with as little syntax
discussion as possible is more important than covering “here's
something you might see after the workshop” code.  There's just not
enough time to go down those rabbit holes (which is why I'd also like
to avoid the ‘from … import …’ syntax).

wking:
On Mon, Apr 20, 2015 at 12:13:33PM -0700, Bernhard Konrad wrote:
> I would also like to reiterate that it would be great if we found a
> way for a more data-driven discussion (i.e. experiment with either
> model), instead of only relying on (very valuable, but nevertheless)
> anecdotal evidence and personal opinions.

+1.  Until then, we can bikeshed over the central phrasing, and then
instructors who disagree strongly enough can adjust the lesson for
their own workshops ;).  It's much more fun when there's an objective
measure to settle disagreements, but such measures have eluded us so
far :p.

iglpdc:
>I think this is just so frozen into the numpy community that it's hard
>for them to transition out of it ;).  But I hardly ever see ‘import …
> as …’ outside of the numpy ecosystem.  

Maybe, but this is where we are :-)

In any case, maybe a good compromise would be to add to your PR an extra sentence (maybe in a callout or something in the extra files) explaining `import ... as` as an alternative syntax and that is commonly use within some fields.

wking:
On Mon, Apr 20, 2015 at 12:36:36PM -0700, Ivan Gonzalez wrote:
> In any case, maybe a good compromise would be to add to your PR an
> extra sentence (maybe in a callout or something in the extra files)
> explaining `import ... as` as an alternative syntax and that is
> commonly use within some fields.

No ‘import … as …’ in the main flow with a short ‘import … as …’
callout box (that instructors can skip or not as they see fit) works
for me.

drj11:
FWIW, I'm +1 for dropping `import ... as ...`. There's a part of the Python community that thinks it's just wrong, even while that part might recognise that subcommunities like numpy practically make a badge of it.

kynan:
I don't think `import ... as ...` adds such a big cognitive load. Consistency with code "in the wild" is important imho since especially beginners mostly work with copy & paste.

Either way I think the examples and exercises have to be consistent i.e. I'm -1 on introducing the syntax later since that will really confuse people.

jttkim:
@kynan in my experience `import ... as ...` is one of those things that incur a negligible cognitive load to those of us with a cognitive system sufficiently conditioned for and by coding, but it can be a considerable issue for beginners. To elaborate on my earlier remark that merging code that uses different aliases for the same module is a rotten job, to someone who realises in seconds that `np` is an identifier of an object of type module with the canonical name `numpy`, such merging jobs are just dull and unnecessary, but I've seen beginners despairing over trying to figure out why some alias works for their buddy (who they're copying & pasting code from) but not for them.

As a cognitive science speculation I think this is a variant of object permanence [1]: While seasoned coders have a reasonably complete mental model of the context relevant to their code (and are quick to figure out where they may have their model wrong), beginners tend to be aware of the code on the screen only. Once code grows to a size where the `import` statements at the top are no longer in view, the difficulty learners have to resolve troubles resulting from missing / variant `import` statements quite noticeably jump up.

[1] http://en.wikipedia.org/wiki/Object_permanence

tbekolay:
As of now, I count 5 in favour of dropping, 4 in favour of keeping, and 1 neutral (as long as it's consistent). It's not a strong consensus, but in the interest of moving forward I'm going to merge this and add a note about it to `discussion.md` as an FAI (Frequently Argued Issue).

wking:
On Thu, Apr 30, 2015 at 06:35:23AM -0700, Trevor Bekolay wrote:
> As of now, I count 5 in favour of dropping, 4 in favour of keeping,
> and 1 neutral (as long as it's consistent).

That sounds close enough that the callout middle ground might be
useful [1].

[1]: https://github.com/swcarpentry/python-novice-inflammation/pull/61#issuecomment-94549045

tbekolay:
Ah, I missed that callout suggestion. I'll add it to #76.

PR #62, opened 2015-04-18
Use command line throughout whole command line lesson

DamienIrving:
At the moment the command line lesson (`08-cmdline.html`) begins by calling python scripts from within the IPython notebook (using `%run script.py` or `!ipython script.py`) and then about half way through the lesson it switches to calling scripts from the command line itself (i.e. `$ python script.py`). This is very confusing for learners and instructors, so I'm proposing that we just use the command line from beginning to end (i.e. don't even open an IPython notebook for this lesson - just teach it entirely from the command line). 

This PR also fixes some little things towards the end of the lesson because no reference was made to `readings-05.py` or `readings-06.py` in the lesson notes.

### DIFF
--- a/08-cmdline.html
+++ b/08-cmdline.html
@@ -63,27 +63,22 @@
 <p>Using the text editor of your choice, save the following in a text file called <code>sys-version.py</code>:</p>
 <pre class="sourceCode python"><code class="sourceCode python"><span class="ch">import</span> sys
 <span class="dt">print</span> <span class="st">&#39;version is&#39;</span>, sys.version</code></pre>
-<p>The first line imports a library called <code>sys</code>, which is short for &quot;system&quot;. It defines values such as <code>sys.version</code>, which describes which version of Python we are running. We can run this script from within the IPython Notebook like this:</p>
-<pre class="in"><code>%run sys-version.py</code></pre>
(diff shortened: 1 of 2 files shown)


gvwilson:
+1.

wking:
On Fri, Apr 17, 2015 at 08:35:23PM -0700, Damien Irving wrote:
>     M 08-cmdline.html (33)

The changes look good to me, but you shouldn't be contributing the
rebuilt HTML, since it increases the chance of a merge conflict, and
resolving merge conflicts in auto-generated content isn't worth the
trouble.  See [1] (point 3).  Instead, the maintainers should
periodically rebuild the HTML from the master branch independently of
in-flight PRs.

[1]: https://github.com/swcarpentry/python-novice-inflammation/blob/29c1ff8e4a75900056f671032ee2c6e942192c0b/CONTRIBUTING.md#getting-started

DamienIrving:
Ah, thanks @wking - I should have read CONTRIBUTING.md more closely. Happy to re-submit the pull request without the .html file if the maintainers would prefer?

tbekolay:
No worries about the .html files being included; it's better practice to not include them, but not a hard and fast rule. Thanks for these changes, big improvement IMO!

PR #64, opened 2015-04-22
Add true/false text to python conditional flowchart

richford:
This pull request adds true/false text to the python conditional flowchart. This is a fairly minor change so I understand if it doesn't count as my PR for the completion of instructor training. But I want to keep my PRs atomic, so I'm submitting this for a quick and minor improvement.


tbekolay:
Looks great, thanks!

PR #65, opened 2015-04-22
Fixing link to `library` and other small fixes

kylerbrown:
Fixing link the glossary term "library" from the lesson one page to the reference page.

Also removed potentially demotivating word "simple".

For the Operations across axes figure, I moved "average across axis 1" to the left because the "1" was being cut off.


tbekolay:
Great, thanks! I think the SVG looks fine on certain computers (with certain fonts), but if moving it is necessary on some machines, definitely worth doing.

PR #67, opened 2015-04-24
Change order of output and error and corrects error message in 02-loop.m...

hjelmj:
...d

This commit changes the order of the output after executing the print
statements to demonstrate an "out-of-index" error in 02-loop.md.

The error message that followed the print statement also refers to
a function `print_elements` that is not defined (at the moment) in this lesson.

As this lesson (02-loop.md) now comes before "Creating functions" (06-func.md)
I have assumed that we are interested in removing references to
`print_elements`.

### DIFF
--- a/02-loop.md
+++ b/02-loop.md
@@ -95,7 +95,7 @@ for char in word:
     print char
 ~~~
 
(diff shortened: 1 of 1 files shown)


tbekolay:
Your assumption is correct. Thanks for the PR!

Issue #70, opened 2015-04-25
comparison operator typo in conditionals lesson (with fix)


janetriley:
In 

```
if 4 > 5:
    print 'A'
elif 4 =< 5:
    print 'B'
elif 4 < 5:
    print 'C'
```

The first elif should be `==`

[Pull request for fix is here.] (https://github.com/swcarpentry/python-novice-inflammation/pull/71)

PR #71, opened 2015-04-25
Changed =< to ==

janetriley:
Fix issue #70 .


tbekolay:
Great, thanks for the PR! In the future, don't worry about remaking the HTML, I'll do that when I merge the PR to ensure that all the HTML is generated with the same environment.

PR #75, opened 2015-04-30
Removed repeated challenge

ewanbarr:
There appears to have been a mistake in a including the challenges, as there was an extra:
"
>
> Plot scaling: why do all of our plots stop just short of the upper end of our graph?
"

in the first challenge. This is marked as its own challenge further down.


tbekolay:
Good catch, thanks!

PR #76, opened 2015-04-30
Import cleanup

tbekolay:
As per the discussion in #61, I added a pointer to why we're not using aliasing in this lesson to the `instructors.md` file. I put this in a Frequently Argued Issues (FAI) section, which I stole from @jdblischak (thanks for the idea!)

In grepping to make sure we're not aliasing imports anywhere in this lesson, I noticed that we do `from ... import` in `09-errors.md`, so I took out that syntax there as well. We still do this in `10-supplementary.md`, but it's explained there, and supplementary material, so I left it in.

### DIFF
--- a/01-numpy.md
+++ b/01-numpy.md
@@ -640,6 +640,15 @@ and that we want a tight layout.
 if we leave out that call to `fig.tight_layout()`,
 the graphs will actually be squeezed together more closely.)
 
+> ## Scientists dislike typing {.callout}
+>
+> We will always use the syntax `import numpy` to import NumPy.
+> However, in order to save typing, it is
+> [often suggested](http://www.scipy.org/getting-started.html#an-example-script)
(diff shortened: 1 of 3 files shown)


wking:
On Thu, Apr 30, 2015 at 07:13:09AM -0700, Trevor Bekolay wrote:
> In grepping to make sure we're not aliasing imports anywhere in this
> lesson, I noticed that we do `from ... import` in `09-errors.md`, so
> I took out that syntax there as well. We still do this in
> `10-supplementary.md`, but it's explained there, and supplementary
> material, so I left it in.

I don't feel so strongly about this.  I personally like to avoid the
‘from … import …’ syntax, but I don't think it's as crazy as creating
your own local name for a module ;).  Removing it does reduce our
syntax load though, so +1 there.

PR #77, opened 2015-05-03
Removal for function (too early in lesson program)

ewanbarr:
In this lesson there is an example that uses a function "def sign(num)" to illustrate some decision making capability. I don't know if this material has been reordered recently, but given that function definitions do not appear until the next lesson, I would argue that this be removed and just shown as a simple block of code that prints whether a number is -ve or +ve. I have suggested a potential replacement.

### DIFF
--- a/05-cond.md
+++ b/05-cond.md
@@ -65,18 +65,17 @@ which is short for "else if".
 This makes it simple to write a function that returns the sign of a number:
 
 ~~~ {.python}
-def sign(num):
-    if num > 0:
-        return 1
-    elif num == 0:
-        return 0
-    else:
-        return -1
+num = -3
 
-print 'sign of -3:', sign(-3)
+if num > 0:
+    print num,"is +ve"
+elif num == 0:
+    print num,"is 0"
+else:
+    print num,"is -ve"
 ~~~
 ~~~ {.output}
-sign of -3: -1
+"-3 is -ve"
 ~~~
 
 One important thing to notice in the code above is that we use a double equals sign `==` to test for equality


05-cond.md
```diff
@@ -65,18 +65,17 @@ which is short for "else if".
 This makes it simple to write a function that returns the sign of a number:
```
tbekolay: This line should be reworded to not reference functions. Perhaps something like "The following Python code uses `elif` to print the sign of a number."

05-cond.md
```diff
@@ -65,18 +65,17 @@ which is short for "else if".
...
-    elif num == 0:
-        return 0
-    else:
-        return -1
-
-print 'sign of -3:', sign(-3)
+num = -3
+
+if num > 0:
+    print num,"is +ve"
```
tbekolay: Not sure if this is just me, but I was always confused in physics class when someone wrote "+ve", and would prefer the word "positive" to this. But, I'm willing to concede this if most people prefer "+ve"!

05-cond.md
```diff
@@ -65,18 +65,17 @@ which is short for "else if".
...
+elif num == 0:
+    print num,"is 0"
+else:
+    print num,"is -ve"
```
tbekolay: Ditto here, prefer "negative".

tbekolay:
You are correct, we did reorder the lessons recently, thanks for your PR! I made some suggestions on the commit. Please let me know what you think, and either update the PR (by adding another commit and pushing to that branch), or just let me know if you're okay with those changes and I can make them if you'd prefer.

05-cond.md
gvwilson: Please use "positive" - a lot of our readers will have English as a 
second language, and we shouldn't use shortcuts.

On 2015-05-04 10:48 AM, Trevor Bekolay wrote:
>
> In 05-cond.md 
> <https://github.com/swcarpentry/python-novice-inflammation/pull/77#discussion_r29590226>:
>
> > @@ -65,18 +65,17 @@ which is short for "else if".
> >  This makes it simple to write a function that returns the sign of a number:
> >
> >  ~~~ {.python}
> > -def sign(num):
> > -    if num > 0:
> > -        return 1
> > -    elif num == 0:
> > -        return 0
> > -    else:
> > -        return -1
> > -
> > -print 'sign of -3:', sign(-3)
> > +num = -3
> > +
> > +if num > 0:
> > +    print num,"is +ve"
>
> Not sure if this is just me, but I was always confused in physics 
> class when someone wrote "+ve", and would prefer the word "positive" 
> to this. But, I'm willing to concede this if most people prefer "+ve"!
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/swcarpentry/python-novice-inflammation/pull/77/files#r29590226>.
>

-- 
Dr. Greg Wilson    | gvwilson@software-carpentry.org
Software Carpentry | http://software-carpentry.org

iglpdc: Yep.

ewanbarr: Hey,

I think that either "positive"  or "+"  (mathematics is meant to be a
universal language after all) would be fine. I agree with the comments on
the use of +ve as a contraction.

I am happy for the suggested change to be committed.

Ewan

tbekolay: Done, thanks again for your contribution!

PR #78, opened 2015-05-05
Write a legend for the instructor guide

### LINKED ISSUE #46: What is wrong with the inflammation data?
I am preparing for teaching next week, and I have some questions:

1. Throughout the lesson, it is indicated that there is something wrong with the data. I have yet to figure out what exactly the problem is?

2. A more teaching technical thing: in 02 Checking our data, it is indicated that you can stuff the shown if statement into the for loop shown in 01 Processing multiple files. Do people do that when they teach, or do they just show them on screen and explain?

3. When I do put that if statement in there, I get either suspicious maxima or adding up to zero on all of them. Is this actually correct?

tbekolay:
Also included the gen_inflammation.py script that was in the old `bc` repository. This seems a good place to put it, since it is Python after all...

Fixes #46.

### DIFF
--- a/instructors.md
+++ b/instructors.md
@@ -5,7 +5,12 @@ subtitle: Instructor's Guide
 ---
 ## Legend
 
-FIXME: story behind running example.
+We are using a dataset with records on inflammation from patients following an
+arthritis treatment.
+
+We make reference in the lesson that this data is somehow strange. It is strange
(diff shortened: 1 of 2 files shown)


tbekolay:
Rebased this due to the recent merging. This is a pretty small FIXME fix, so good to get to this before the bigger PRs.

PR #79, opened 2015-05-05
Split debug into a separate topic, and other fixes

tbekolay:
Topic 7 required a second level of subheadings, which indicated to me
that debugging should be its own topic. One thing that's missing from this topic
right now is a set of challenges. I added a single challenge, which
I have done while teaching in the past, but additional challenges
would be good too.

This also includes several other fixes so that this repo passes `tools/check.py`,
except for some things that are currently in discussion in [lesson-template](https://github.com/swcarpentry/lesson-template).


abostroem:
Looks good to me, lets merge in the morning if no other comments

tbekolay:
Rebased this branch due to conflicts with newly merged stuff.

PR #80, opened 2015-05-06
5.3 clean-up: Part 1

abostroem:
Clarifying text for section 01 - 05. I'll finish 06-09 tomorrow.

PR #92, opened 2015-05-09
Fix typo in first lesson

mkcor:
Keep reading and preparing...


tbekolay:
I think this is fixed in #80; I'm merging that now, so perhaps wait a bit to read through?

abostroem:
Confirming this is fixed in #80.

PR #93, opened 2015-05-10
Add getready callout

iglpdc:
[This PR in lesson-example](https://github.com/swcarpentry/lesson-example/pull/3) add a "getready" callout box in the `index.md` that can be used to provide the learners with instructions on "what they should do to follow the lessons".

The PR here adds the callout box. I've also added a zip file of it linked in the callout box with the inflammation datafiles in `data`.


tbekolay:
Great, thanks! Note to self and @abostroem: if the `.csv` files change, we must make sure to update this zip file.

PR #95, opened 2015-05-10
Removed the rest of the KeyError references

tbekolay:
in the Errors and Exceptions lesson.

I chose to keep the KeyError in the first challenge, because it is
a common thing to run into an exception that you've never heard of
before; students should have no problems answering the 6 questions
in the challenge despite not seeing a KeyError before. If anything,
it makes this challenge a bit cleaner, as we're mostly asking for
generic answers rather that should be the same regardless of the
error type.

I think this is the last minor fix necessary for 5.3. IMO!


abostroem:
Looks good to me.

PR #102, opened 2015-05-14
Fixing typo in N squared

kylerbrown:
N^(2/2) = N, I think you mean N^2, especially because N^2 is referenced  two paragraphs below.


abostroem:
Thanks for catching this, I spent a while trying to figure out what N^2/2 meant when I did my first read through,  I swear I changed it when I made my corrections, but apparently not...

PR #103, opened 2015-05-14
mention: from __future__ import division

ppomorsk:
This is another way to fix a bug caused by expecting a floating point answer from integer division.  This could be helpful to users who are really uncomfortable with the way Python 2 does integer division.

### DIFF
--- a/06-func.md
+++ b/06-func.md
@@ -169,6 +169,16 @@ a/b is: 3
 float(a)/b is: 3.33333333333
 ~~~
 
+It is also possible to make Python use integer division that produces a floating point answer, by importing a feature that was added in later versions.  
+
+~~~ {.python}
+from __future__ import division
+print '10/3 is:', 10/3
+~~~
+~~~ {.output}
+10/3 is: 3.33333333333
+~~~
+
 Let's fix our `fahr_to_kelvin` function with this new knowledge:
 
 ~~~ {.python}


abostroem:
This seems like it might create more confusion that it helps. For instance I can see students typing that in the notebook and then having integer division turned off for the remaining part of the session. I think this could be mentioned in passing (possibly called out in the instructor notes), but I'm -1 for including it in the lesson. Other thoughts?

iglpdc:
I agree with @abostroem. It can be confusing. 

I maybe good thought to have a callout here mentioning that this has changed in Python 3. I think Python 3 is not mentioned at all in the lesson. So, it could be good to have a callout "Two Python dialects" or something like that, where we explain that there are to major versions of Python, that they are incompatible, and this is one of the differences. 

People is likely to see this in the internet or try to use code for one version in the other. Also it's not strange to have someone who accidentally installed Python 3 and is "complaining" about not getting in `10/3 = 3`.

drj11:
I agree @abostroem too. As well as the already mentioned risk of having some part of the group with integer division and some part with floating point division, I think the syntax and semantics of `from __future__` are not novice material: the double underscore thing just looks weird; `from __future__` has the magic effect of _changing the language_ which no other sort of import can do.

+1 for a "Two Python Dialects" callout.

ppomorsk:
I agree with the comments, this change may indeed introduce confusion.  

Python 3 does actually get a one sentence mention in this lesson already. (quote: " It's still confusing, though, so Python 3 produces a floating-point answer when dividing integers if it needs to.") This is the only mention of Python 3 that I can see in the entire "Programming with Python" workshop.

It would be good to add some note about the two dialects of Python, maybe somewhere in the introduction.  

A deeper question is whether this workshop should just be entirely switched to Python 3, but that's a topic for a longer discussion.

PR #104, opened 2015-05-15
Update 01-numpy.md

mboisson:
Removed 4 exercises that were duplicates from higher up in the page.


gvwilson:
Thanks for catching the duplication, but can you please remove the exercises that are in the middle of the file and leave them all at the end?  Our rule is that if you really need exercises in the middle of a lesson, you should split the lesson into two pieces - we'll probably do that for this lesson after the 5.3 release, but we're not going to do that refactoring right now.
Thanks,
Greg

mboisson:
Ok. I modified the pull request to remove the exercises from the middle of the file rather than the ones from the end.

Maxime

PR #107, opened 2015-05-16
Replaced references to svg files with references to png

fonnesbeck:
Four Markdown pages with replaced file references.

### DIFF
--- a/01-numpy.md
+++ b/01-numpy.md
@@ -124,7 +124,7 @@ we can print several things at once by separating them with commas.
 If we imagine the variable as a sticky note with a name written on it,
 assignment is like putting the sticky note on a particular value:
 
-![Variables as Sticky Notes](fig/python-sticky-note-variables-01.svg)
+![Variables as Sticky Notes](fig/python-sticky-note-variables-01.png)
 
 This means that assigning a value to one variable does *not* change the values of other variables.
 For example,
@@ -138,7 +138,7 @@ print 'weight in kilograms:', weight_kg, 'and in pounds:', weight_lb
 weight in kilograms: 57.5 and in pounds: 126.5
 ~~~
 
-![Creating Another Variable](fig/python-sticky-note-variables-02.svg)
+![Creating Another Variable](fig/python-sticky-note-variables-02.png)
 
 and then change `weight_kg`:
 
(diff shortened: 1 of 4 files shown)


tbekolay:
Why prefer `png` over `svg`? I would favour `svg` as they're scalable (will look good regardless of how much you zoom in). This is especially important on retina displays and the like. But, if there are good reasons for favouring `png` we can switch them over.

fonnesbeck:
They do not display in Github pages because they have the potential for cross-site scripting vulnerabilities. Hence, the look like this:

gh-pages

abostroem:
Can someone who was involved in the lesson template design comment on this? @gvwilson @r-gaia-cs ? I was also under the impression that we preferred svg

fonnesbeck:
I guess it depends on whether we want users to be able to read the lessons on Github pages or not. I assumed this is why there are two versions of each graphic.

r-gaia-cs:
> Can someone who was involved in the lesson template design comment on this?

As @tbekolay said, we prefer SVG. The reasons are (1) they are scalable, (2) editable and, in some cases, (3) smaller than PNG (\*). The only side effect is that SVG are font dependent but we are OK with that for now.

> I guess it depends on whether we want users to be able to read the lessons on Github pages or not. I assumed this is why there are two versions of each graphic.

We are using [GitHub Pages](http://pages.github.com/) and users can get the SVG on [it](https://swcarpentry.github.io/python-novice-inflammation/discussion.html):

screen shot 2015-05-17 at 18 19 19

If someone try to [preview](https://github.com/swcarpentry/python-novice-inflammation/blob/gh-pages/discussion.md) the lessons on GitHub (**not GitHub Pages**) they will not get the SVG inplace:

screen shot 2015-05-17 at 18 21 16

@abostroem Did I miss something?

@fonnesbeck Did this make things clear? Suggestions of how we can improve `CONTRIBUTING.md` are **very** welcome.

(\*) About the size, SVG can be smaller than PNG but this depends of (1) the program used for create the files and (2) the resolution of the PNG.

gvwilson:
We prefer SVG because they rescale much more cleanly for smaller 
devices.  However, if they aren't displaying in the GitHub pages, that's 
a moot point.  Can someone please check the display?

fonnesbeck:
@r-gaia-cs is correct, it is only on Github preview that the svg restriction applies, so that's probably not a big deal. Closing.

gvwilson:
Thanks @fonnesbeck - this still counts as a PR for training purposes, so 
consider that box ticked.

PR #108, opened 2015-05-18
Improving Python lessons a bit 

narayananr:
I made two changes. On the lesson "Repeating Actions with Loops", I expanded the example a bit on looping through a string. And on the lesson "Analyzing Data from Multiple Files", the glob example was trying to "glob" the files that were not in student's working directory for the lesson. I changed it so that the "glob" will list of files in the student's working directory.

### DIFF
--- a/04-files.md
+++ b/04-files.md
@@ -37,7 +37,7 @@ which means we can loop over it
 to do something with each filename in turn.
 In our case,
 the "something" we want to do is generate a set of plots for each file in our inflammation dataset.
-Let's test it by analyzing the first three files in the list:
+Let's test it by analyzing the first three files in the list, instead of all the csv files:
 
 ~~~ {.python}
 filenames = glob.glob('*.csv')
--- a/04-files.md
+++ b/04-files.md
@@ -21,14 +21,14 @@ that finds files whose names match a pattern.
 We provide those patterns as strings:
 the character `*` matches zero or more characters,
 while `?` matches any one character.
-We can use this to get the names of all the html files:
+We can use this to get the names of all the csv files in the data directory:
 
 ~~~ {.python}
-print glob.glob('*.html')
+print glob.glob('*.csv')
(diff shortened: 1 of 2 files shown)


abostroem:
@narayananr thanks for these changes. The first change modifies a line that I just changed for a different pull request. Can you take a look and let me know what you think?

I also just changed this lesson to look at html files rather than csv files since we now keep the csv files in a data directory and students are generally working in the top level directory.

The third change is good, but can you change the word csv to html?

04-files.md
```diff
@@ -21,14 +21,14 @@ that finds files whose names match a pattern.
 We provide those patterns as strings:
 the character `*` matches zero or more characters,
 while `?` matches any one character.
-We can use this to get the names of all the html files:
+We can use this to get the names of all the csv files in the data directory:
```
QuLogic: The code below does not use the "data" directory, so maybe this should say "current" directory, since that's what it does.

abostroem:
@QuLogic I don't think it adds anything to the lesson for students to list CSV files in the data directory vs html files in the current directory and I think it adds a layer of complication (because they have to look in a different direction). Is there a reason you prefer the CSV files?

QuLogic:
I think you meant to tag @narayananr; I didn't make that change.

narayananr:
Hi

I suggested the change because I had the same thought.  This lesson

http://swcarpentry.github.io/python-novice-inflammation/04-files.html

currently, uses glob on "html" files, and few statements after it uses glob
on "csv" files on data directory. I just thought of making the glob
statement the same (either html or csv depending on the current working
directory). I might have made mistake assuming the working directory.

Thanks
Narayanan

tbekolay:
Somewhat improved versions of the `.csv` change have already been merged, but the expanded text on the print statements is a welcome change. Thanks @narayananr!

PR #109, opened 2015-05-20
Minor edits to 06-func.md for grammar and typos.

malvikasharan:
I roughly edited the grammar and typos. This can be further improved but I was not sure how intensively can we edit these pre-existing documents.

### DIFF
--- a/06-func.md
+++ b/06-func.md
@@ -22,10 +22,9 @@ and didn't want to generate a figure for every single one?
 Commenting out the figure-drawing code is a nuisance.
 Also, what if we want to use that code again,
 on a different dataset or at a different point in our program?
-Cutting and pasting it is going to make our code get very long and very repetative,
-very quickly.
-We'd like a way to package our code so that it is easier to reuse,
-and Python provides for this by letting us define things called 'functions' -
+Cutting and pasting it will make our code very long and very repetitive,
+very quickly whereas we would like a way to package our code that makes it easy to reuse.
+Python provides a solution for this by letting us define 'functions' -
 a shorthand way of re-executing longer pieces of code.
 
 Let's start by defining a function `fahr_to_kelvin` that converts temperatures from Fahrenheit to Kelvin:
@@ -62,13 +61,13 @@ boiling point of water: 273.15
 
 We've successfully called the function that we defined,
 and we have access to the value that we returned.
-Unfortunately, the value returned doesn't look right.
+Unfortunately, the returned value doesn't look right.
 What went wrong?
 
 ### Debugging a Function
 
 *Debugging* is when we fix a piece of code
-that we know is working incorrectly.
+that we know is not working correctly.
 In this case, we know that `fahr_to_kelvin`
 is giving us the wrong answer,
 so let's find out why.
@@ -137,11 +136,9 @@ The computer does this for historical reasons:
 integer operations were much faster on early machines,
 and this behavior is actually useful in a lot of situations.
 It's still confusing,
-though,
-so Python 3 produces a floating-point answer when dividing integers if it needs to.
+therefore, Python 3 produces a floating-point answer when dividing integers if it needs to.
 We're still using Python 2.7 in this class,
-though,
-so if we want `5/9` to give us the right answer,
+therefore, if we want `5/9` to give us the right answer,
 we have to write it as `5.0/9`, `5/9.0`, or some other variation.
 
 Another way to create a floating-point answer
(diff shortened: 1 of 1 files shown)


abostroem:
@malvikasharan thank you for this PR. The second change with "returned value" reads much better. 
For the first change: 
I like changing "is going" --> will.
I think connecting the second sentence to the first makes it more difficult to read. How about something along the lines of "Python lets us avoid this by defining functions - a shorthand way of re-executing longer pieces of code. Functions allow us to package our code that makes it easy to reuse.

06-func.md
```diff
@@ -22,10 +22,9 @@ and didn't want to generate a figure for every single one?
 Commenting out the figure-drawing code is a nuisance.
 Also, what if we want to use that code again,
 on a different dataset or at a different point in our program?
-Cutting and pasting it is going to make our code get very long and very repetative,
-very quickly.
-We'd like a way to package our code so that it is easier to reuse,
-and Python provides for this by letting us define things called 'functions' -
+Cutting and pasting it will make our code very long and very repetitive,
+very quickly whereas we would like a way to package our code that makes it easy to reuse.
+Python provides a solution for this by letting us define 'functions' -
```
tbekolay: I'm +1 Azalee's rewording of this; putting the two sentences together makes it a bit long.

06-func.md
```diff
@@ -137,11 +136,9 @@ The computer does this for historical reasons:
 integer operations were much faster on early machines,
 and this behavior is actually useful in a lot of situations.
 It's still confusing,
-though,
-so Python 3 produces a floating-point answer when dividing integers if it needs to.
+therefore, Python 3 produces a floating-point answer when dividing integers if it needs to.
```
tbekolay: I like dropping the 'though', but I don't think therefore makes grammatical sense here because "therefore" doesn't act as a conjunction like and, or, but, so; if you wanted to say therefore, you would need to start a new sentence. I'd suggest keeping 'so', but dropping 'thought'; i.e., `It's still confusing, so Python 3 produces...`

06-func.md
```diff
@@ -137,11 +136,9 @@ The computer does this for historical reasons:
...
 We're still using Python 2.7 in this class,
-though,
-so if we want `5/9` to give us the right answer,
+therefore, if we want `5/9` to give us the right answer,
```
tbekolay: same here `We're still using Python 2.7 in this class, so if we want ...`.

06-func.md
```diff
@@ -281,8 +278,8 @@ for f in filenames:
 ~~~
 
 By giving our functions human-readable names,
-we can more easily read and understand what is happening in the `for` loop.
-Even better, if at some later date we want to use either of those pieces of code again,
+we can easily read and understand what is happening in the `for` loop.
+Even better, if in future we want to use either of those pieces of code again,
```
tbekolay: 'in the future'

06-func.md
```diff
@@ -372,9 +369,8 @@ difference in standard deviations before and after: -3.5527136788e-15
 Again,
 the difference is very small.
 It's still possible that our function is wrong,
-but it seems unlikely enough that we should probably get back to doing our analysis.
-We have one more task first, though:
-we should write some [documentation](reference.html#documentation) for our function
+but it seems unlikely therefore we should probably get back to doing our analysis.
```
tbekolay: Here also, we could use "so" but not "therefore". I kind of like the original in this case though...

tbekolay:
I made some inline comments similar to Azalee's. Aside from those nits, these are great, thank you!

06-func.md
malvikasharan: Agreed! thanks for the feedback.

06-func.md
malvikasharan: yes, makes sense.

06-func.md
malvikasharan: Hmm, i think i used therefore too often.

06-func.md
QuLogic: I'm not sure what sort of differentiation was made between integers and floating-point in the rest of the tutorials, but this statement is not exactly correct. Python 3 _always_ produces a floating point answer from division, not only "if it needs to".
```python
$ python3 -c 'print(3/3)'
1.0
```

06-func.md
```diff
@@ -226,7 +223,7 @@ or the next person who reads it won't be able to understand what's going on.
 ### Tidying up
 
 Now that we know how to wrap bits of code up in functions,
-we can make our inflammation analyasis easier to read and easier to reuse.
+we can make our inflammation analyasis easy to read and reuse.
```
QuLogic: You can correct the spelling of `analysis` here too.

malvikasharan: yes, totally missed that!

abostroem:
@malvikasharan How close are you to addressing these comments?

PR #110, opened 2015-05-20
adding problems with modifying lists in place

jeremycg:
added a short section to the lists lesson warning against modifying
lists in place when two or more variables are pointing to the same list.

It may be worth including a sticky note figure for this point? I think it is a worthwhile addition to the lesson as this point cost me several hours of bugfixing when learning, and we have explicitly talked about the opposite behaviour in lesson 1. 

Apologies for the slowness of this pull request.

### DIFF
--- a/03-lists.md
+++ b/03-lists.md
@@ -119,8 +119,9 @@ primes [1, 3, 5, 7, 2]
 odds [1, 3, 5, 7, 2]
 ~~~
 
-This is because python stores a list in memory, and then can use multiple pointers to the same list.
-We should almost always explicitly use a list() command, so we do not modify a list we did not mean to:
+This is because python stores a list in memory, and then can use multiple names to refer to the same list.
+If all we want to do is copy a (simple) list, we can use the list() command, so we do not modify a list we did not mean to:
+
 ~~~ {.python}
 odds = [1, 3, 5, 7]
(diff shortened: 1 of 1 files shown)


tbekolay:
I agree that this can be very counterintuitive to people; the feeling is that when I *want* a copy, it should be a copy, and when I don't, it shouldn't be. Of course, computers don't know our intent, they have predefined rules.

So, I like the first part of these changes, showing that the list isn't copied but just referred to twice. I don't think that we should introduce the word "pointer", as that opens a huge can of worms. I also don't think that we should recommend "almost always explicitly use a list() command"; that's a very black-and-white statement that can cause bad habits (e.g., in an `__init__` method, should you be making copies of all incoming lists? Probably not.)

As for how to proceed with this PR, do you think you could remove the word "pointer" and make a less-strong suggestion as to when to use `list` and when not to?

On a related sidenote, [this PyCon talk by Ned Batchelder](https://www.youtube.com/watch?v=_AEJHKGk9ns) is a wonderful thing to point people to if they're curious why this happens, and why it doesn't seem consistent (but actually is).

jeremycg:
Thanks for the feedback - I have changed pointer to name, and made the language about when to use list() weaker.

Thanks for the video - it's a great overview of some weird python stuff.

PR #114, opened 2015-05-21
Added execution of final if statement to 05-cond.md

DamienIrving:
The concept of using an if statement to check the inflammation data is introduced at the end of this lesson but is never actually executed to check that it works. I've added some content that does just that.

I think this is a particularly nice segue into the following lesson on functions. The way I've written the end of the lesson requires repetition / cut and paste, which is something that can be avoided by using functions...

### DIFF
--- a/05-cond.md
+++ b/05-cond.md
@@ -141,8 +141,38 @@ else:
     print 'Seems OK!'
 ~~~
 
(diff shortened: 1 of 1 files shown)


tbekolay:
Looks great, thanks!

PR #115, opened 2015-05-21
Update to challenges in 06-func.md 

DamienIrving:
This pull request fixes a typo in the markdown for one of the output cells and also updates the challenges in `06-func.md` so that they are more consistent with the order in which particular concepts are introduced throughout the lesson.


tbekolay:
Definitely makes more sense this way!

PR #116, opened 2015-05-21
Minor edits to challenges in 08-cmdline.md

DamienIrving:
I've made a couple of minor edits to the challenges in `08-cmdline.md`:
* For the addition/subtraction lesson I've changed it so `arith.py` takes the word `addition`or `subtraction` as an input argument and not `+` or `-`, because the use of the symbols always creates a lot of unnecessary confusion
* There was a typo in one of the challenges where it told people to edit `count-stdin.py` when it should have said `readings.py`

### DIFF
--- a/08-cmdline.md
+++ b/08-cmdline.md
@@ -520,13 +520,13 @@ the program now does everything we set out to do.
 > Write a command-line program that does addition and subtraction:
 >
 > ~~~ {.python}
-> $ python arith.py 1 2 addition
+> $ python arith.py add 1 2
 > ~~~
 > ~~~ {.output}
 > 3
 > ~~~
 > ~~~ {.python}
-> $ python arith.py 3 4 subtraction
+> $ python arith.py subtract 3 4
 > ~~~
 > ~~~ {.output}
 > -1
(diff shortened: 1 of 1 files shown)


tbekolay:
Thanks for the typo change! I'm not sure about the changes to the `arith.py` challenge. Can you be a bit more specific about why symbols cause confusion? Do students not realize they are strings like any other?

If using words rather than symbols is better, can I suggest something more like
```
$ python arith.py add 1 2
$ python arith.py subtract 3 4
```
? I think using active verbs makes this more clear what is expected output (and essentially looks like a function, which we've introduced by this point).

DamienIrving:
@tbekolay Often the symbols create confusion because people start google searching and end up trying to implement the `eval()` function, which is a bit beyond the scope of this lesson. I think the numbers get the point across about that fact that input arguments come as strings which may need to be converted to integers or strings, so the symbols can be left out without compromising the main lesson of that challenge. 

I like your suggestion about the active verbs, so I've updated the PR accordingly.

tbekolay:
Thanks for the clarification! I agree, we really don't want anything resembling `eval`, and I think this change makes that clear.

Merging!

PR #118, opened 2015-05-25
Fixed typos in 01-numpy.md

DamienIrving:
There are a number of typos in `01-numpy.md` that this PR corrects.

It also imports `numpy` and `matplotlib.pyplot` as `np` and `plt` respectively in the final subplot to ensure consistency with `04-files.md`.

### DIFF
--- a/01-numpy.md
+++ b/01-numpy.md
@@ -509,8 +509,8 @@ Let's take a look at the average inflammation over time:
 
 ~~~ {.python}
 ave_inflammation = data.mean(axis=0)
-ave_plot = pyplot.plot(ave_inflammation)
-pyplot.show(ave_plot)
+ave_plot = matplotlib.pyplot.plot(ave_inflammation)
+matplotlib.pyplot.show(ave_plot)
 ~~~
 
 ![Average Inflammation Over Time](fig/01-numpy_73_0.png)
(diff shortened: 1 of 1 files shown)


01-numpy.md
```diff
@@ -490,8 +490,16 @@ Visualization deserves an entire lecture (or course) of its own,
...
 While there is no "official" plotting library,
 this package is the de facto standard.
+
 First,
-we will import the `pyplot` module from `matplotlib`
+we need to to run a special IPython notebook command
+to ensure that our figures display within the notebook itself.
+
+~~~ {.python}
+% matplotlib inline
```
tbekolay: While we mention running this in an IPython notebook at the outset, these lessons are designed to run in a normal Python interpreter. Perhaps we can put this in a callout? It might be a good thing to have specific styling for IPython specific callouts... I made issue #119 for that :)

01-numpy.md
```diff
@@ -557,12 +568,12 @@ axes3). Once a subplot is created, the axes are can be titled using the
 Here are our three plots side by side:
 
 ~~~ {.python}
-import numpy
-import matplotlib.pyplot
+import numpy as np
```
tbekolay: We have previously decided against this syntax. See the [FAI](https://github.com/swcarpentry/python-novice-inflammation/blob/gh-pages/instructors.md#frequently-argued-issues-fai) and PR #61.

tbekolay:
As mentioned inline, we voted to drop the `import ... as` syntax in #61. If we're still doing this in `04-files.md`, then that should be fixed; I thought I grepped to get all of those, but I may have missed some.

DamienIrving:
@tbekolay Thanks for the feedback. I think it will be easiest for me to close this PR and submit a new one...

PR #120, opened 2015-05-25
Matplotlib fixes

DamienIrving:
This PR:
* Corrects a number of typos in `01-numpy.md` relating to the use of `matplotlib.pyplot`
* Removes the use of abbreviated library names (`np` and `plt`) in `04-files.md`
* Adds a callout to `01-numpy.md` about `% matplotlib inline`

### DIFF
--- a/01-numpy.md
+++ b/01-numpy.md
@@ -511,7 +511,7 @@ inflammation rises and falls over a 40-day period.
 > If you're using an IPython / Jupyter notebook,
 > you'll need to execute the following command
 > in order for your matplotlib images to appear
-> when `show()` is called:
+> in the notebook when `show()` is called:
 >
 > ~~~ {.python}
 > % matplotlib inline
--- a/01-numpy.md
+++ b/01-numpy.md
@@ -512,7 +512,11 @@ inflammation rises and falls over a 40-day period.
 > you'll need to execute the following command
 > in order for your matplotlib images to appear
 > when `show()` is called:
-> `% matplotlib inline`. 
+>
+> ~~~ {.python}
+> % matplotlib inline
+> ~~~
+>  
 > The `%` indicates an IPython magic function - 
 > a function that is only valid within the notebook environment. 
 > Note that you only have to execute this function once per notebook.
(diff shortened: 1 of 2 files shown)


DamienIrving:
(@tbekolay This is the new PR to replace #118)

01-numpy.md
```diff
@@ -505,19 +505,35 @@ matplotlib.pyplot.show(image)
 Blue regions in this heat map are low values, while red shows high values.
 As we can see,
 inflammation rises and falls over a 40-day period.
+
+> ## Some IPython magic {.callout}
+>
+> If you're using an IPython / Jupyter notebook,
+> you'll need to execute the following command
+> in order for your matplotlib images to appear
```
QuLogic: To appear _in_ the notebook, because there's generally nothing stopping them from appearing otherwise (unless you're over SSH or something).

abostroem:
@QuLogic I like "in the notebook" but I will say that when you don't use any magic the backend doesn't really work. You can click on or modify any of the windows that pop up. 

Personally, one of the things that sold me on python was the interactivity of the plotting window which you lose when you start plotting inline. I think in the past I've just used %matplotlib without the inline. Opinions on interactivity vs. ease of having everything in one window?

jdcorless:
@abostroem in a workshop last week, I tried to demo the separate window plotting before the `inline` command because I too like that feature (note that I had the whole class using the IPython notebook for the Python lesson).  This worked fine for those on Windows and Linux, but several students with Macs had their system freeze because of lack of X window installation (root cause thanks to Mac experts, which I am not).  But I agree with your comment about the utility of the separate window plotting.

iglpdc:
> This worked fine for those on Windows and Linux, but several students with Macs had their system  freeze because of lack of X window installation (root cause thanks to Mac experts, which I am not). But I agree with your comment about the utility of the separate window plotting.

Probably, they didn't have Xcode, X11 or some related tools available. Apple stopped shipping them in the OSX default installation a few versions ago [1].

[1] https://support.apple.com/en-us/HT201341

DamienIrving:
I've added "in the notebook" as suggested. With respect to the question of inline or no inline, I think for beginners having the image appear inline is by far the simplest solution. We definitely do not want to head down the X11 / Xcode road on the very first python lesson!

Issue #84, opened 2015-05-07
Add exercise to loops.md


abostroem:
We should have at least one exercise at the end of each lesson.

PBarmby:
Azalee -- did you mean "lists" rather than loops? There are exercises in loops but none in lists.

abostroem:
Yes, thanks for catching that

PR #122, opened 2015-05-29
Added exercise on lists

emilydolson:
As issue https://github.com/swcarpentry/python-novice-inflammation/issues/84 pointed out, there were no exercises for the lists module, so I wrote this one. It hasn't been tested in a software carpentry setting (that I know of), but it worked well in the Intro to Python class that I TAed. It hints at the idea of accumulators, which I think is valuable to touch on.

It would probably be good to also have an exercise or two that deals with indexing into lists, but I'm having trouble coming up with one that is neither obvious from the examples nor really convoluted.

### DIFF
--- a/03-lists.md
+++ b/03-lists.md
@@ -132,3 +132,15 @@ print 'odds after reversing:', odds
 ~~~ {.output}
 [11, 7, 5, 3]
 ~~~
+
+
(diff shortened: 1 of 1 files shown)


abostroem:
Looks good to me. Thanks for addressing one of the issues.

PR #126, opened 2015-06-10
Moving motivation to instructors guide

gvwilson:
1.  Incorporate deletion of slideshow files from https://github.com/swcarpentry/lesson-template/pull/273
2.  Move content of `motivation.md` to `instructors.md`
3.  Pull in other recent changes to lesson template.


abostroem:
@gvwilson Why does this include 250+ commits changing 142 files? Moving the motivation to the instructor notes doesn't seem like it should be that big.

gvwilson:
You're right, it shouldn't - I'll close this one and try again.

PR #128, opened 2015-06-13
Fix matplotlib and numpy imports

guyer:
Code examples are inconsistent with the imports that have already taken
place.
Introduced `%matplotlib inline` for Jupyter


abostroem:
@guyer We decided to change everything to no use the from ... import ... (see #61 ).
I just merged a pull request that fixed a number of these issues (#120) . In light of both of these can you update your changes?

abostroem:
@guyer Do you have a timeline for addressing these comments. I really like the callout to %matplotlib inline and would like to incorporate it into the lessons.

Issue #127, opened 2015-06-12
What changes should we make for our 5.4 release?


gvwilson:
Please use this issue for discussion of changes we should make by mid-August 2015.

embray:
As mentioned on the blog, I would like to see Python 3 on the table.

jgalgarra:
I support the motion, teaching Python 2 does not add any advantage these days

twitwi:
python++

jrherr:
I'm still trying to upgrade myself to Python 3 (not there yet fully...), but I also support this motion.

chendaniely:
Titus uses the `argparse` module in his reproducible computational analysis lesson [1].  Should we use that for our python lesson? or is that adding too much to an already bloated lesson?

You can see the code where he uses `argparse` at the 45:05 minute mark (the link below directly links there).

[1] https://youtu.be/SRItP6PSu4U?t=42m5s

gvwilson:
We can only add things if we identify something to take out...

abostroem:
@chendaniely This was already discussed in #45

katyhuff:
I'd like to suggest that 5.4 could use a gentler introduction to python. If we're being really honest, I never end up using this version of the python lessons because I think it has a fundamental flaw. Namely : this lesson starts with numpy rather than how to start up python or what it means to assign a variable. There's so much that has to be understood before I think students are ready to "import numpy". 

When the audience are novice, I always end up using an [updated version of the old THW material](https://github.com/katyhuff/2014-04-14-wise/blob/gh-pages/advanced/python/py-intro/index.md) or @jkitzes [materials](https://github.com/jkitzes/datasci-lessons/blob/gh-pages/lessons/python/python-full.ipynb). Both of those choose to introduce concepts like variables before introducing the concept of libraries, which I think is extremely important.

I recognize a lot of thought went into these 5.3 lessons, so I'm sure there are reasons for all of the decisions. I also know that the skills I mention are called out as prerequisites in the intro, but I think there should at least be a 00 lesson that demystifies variables, lists, and arrays before importing numpy and loading an array into memory... just to make sure everyone in the room is on the same starting block.

teamneem:
+1 to a gentler introduction.  I use the old version, too.  I feel like it loses students very early on because they don't have a context to build upon.  I usually end up doing the old version and then doing an abridged version of the inflammation stuff

gvwilson:
[...]

PR #129, opened 2015-06-16
Update sample code files to py3k

abought:
## Purpose
Per #127 , we would like to update these lessons to work with python 3. Might as well start somewhere!

## Summary of changes
All example files (mainly used in `10-cmdline.md`) have been updated to work with both Python 2 *and* 3. Sample code inside the markdown file has been updated accordingly. 

These changes were fairly minor, and mainly required using `print` as a function instead of a statement.

Minor changes were also made for PEP-8. The HTML was regenerated from `make preview`.

## Side effects/ Quirks
The sample code embedded in the lessons does not actually include `from __future__ import print_function`, mainly to reduce confusion while teaching the lesson. This statement is used in the runnable code files to ensure that things run with both Python 2 and 3 during the transition period.

No other lesson files were converted in this PR.

### DIFF
--- a/10-cmdline.html
+++ b/10-cmdline.html
@@ -72,14 +72,14 @@
 <h2 id="command-line-arguments">Command-Line Arguments</h2>
 <p>Using the text editor of your choice, save the following in a text file called <code>sys-version.py</code>:</p>
 <pre class="sourceCode python"><code class="sourceCode python"><span class="ch">import</span> sys
-<span class="dt">print</span> <span class="st">&#39;version is&#39;</span>, sys.version</code></pre>
+<span class="dt">print</span>(<span class="st">&#39;version is&#39;</span>, sys.version)</code></pre>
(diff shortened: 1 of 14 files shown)


abought:
Just to bring up to date:

There was talk of creating an official Python 3 branch, to keep changes separate until the entire lesson had been converted. Is that a thing? If so, should I resubmit the PR there?

@gvwilson had proposed that we make the lessons *either* python 3 or 2, rather than target dual compatibility. I'm happy to revise and resubmit accordingly; just let me know what and where. :)

gvwilson:
We now have a `py3k` branch - please start submitting PRs there to 
upgrade the lesson.

r-gaia-cs:
@abostroem Thanks for all the work. This is super seeded by #142.

abostroem:
@r-gaia-cs Just checking that your comment means we can close this PR

r-gaia-cs:
> Just checking that your comment means we can close this PR

@abostroem Yes.

abought:
Great to see the conversion moving along!

PR #131, opened 2015-06-19
Moving motivation to instructors guide

gvwilson:
Also clarifying instructions in CONTRIBUTING.md.


instructors.md
```diff
@@ -3,14 +3,37 @@ layout: page
 title: Programming with Python
 subtitle: Instructor's Guide
 ---
-## Legend
```
abostroem: I would like to see this information moved somewhere rather than deleted.

PR #137, opened 2015-07-06
Make exponentiation loop exercise more explicit

jiffyclub:
The mention of the pow function confuses people so that they think
they are supposed to use the pow function in their solution.
Here I've made the exercise more explicit by specifically telling
the students to use multiplication.
(I don't think the purpose of the exercise is to test whether students
remember that exponentiation is the same as multiplication.)


abostroem:
Agreed. Thanks for the rephrasing!

PR #139, opened 2015-07-07
Fixed #132 where inconsistent function name used

### LINKED ISSUE #132: detect_problems and detectProblems
http://swcarpentry.github.io/python-novice-inflammation/06-func.html

`detect_problems` is defined using an underscore and called using camel case.
Propose to stick with PEP8 rules and make it an underscore.

This is also a bug in the lesson.

arthur-e:


abostroem:
Looks good to me. Thanks for catching that.

PR #140, opened 2015-07-13
Suggested change to explanation of mutable and immutable types

confluence:
This is a suggested rewrite of the callout box which explains mutable and immutable types. I found the previous phrasing confusing because 1) it doesn't emphasize that this is about in-place modification of values, not changing the values of variables, and 2) the last part of the box doesn't adequately motivate why code with in-place modification is more difficult to follow than code without.


abostroem:
Thanks for modifying the language - this is a better discussion

PR #141, opened 2015-07-15
updating 01-numpy.md to python3

kylerbrown:
The only thing that needed to be changed were the print statements.

I did not change them to use the [string formatting mini-language](https://docs.python.org/3.3/library/string.html#format-specification-mini-language), as I think that would introduce  unnecessary complexity.


abostroem:
@r-gaia-cs Just confirming that this PR can be closed.

r-gaia-cs:
> Just confirming that this PR can be closed.

@abostroem Yes.

PR #142, opened 2015-07-15
Convert print statement to print function 

rgaiacs:
This is part of porting the lesson to Python 3.


r-gaia-cs:
This super seeds #141. **This uses `format` method from String**.

andreabedini:
hi @r-gaia-cs are you sure about the decision of using the format method? `print('weight in pounds:', 2.2 * weight_kg)` works perfectly fine and it's way simpler.

kylerbrown:
I feel the `format` method introduces unnecessary complexity. Also, the `format` method was introduced in Python 2.6, so switching to it is somewhat unrelated to moving to Python 3.

r-gaia-cs:
> are you sure about the decision of using the `format` method?

Is my preference.

> `print('weight in pounds:', 2.2 * weight_kg)` works perfectly fine and it's way simpler.

I agree it is simpler. I want to use the `format` method because is probably what people will use if they want complex (and fancy) outputs. I can see myself teaching at a workshop and one of the students asking "And how can I print the value of `2.2 * weight_kg` with three decimal values?" that I can only reply with `format` method and pointing to https://docs.python.org/3/library/string.html#format-specification-mini-language.

andreabedini:
> Is my preference.

can't argue about that :)

> I agree it is simpler. I want to use the format method because is probably what people will use if they want complex (and fancy) outputs. I can see myself teaching at a workshop and one of the students asking "And how can I print the value of 2.2 * weight_kg with three decimal values?" that I can only reply with format method and pointing to https://docs.python.org/3/library/string.html#format-specification-mini-language.

Perhaps there's space for both, we can do the simple way first and then tell the students there's a more powerful way if needed. IMHO using format for very simple statements looks like killing a fly with a bazooka. My 2c anyway.

msarahan:
+1 to keeping it simple (no format),  though I am embarrassed to admit I thought some manner of string formatting (either old style %f or new style {})  were required for even simple tasks,  and didn't know the simpler option until now (thanks!)

HappyPenguin:
Hi - I'm a mailing list lurker but as someone who's only recently "graduated" from SWC student level I'd like to vote for including the format method. I really love putting numbers in text using, for example {:2.2f}. I think it's really intuitive and much easier to read than the % style.

aaren:
Keep it simple to start with - you're just trying to demonstrate simple features of numpy and `.format` distracts from that I think. Later on, make students aware of `.format` and link to the specification.

I think throwing `.format` in at the start introduces too many concepts at once. Maybe it would be better to build it up - show simple things like `'a b c'.split()` or `'word'.upper()` to introduce the idea of simple methods on strings first.

hjelmj:
I like the idea of keeping it simple, but to make the students aware of .format later on. Maybe by including it in an exercise towards the end of one of the first lessons?

benwaugh:
+1 to using the simpler version but mentioning that the `format` method exists. The extra complication of using `format` to start with is likely to interfere with learning of the more important concepts.

snim2:
I would be inclined to use `format()`. Whilst it is good to keep things simple for learners, it is also important to think about how they will cope after SWC when they start reading code and documentation without an instructor in the room. `format()` is very common in Python 3 and I think it would do the learners a disservice if they hadn't seen it at all during the workshops. At the very least they need to know that `format()` exists, that it takes a number of potentially complex formatting instructions, and they need to know where in the documentation those formatting instructions can be found (for future reference).

jlant:
+1 to including `format` towards the end of one of the first lessons.  I think adding it as an exercise would be good.

uiuc-cse:
+1 to @andreabedini's suggestion of delaying `format` to the end.  It can be trivially introduced for advanced or C-savvy audiences but introduces an unnecessary mental load for those unaccustomed to string formatting.

PBarmby:
The consensus above seems reasonable to me: don't introduce `format` right away to beginners, but make sure it gets seen at some point.

jiffyclub:
I'd suggest not using `.format` initially, and possibly not at all unless a meaningful reason can be found for introducing it. The string formatting mini-language is powerful but complex (like regex), and that makes it potentially quite a significant detour.

uiuc-cse:
So this is a case where a micro-lesson may be appropriate, perhaps similar
to the shell-extras collection.

Sincerely,
Neal Davis, PhD
Training Coördinator
*Computational Science and Engineering*

*2102B National Center for Supercomputing Applications*
*1205 W Clark St, MC-257 · **Urbana, IL  61801*
217·244·4181

r-gaia-cs:
Thanks to all for the comments. I will remove the `format` method soon.

r-gaia-cs:
@abostroem Could you review this pull request? I dropped the `format` method.

02-loop.md
```diff
@@ -114,7 +114,7 @@ The improved version of `print_characters` uses a [for loop](reference.html#for-
 to repeat an operation---in this case, printing---once for each thing in a collection.
 The general form of a loop is:
 
-~~~ {.python}
+~~~
```
abostroem: Why did the {.python} disappear in line 117?

abought:
I agree with the basic points about format method (mention in followup, but don't lead with it).

That said: Software Carpentry does have a repo for an intermediate python lesson- is it actively being worked on? A lot of programming lessons focus on either novices or the very advanced, but teaching `str.format` in more detail seems like ideal content for people looking to improve. If we do our jobs right, that segment of the population should grow, and it would be a shame to neglect them. :)

https://github.com/swcarpentry/python-intermediate-mosquitoes

06-func.md
```diff
@@ -52,136 +52,16 @@ Let's try running our function.
...
 
 ~~~ {.python}
-print 'freezing point of water:', fahr_to_kelvin(32)
-print 'boiling point of water:', fahr_to_kelvin(212)
+print('freezing point of water:', fahr_to_kelvin(32))
+print('boiling point of water:', fahr_to_kelvin(212))
 ~~~
 ~~~ {.output}
 freezing point of water: 273.15
-boiling point of water: 273.15
```
abostroem: Was this change intentional?

06-func.md
```diff
@@ -52,136 +52,16 @@ Let's try running our function.
...
+boiling point of water: 373.15
 ~~~
```
abostroem: I think we should introduce a different bug so that we can still have the discussion of debugging. I alway try to keep in mind that whatever the quirks of a language - we are really using it as a vehicle for teaching programming skills - and debugging is one of them

06-func.md
```diff
@@ -522,13 +402,13 @@ The example below shows how Python matches values to parameters:
 
 ~~~ {.python}
 def display(a=1, b=2, c=3):
-    print 'a:', a, 'b:', b, 'c:', c
+    print('a: {} b: {} c: {}'.format(a, b, c))
```
abostroem: Remove format statement?

06-func.md
```diff
@@ -682,7 +562,9 @@ the second parameter in the list.
 > A call to your function should look like this:
 >
 > ~~~ {.python}
-> print fence('name', '*')
+> print(fence('name', '*')))
```
abostroem: Extra )

06-func.md
```diff
@@ -696,7 +578,9 @@ the second parameter in the list.
 > A call to your function should look like this:
 >
 > ~~~ {.python}
-> print outer('helium')
+> print(outer('helium')))
```
abostroem: extra )

06-func.md
```diff
@@ -738,6 +622,6 @@ the second parameter in the list.
 > f2k(41)
 > f2k(32)
 >
-> print k
+> print(k))
```
abostroem: extra )

abostroem:
@r-gaia-cs This is a great start! I've made a few minor comments in line. The biggest thing is whether we should remove the section in 06-func.md which discusses debugging or replace it with a different bug. Comments welcome.

02-loop.md
r-gaia-cs: As far as I remember we are using `{.python}` for code that learners could type and will actually work when following the lesson. Since this isn't the case I drop it.

If I missed something about how we use `{.python}` sorry about it. If the documentation isn't good enough I will improve it.

06-func.md
r-gaia-cs: > I think we should introduce a different bug so that we can still have the discussion of debugging.

I totally agree on this. I just don't know a good bug to introduce right now and for sake of consistence of the lesson the easy solution is drop this part. Since this will not goes live right now I will request to create one issue for this and doesn't block the merge because of it.

> I alway try to keep in mind that whatever the quirks of a language - we are really using it as a vehicle for teaching programming skills - and debugging is one of them

We think in the same way. =)

r-gaia-cs:
> I've made a few minor comments in line.

Addressed.

> The biggest thing is whether we should remove the section in 06-func.md which discusses debugging or replace it with a different bug.

As I said before, I will ask to not block this pull request because of it since this pull request is against `py3k` branch so it won't go live now. Let create a issue for find a good new bug and keep the working going.

02-loop.md
abostroem: I'd like to make this PR only Python 3 changes and make this change as a separate PR

06-func.md
abostroem: Let's put a comment in saying that in earlier version of python this returns a truncated integer and see how the lesson does without the bug lesson here.

abostroem:
@rgaiacs I've added two notes to the inline comments - once those are taken care of we're ready to merge :)

rgaiacs:
> 've added two notes to the inline comments - once those are taken care of we're ready to merge :)

@abostroem I toke care of the comments. Could you review the call out box that I add?

abostroem:
@rgaiacs I think you accidentally committed the html files. Is it possible for you to have this PR only update the md files (I'll regenerate the html files after I merge the PR)?

05-cond.md
```diff
@@ -199,12 +199,18 @@ and false statements. However, they aren't the only values in Python that are tr
 > (Note that if the body of a conditional is a single statement, we can write it on the same line as the `if`.)
```
QuLogic: None of the statements below do this any more (although I wouldn't encourage it anyway.)

06-func.md
```diff
@@ -52,136 +52,64 @@ Let's try running our function.
...
-print 'freezing point of water:', fahr_to_kelvin(32)
-print 'boiling point of water:', fahr_to_kelvin(212)
-~~~
-~~~ {.output}
-freezing point of water: 273.15
-boiling point of water: 373.15
-~~~
+> ## Integer division {.callout}
+>
+> We are using Python3 and integer division always return float points.
```
QuLogic: float points -> floating point.

06-func.md
```diff
@@ -52,136 +52,64 @@ Let's try running our function.
...
+> **Unfortunately**, in Python2 this isn't the case.
+>
+> ~~~ {.bash}
+> $ python -c "print(5/9)"
+> ~~~
+> ~~~ {.output}
+> 0
+> ~~~
+>
+> If you are using Python2 and want a float point as return of division
```
QuLogic: float -> floating. (Though is "floating point" actually defined in this lesson at all?)

rgaiacs:
@QuLogic Thanks for the review.

@abostroem I rebased this pull request so that it could be merge smoothly.

tbekolay:
I'm not sure why Github thinks that d3123318889ad and e9413cb19ac5fc are part of this PR, but locally the history looks correct to me, so I'm going to merge this. Thanks for the rebase!

PR #144, opened 2015-07-29
Instructor solutions update

PBarmby:
This replaces PR #99. I think all of the changes requested by @abostroem have been made. Since the discussion about where to put the solutions is still ongoing (see [lesson-template#267](https://github.com/swcarpentry/lesson-template/issues/267)) I haven't done anything about that.

I hope to get to the solutions for 10-cmdline at some point.

PR #146, opened 2015-08-05
Instruct solns cmdline

PBarmby:
Initial draft of solutions for 10-cmdline. Doubtless could use improvements, but I was aiming for easy-to-understand, not elegant.


abostroem:
@PBarmby These look great. Do you think we should reference the code (I'm specifically thinking of readings-07 and readings-08) or just leave it as is?

abostroem:
How were you picturing merging this with #144 ? Should I merge #144 and then this directly after?

PBarmby:
Yeah, makes sense to reference the code files in the instructor solution file. I'll do that for readings-07 and readings-08  and make sure it's there for all of the other ones.
Re: order of this merge and #144. Shouldn't matter since they are from different branches of mine; although both modify instructors.md and add things to code/, there shouldn't be any conflicts.

PBarmby:
OK, above changes done, ready to go.

PR #148, opened 2015-08-12
Added single line about libraries for instructor training

bjoyce3:
Small commit to finish instructor training for Blake Joyce (University of Arizona). Fixed commit errors with spacing.


abostroem:
This is great - thanks!

PR #150, opened 2015-08-14
Changed all print statments to print function calls

jameshiebert:
The first big difference between Python 2 and 3 is that [print is now a function](https://docs.python.org/3.0/whatsnew/3.0.html#print-is-a-function). This is a really easy change to make (emacs's M-x replace-regexp is my new friend), but one has to make it nearly *everywhere*, so this is a big diff.

I *think* that I found all of the occurrences, but feedback is welcome.


jameshiebert:
Looks like this was already done in the unmerged #142. Nevermind.

PR #151, opened 2015-08-22
enable plot to show

mroswell:


mroswell:
Actually, probably better just to change the import to:
import matplotlib.pyplot as plt

abostroem:
Thanks for this PR. After a long community discussion, we decided not to introduce aliasing, so your fix is perfect!

PR #153, opened 2015-08-29
Minor updates for Python 3 compatibility

tbekolay:
Note that this is a PR on the `py3k` branch.

I did a runthrough of the Python 3 lessons after merging @rgaiacs's PR to clean up any loose ends. After making these changes, the lesson runs properly for me in Python 3.

Often, these changes are just noticing when Python 3 gives different outputs than Python 2 and updating accordingly, and a few spots where we still had print statement (these were added after #142 was made).

One odd thing is that in Python 3, the list returned by `glob` isn't sorted. I'm not sure if this is also true in Python 2, but to be sure that the output is correct, I added `sorted` calls.

Another somewhat large change is that the exception classes are changed in Python 3; importantly, there's no longer a base IOError. Instead, the specific IOErrors are separated out. Updated that lesson accordingly.

Things I noticed but didn't change:

1. I found many floats that truncated differently in the lessons (e.g., 126.5) that truncated differently in my Python 3 environment (e.g., 126.50000000000001). This might be something specific to my machine though, not necessarily Python 3?
2. In 02-loop, the challenge "From 1 to N" mentions that range(3) produces [0, 1, 2]. In Python 3 it instead produces an object, which prints as range(0, 3). You can still iterate over it though, so it doesn't directly affect the challenge, but the pedant in me has a "well, actually" moment when I read the challenge.
3. In Lesson 6, the difference in standard deviations before and after is exactly zero on my machine. I'm not sure if this is, again, a machine difference, or if it's a Python 3 difference.

### DIFF
--- a/04-files.md
+++ b/04-files.md
@@ -21,10 +21,11 @@ that finds files whose names match a pattern.
 We provide those patterns as strings:
 the character `*` matches zero or more characters,
 while `?` matches any one character.
-We can use this to get the names of all the html files:
+We can use this to get the names of all the html files.
+We use the `sorted` function to ensure this list is in alphabetical order:
 
 ~~~ {.python}
-print(glob.glob('*.html'))
+print(sorted(glob.glob('*.html')))
 ~~~
 
 ~~~ {.output}
@@ -43,7 +44,7 @@ Let's test it by analyzing the first three files in the list:
 import numpy
 import matplotlib.pyplot
 
-filenames = glob.glob('*.csv')
+filenames = sorted(glob.glob('*.csv'))
 filenames = filenames[0:3]
 for f in filenames:
     print(f)
--- a/06-func.md
+++ b/06-func.md
@@ -65,47 +65,47 @@ and we have access to the value that we returned.
 
 > ## Integer division {.callout}
 >
-> We are using Python3 and integer division always return floating point.
+> We are using Python 3 and integer division always return floating point.
 >
 > ~~~ {.python}
-> $ python -c "print(5/9)"
+> $ python3 -c "print(5/9)"
 > ~~~
 > ~~~ {.output}
 > 0.5555555555555556
 > ~~~
 >
-> **Unfortunately**, in Python2 this isn't the case.
+> **Unfortunately**, in Python 2 this isn't the case.
 >
 > ~~~ {.bash}
-> $ python -c "print(5/9)"
+> $ python2 -c "print(5/9)"
 > ~~~
 > ~~~ {.output}
 > 0
 > ~~~
 >
-> If you are using Python2 and want a floating point as return of division
-> you need to use one of the following forms
+> If you are using Python 2 and want a floating point as return of division
+> you need to use one of the following forms:
 >
 > ~~~ {.bash}
-> $ python -c "print(float(5)/9)"
+> $ python2 -c "print(float(5)/9)"
 > ~~~
 > ~~~ {.output}
 > 0.555555555556
 > ~~~
 > ~~~ {.bash}
-> $ python -c "print(5/float(9))"
+> $ python2 -c "print(5/float(9))"
 > ~~~
 > ~~~ {.output}
 > 0.555555555556
 > ~~~
 > ~~~ {.bash}
-> $ python -c "print(5.0/9)"
+> $ python2 -c "print(5.0/9)"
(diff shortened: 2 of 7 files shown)


rgaiacs:
@tbekolay Thanks for the pull request. Looks great to me.

> In 02-loop, the challenge "From 1 to N" mentions that range(3) produces [0, 1, 2]. In Python 3 it instead produces an object, which prints as range(0, 3). You can still iterate over it though, so it doesn't directly affect the challenge, but the pedant in me has a "well, actually" moment when I read the challenge.

That is true. At the begin Python only had lists so `range` returned a list. Later was introduced [generators](https://www.python.org/dev/peps/pep-0255/) to help improve Python performance. For backward compatibility, in Python 2 users have `range` that return a list and `xrange` that return a "generator". But in Python 3, `xrange` was renamed to `range` and if you need a list you need to write `list(range(0, 3))`.

If I made a mistake, please correct me. =)

tbekolay:
> if you need a list you need to write list(range(0, 3))

Right yeah. Do you think it makes sense to include the `list` call in that challenge? Personally I think it makes things more complicated for little benefit, but we also don't want to mislead students.

rgaiacs:
> > if you need a list you need to write list(range(0, 3))
>
> Right yeah. Do you think it makes sense to include the `list` call in that challenge?

My first thought is yes but I agree with you when you say

> Personally I think it makes things more complicated for little benefit, but we also don't want to mislead students.

04-files.md
```diff
@@ -21,10 +21,11 @@ that finds files whose names match a pattern.
 We provide those patterns as strings:
 the character `*` matches zero or more characters,
 while `?` matches any one character.
-We can use this to get the names of all the html files:
+We can use this to get the names of all the html files.
+We use the `sorted` function to ensure this list is in alphabetical order:
```
gvwilson: Please modify to show `glob.glob` first (with items not sorted), then introduce `sorted` in a callout box with a note saying "you should only bother to sort if you really care, and you usually only really care as the last step of processing".

04-files.md
```diff
@@ -43,7 +44,7 @@ Let's test it by analyzing the first three files in the list:
 import numpy
 import matplotlib.pyplot
 
-filenames = glob.glob('*.csv')
+filenames = sorted(glob.glob('*.csv'))
```
gvwilson: As above, remove `sorted`.

06-func.md
```diff
@@ -65,47 +65,47 @@ and we have access to the value that we returned.
...
 > ~~~ {.python}
-> $ python -c "print(5/9)"
+> $ python3 -c "print(5/9)"
 > ~~~
 > ~~~ {.output}
 > 0.5555555555555556
 > ~~~
 >
-> **Unfortunately**, in Python2 this isn't the case.
+> **Unfortunately**, in Python 2 this isn't the case.
```
gvwilson: s/isn't/wasn't/ please

06-func.md
```diff
@@ -65,47 +65,47 @@ and we have access to the value that we returned.
...
 > ~~~
 > ~~~ {.output}
 > 0.555555555556
 > ~~~
 > ~~~ {.bash}
-> $ python -c "print(5/9.0)"
+> $ python2 -c "print(5/9.0)"
 > ~~~
 > ~~~ {.output}
 > 0.555555555556
```
gvwilson: ...and then a callout box: if you want an integer result from 4/2, you use `4//2`.

instructors.md
```diff
@@ -158,30 +165,35 @@ Solutions to exercises:
...
+>
+> C gets printed, because the first two conditions, `4 > 5` and `4 == 5` are not true, but `4 < 5` is true.
+>
 
 
 > ## What is truth? {.challenge}
 >
 > After reading and running the code below,
 > explain the rules for which values are considered true and which are considered false.
-> (Note that if the body of a conditional is a single statement, we can write it on the same line as the `if`.)
```
gvwilson: Why take this out? Seems to be true in Py3k

rgaiacs: This was me.

> Why take this out?

Because from my understanding we want to teach best practices to students and based on that I tough that don't mention this has more positive effects than negative ones.

gvwilson: OK - let's take it out.

PR #154, opened 2015-08-29
Many small changes from runthrough

tbekolay:
In going through the Python 3 runthrough (see #153) I saw a bunch of little things that I felt should be changed.

Note that this branch is based off of #153 (i.e., it picks up where that PR leaves off -- it assumes that that PR has already been merged) so I recommend that we deal with #153 first, and even merge `py3k` into `gh-pages` before looking at this PR (it'll make merging much simpler).

I also think that we should ensure that these fixes aren't already done in other PRs. I was just blindly going through and changing things I saw that needed changing, but some of them struck me as things I've seen fly through my email at one time or another, and I definitely would rather an earlier PR be merged before looking at this one.

The history of this PR will be easier to read once those other things are merged, but if you want to check out the changes now, look at the Commits tab. This PR is essentially all of the commits from 20923d8c9a5770a9 onwards.

### DIFF
--- a/instructors.md
+++ b/instructors.md
@@ -368,8 +368,8 @@ Solutions to exercises:
 > k = 0
 >
 > def f2k(f):
->   k = ((f-32)*(5.0/9.0)) + 273.15
->   return k
+>     k = ((f-32)*(5.0/9.0)) + 273.15
+>     return k
 >
 > print(f2k(8))
 > print(f2k(41))
--- a/instructors.md
+++ b/instructors.md
@@ -95,11 +95,13 @@ Solutions to exercises:
 
 > ## From 1 to N {.challenge}
 > Using `range`,
-> write a loop that uses `range` to print the first 3 natural numbers:
+> write a loop that uses `range` to print the first 3 natural numbers.
 >
-> ~~~ {.output}
-> for i in range(1,4):
+> ~~~ {.python}
+> for i in range(1, 4):
 >    print(i)
+> ~~~
+> ~~~ {.output}
 > 1
 > 2
 > 3
@@ -123,7 +125,8 @@ Solutions to exercises:
 >
 > Write a loop that takes a string,
 > and produces a new string with the characters in reverse order.
-> ~~~ {.output}
+>
+> ~~~ {.python}
 > newstring = ''
 > oldstring = 'Newton'
 > length_old = len(oldstring)
@@ -140,11 +143,12 @@ After discussing these challenges could be a good time to introduce the `b*=2` s
 ## [Storing Multiple Values in Lists](03-lists.html)
 
 Solutions to exercises:
->## Turn a string into a list {.challenge}
+
+> ## Turn a string into a list {.challenge}
 >
-> Use a for-loop to convert the string "hello" into a list of letters:
+> Use a `for` loop to convert the string `"hello"` into a list of letters:
 >
-> ~~~ {.output}
+> ~~~ {.python}
 > my_list = []
 > for char in "hello":
 >     my_list.append(char)
@@ -154,12 +158,12 @@ Solutions to exercises:
 > ["h", "e", "l", "l", "o"]
 > ~~~
 
-
 ## [Analyzing Data from Multiple Files](04-files.html)
 
 ## [Making Choices](05-cond.html)
 
 Solutions to exercises:
+
 > ## How many paths? {.challenge}
 >
 > Which of the following would be printed if you were to run this code? Why did you pick this answer?
@@ -209,7 +213,8 @@ Solutions to exercises:
 >
 > Write some conditions that print `True` if the variable `a` is within 10% of the variable `b`
 > and `False` otherwise.
-> ~~~ {.output}
+>
+> ~~~ {.python}
 > a = 5
 > b = 5.1
 >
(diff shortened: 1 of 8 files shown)


tbekolay:
With #153 and #155 merged, this is now ready for review. I've rebased it, so only the appropriate commits show up.

05-cond.md
```diff
@@ -86,7 +86,7 @@ We can also combine tests using `and` and `or`.
 `and` is only true if both parts are true:
 
 ~~~ {.python}
-if (1 > 0) and (-1 > 0):
+if 1 > 0 and -1 > 0:
```
abostroem: @tbekolay why did you remove the parentheses? I think they improve readability even if they are not strictly necessary.

05-cond.md
```diff
@@ -98,7 +98,7 @@ one part is not true
 while `or` is true if at least one part is true:
 
 ~~~ {.python}
-if (1 < 0) or (-1 < 0):
+if 1 < 0 or -1 < 0:
```
abostroem: Same as line 89

instructors.md
```diff
@@ -529,36 +551,35 @@ Solutions to exercises:
...
 > ~~~
-> ~~~ {.output}
-> The first assertion checks that the input sequence `values` is not empty.
->   An empty sequence such as `[]` will make it fail.
-> The second assertion checks that the first value in the list is positive.
->   Input such as `[-1,0,2,3]` will make it fail.
-> The third assertion checks that the running total always increases.
->   Input such as `[0,1,3,-5,4]` will make it fail.
-> ~~~
+>
```
abostroem: I think we were using .output to indicate that this was the solution. Do you think we need some kind of visual cue to separate the question from the answer?

instructors.md
```diff
@@ -570,51 +591,50 @@ Solutions to exercises:
...
 > ~~~ {.output}
 > -1
 > ~~~
 >
-
-> ~~~ {.output}
+> ~~~ {.python}
 > # this is code/arith.py
->import sys
+> import sys
```
abostroem: Why the space in front of sys?

instructors.md
```diff
@@ -570,51 +591,50 @@ Solutions to exercises:
...
 >
->def main():
->    assert len(sys.argv) == 4, 'Need exactly 3 arguments'
+> def main():
```
abostroem: again, why the space?

instructors.md
```diff
@@ -570,51 +591,50 @@ Solutions to exercises:
...
+>         operand1, operand2 = float(sys.argv[2]), float(sys.argv[3])
+>     except ValueError:
+>         print('cannot convert input to a number: bailing out')
+>         return
 >
->    do_arithmetic(operand1, operator, operand2)
+>     do_arithmetic(operand1, operator, operand2)
 >
->def do_arithmetic(operand1, operator, operand2):
+> def do_arithmetic(operand1, operator, operand2):
```
abostroem: Space?

instructors.md
```diff
@@ -570,51 +591,50 @@ Solutions to exercises:
...
+>     elif operator == 'subtract':
+>         value = operand1 - operand2
+>     elif operator == 'multiply':
+>         value = operand1 * operand2
+>     elif operator == 'divide':
+>         value = operand1 / operand2
+>     print(value)
 >
->main()
+> main()
```
abostroem: space?

instructors.md
```diff
@@ -631,63 +651,62 @@ Solutions to exercises:
...
->    for item in glob_output: # print the output
->        print(item)
->    return
+> ~~~ {.python}
+> # this is code/my_ls.py
+> import sys
+> import glob
 >
->main()
+> def main():
```
abostroem: space?

abostroem:
@tbekolay these changes mostly look good (and will clean up the look of the page considerably). A few minor inline comments then we can merge.

05-cond.md
tbekolay: Pedantically, I think that we should avoid unnecessary operators because we're implicitly teaching that they *are* necessary. Practically, there are many languages that require parentheses in if statements (`if (expression)`) and so I want to reinforce that this isn't necessary  (and, to me, looks really weird) in Python.

But, I don't feel that strongly about it so I'm happy to add them back in; anyone case to break the tie here?

instructors.md
tbekolay: I think my preference would probably be to remove the question, just to reduce duplication, as when we make minor changes to the questions we're unlikely to update them here as well. Mainly, I changed it because the `{.output}` also makes it a `pre` block, so things like the backticks here don't make a difference, and everything is rendered in a monospaced font.

I think ideally, we would implement a `{.answer}` style, which could even hide the answer until a click or something like that. For the time being, I'll add in a horizontal rule to at least give some visual clue that the answer's started.

instructors.md
tbekolay: I always thought the space between the `>` and the rest of the line was necessary in Markdown? It's certainly [there in the spec](https://daringfireball.net/projects/markdown/syntax#blockquote). I suspect if the first line didn't have the extra space, the formatting would be all messed up... but also, it's inconsistent to have a space in the first line but not subsequent lines.

05-cond.md
tpoisot: I've seen people get confused when there are no parentheses: `1 > 0 and -1` can be parsed as `1 > (0 and -1)` -- unless you've already read the whole line. I tend to use parentheses because they improve *reading* code, which I think is worth emphasizing.

If we want to not use the parentheses, maybe there needs to be something on the precedence of operators, and I think this lesson is dense already.

I think the parentheses should stay.

tbekolay:
I removed the contentious change with the parentheses around if conditions, but left in the spaces in the blockquotes because it's consistent with how we do it everywhere else in the lesson. I added horizontal lines to separate the question from the answer in the instructor guide (see 41e24ef320824a249bd0c20e). The rest is as it was before. Would be good to get this merged soon, as we're starting to get an influx of PRs that are duplicating this effort.

abostroem:
I merged this, but when I run make preview  and do a git status only 07, 10, and discussion show up as different. Any ideas what's going on? (if you have the html generated feel free to push them)

tbekolay:
Not sure how make determines newness, but you can do `make --always-make preview` to force it to regenerate all of them. Thanks for merging :)

PR #156, opened 2015-08-31
Fixed Markdown of instructors.md

rgaiacs:


tbekolay:
Hi @rgaiacs; I think I did this in #154. Not sure if we made any different choices here...

PR #157, opened 2015-08-31
Fixing print statements in instructors.md

gvwilson:
Solutions to programming exercises in `instructors.md` were still using Python 2 `print` syntax.


tbekolay:
These are part of #154 also (was too lazy to separate out the general changes from the Python 3 specific ones in `instructors.md` so I'll rebase that again.

Issue #159, opened 2015-09-10
Should lesson code adhere to PEP8?


petebachant:
Code like `weight_lb = 2.2 * weight_kg` should be `weight_lb = 2.2*weight_kg` according to [PEP8](https://www.python.org/dev/peps/pep-0008/). Does this matter here?

gvwilson:
+1 to PEP-8 compliance.

tbekolay:
I'm all for PEP8 compliance (vigilant about it, even!) but I don't think your example follows PEP8. I'm assuming you're thinking about [this section](https://www.python.org/dev/peps/pep-0008/#other-recommendations). Having no whitespace around `*` only applies when you have a longer expression with implicit order of operations (removing the space is to make the order of operations explicit, even though it doesn't make a different to the interpreter). I think in general PEP8 favours more whitespace than less whitespace; it also favours not rigidly adhering to PEP8 if it makes code less readable. `2.2*weight_kg` is less readable to my eyes.

petebachant:
Looks like I understood PEP8 wrong; sorry about that! My example does not fail the `pep8` test script in either incarnation, though I usually prefer no whitespace around `*` to make it easier to distinguish addition from multiplication. 

Maybe a compliance check could be automated by extracting all code from the lessons and running that through `pep8`? Is there a tool to convert Markdown to Python in this way, commenting out the text so line numbers are preserved?

tbekolay:
Such a check would be great, and could be incorporated into existing checks. I'm not aware of any tool out there to do this right now; if there were, it'd probably have to be customized anyhow since our code blocks are fenced off with `~~~ {.python}`, which I've not seen elsewhere. Still, seems like not too tough of a script to write!

petebachant:
Alright, I will take a crack at it when I get a chance.

franktoffel:
@tbekolay @petebachant 

This small script may help:
https://gist.github.com/Juanlu001/9082229

Are the python lessons in IPYNB format?

PR #160, opened 2015-09-11
Add script to test PEP8 compliance

petebachant:
From #159. So far, this may not be ready since it catches some intentional false positives, but it is catching quite a few that appear to be valid.


petebachant:
Also thinking that this may belong in the lesson template instead--maybe with an automated way to detect lesson pages to test.

petebachant:
Got this down to a single script. Not sure why `subprocess.call` wasn't working before, but it's okay now. May still need some tweaking to filter out more non-errors though.

petebachant:
I just submitted PR swcarpentry/lesson-template#283, which is a refined version of this one, so I'm going to close this.

PR #162, opened 2015-09-15
Update 01-numpy.md

alistairwalsh:
In the challenge 'Sorting out references' multiple assignment is used without any previous mention. I have added the use of assigning multiple values to multiple variables in the numpy methods section. I understand that this is introducing two new concepts at the same time (methods and multiple assignment). 

It does allow the student to chunk the three descriptive statistics though so is might lower the load on working memory
The other options are to: 
1. Introduce simultaneous assignment in it's own section, either in this lesson or move it elsewhere and move the challenge. (This lesson is already large)
2. There is also the option of removing the multiple assignment from the challenge and assigning each variable separately. (Multiple assignment is a neat feature of Python)

### DIFF
--- a/01-numpy.md
+++ b/01-numpy.md
@@ -390,12 +390,16 @@ to tell Python to go and do something for us. `data.shape` doesn't
 need `()` because it is just a description but `data.mean()` requires the `()`
 because it is an action.
 
-NumPy arrays have lots of useful methods:
+NumPy arrays have lots of useful methods, lets use three of those methods to get some descriptive values about the dataset.
 
 ~~~ {.python}
-print('maximum inflammation:', data.max())
-print('minimum inflammation:', data.min())
-print('standard deviation:', data.std())
+# we can assign multiple values to multiple variables at once
+maxval, minval, stdval = data.max(), data.min(), data.std()
+	
+	
+print('maximum inflammation:', maxval)
+print('minimum inflammation:', minval)
+print('standard deviation:', stdval)
 ~~~
 ~~~ {.output}
 maximum inflammation: 20.0


abostroem:
Surprisingly (to me) I'm ok with this change, because multiple assignment is used often and we expect it for the exercise, but I would like to hear other people's opinions? All comments welcome. @gvwilson @jiffyclub @tbekolay

tbekolay:
I don't have a problem with assigning to multiple things with tuples, but I don't quite see the motivation for introducing this syntax at this point specifically. If we think people will be confused about it in that exercise, I would think a callout or some explicit text about it would be in order, rather than adding it to an inline code snippet; in a sense, I think it breaks the visual flow of that portion of [the lesson](http://swcarpentry.github.io/python-novice-inflammation/01-numpy.html) (from a certain point onwards, all the code examples just print some small snippet, which is what this code example also does).

That's just my opinion though; if it could be added somewhere where it would save a line instead of introduce additional lines (which is typically why multiple assignment is often used) I would probably react more positively. Also, if you do go ahead with merging this, there are a few minor style issues that should be fixed before merging: double newlines in the code example, and the text change is missing an apostrophe (and should be two separate sentences, IMO).

alistairwalsh:
@tbekolay 
Rock and a hard place. The lessons are already too long and I don't want to add any length but after presenting this material it was obvious at that point what I was asking the learners to do had not been explained. 

The change I proposed is a bandaid.

Apostrophe accepted though. :-)

tpoisot:
I think this is important to introduce this particular bit of pythonic syntax.

**But** maybe it can go in the `list` parts of the lesson? Currently, the lesson only talks about lists, and not useful data structures (tuples and dicts). I agree that it is a long lesson, but maybe we can use this bit of syntactic sugar to introduce the fact that other data structures exist, and students can look into them?

valentina-s:
I suggest introducing multiple assignment in the function lesson as a way to return several variables. As for fixing this exercise, I am ok with assigning all first, second, ... variables individually.

tbekolay:
I agree that this material would be best in another place, but since we have this material let's merge it for now; we can always move it later. I'll fix the apostrophe and move the multiple assignment comment to the main text in the merge.

PR #163, opened 2015-09-19
Cleaned up some sentences

jladams:
These are my small fixes to a lesson as part of instructor training. It's a couple of small stylistic/grammatical things, nothing too major. I initially missed the instruction to only edit the Markdown files and instead made edits to the HTML, hence the multiple commits.


abostroem:
Thanks for the clarifying changes

PR #164, opened 2015-09-21
Fix incorrect statement about range

vahtras:


abostroem:
Thank you, these little typos can make a big difference to student's level of confusion!

PR #165, opened 2015-09-28
range(3) produces [0,1,2], not [1,2,3]

jpallen:
(This was pointed out by a student in a lesson today).


jpallen:
This is a dup of #164. Should have looked first, sorry.

PR #166, opened 2015-09-29
Insert empty line to restore bullet point list formatting.

hlapp:
I'm not fully sure, but it appears that you have to have an empty line preceding and following an enumeration to have it properly convert in Jekyll. (This is slightly different from Github's Markdown rendering.)


abostroem:
Good catch! Thank you

PR #168, opened 2015-10-20
Inflammation data glob

lduchesne:
Since there are also "small*.csv" files in the data directory, the "Analyzing Data from Multiple Files" can fail to show three inflammation data graphs, which is what happens on my computer.

As for the HTML, I'm not sure why the "make preview" changed so much of the original HTML file.

### DIFF
--- a/04-files.html
+++ b/04-files.html
@@ -30,7 +30,7 @@
           <h2 class="subtitle">Analyzing Data from Multiple Files</h2>
           <section class="objectives panel panel-warning">
 <div class="panel-heading">
-<h2><span class="glyphicon glyphicon-certificate"></span>Learning Objectives</h2>
+<h2 id="learning-objectives"><span class="glyphicon glyphicon-certificate"></span>Learning Objectives</h2>
 </div>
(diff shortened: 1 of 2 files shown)


tbekolay:
Thanks for this @lduchesne! You're right, I think we wrote this assuming that `glob` would return files in sorted order, but that doesn't seem to be the case now (if it was every the case).

We will regenerate the HTML in the process of merging the PR, so no worries about that. Do you want to remove commit 50dbe0a yourself? I can also just omit it in the merge, up to you!

lduchesne:
Just go ahead, I'll let you omit the HTML part.
There's still the problem of file ordering (we might not show inflammation0{1-3}.csv. The wording on that page ("Let’s test it by analyzing the first three files in the list") seems adequate but maybe we should just specify somewhere that the order is unspecified, just so people who have different results understand why :)

tbekolay:
Merged, thanks @lduchesne! Feel free to make another PR sorting the csv list, or mentioning something about the order, if you think it's important.

PR #169, opened 2015-10-23
change data.min to data.max in Checking our Data section of 05-cond.md

richford:
The first conditional statement in "Checking our data" has
```python
if data.min(axis=0)[0] == 0 and data.max(axis=0)[20] == 20:
    print('Suspicious looking maxima')
```

But the sections after "let's test that out" use
```python
if data.max(axis=0)[0] == 0 and data.max(axis=0)[20] == 20:
    print('Suspicious looking maxima')
```
which seems to be the intended test. This pull request changes the first conditional statement to match the other two.


tbekolay:
Right you are, thanks for this change @richford!

PR #172, opened 2015-10-28
Renamed modules with dashes to underscores

zonca:
Cannot import module with dashes.
The lesson text uses underscores.


tbekolay:
We actually have a bunch of files in `code` that have dashes; would you mind changing them all?

zonca:
all the others are scripts, so there is no issue if they have dashes. I recommend to leave dashes on those so we don't have to change the lesson material.

tbekolay:
Fair!

tbekolay:
Thanks for your contribution @zonca!

PR #175, opened 2015-11-09
Update 01-numpy.md

benlaken:
Students in the last session in Oslo were asking how to see what had been declared/loaded already. It is a good command that I often use when working, I think it is easy, and also very nice to know from the start. Showing them the command so early on could help them conceptualise the namespace more effectively.

### DIFF
--- a/01-numpy.md
+++ b/01-numpy.md
@@ -156,18 +156,20 @@ Since `weight_lb` doesn't "remember" where its value came from,
 it isn't automatically updated when `weight_kg` changes.
 This is different from the way spreadsheets work.
 
-You can use the `whos` command at any time to see what variables you have created and what modules you have loaded into your session: 
-
-~~~ {.python}
-whos
-~~~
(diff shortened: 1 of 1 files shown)


tbekolay:
I agree, `whos` is useful and (like `git log`) helps build students' mental model of what they're doing. However, it is not a Python / NumPy command, it is IPython specific. For that reason, I think it should be in a callout box rather than in the main text, as the main text does not assume that you're using IPython, and the callout should mention that `whos` will only work if you're in an IPython terminal, the IPython/Jupyter notebook, etc.

benlaken:
Good idea - I added the {.callout} style (hopefully I got that right?), and the note about the use restriction. I also included a pithy-ish title in line with the style of the other callouts useful for remembering the command, but perhaps the title could be more informative and less pithy.

abostroem:
I'm going to merge this. Thanks for adding it!

PR #176, opened 2015-11-20
Better explanation for base-0 indexing

khinsen:
The argument "that’s simpler for computers to do" is not credible.
Compilers and interpreters can easily handle an offset of one.
The choice of index base is made with human convenience in mind.

### DIFF
--- a/01-numpy.md
+++ b/01-numpy.md
@@ -235,7 +235,7 @@ but `data[0, 0]` might.
 Programming languages like Fortran and MATLAB start counting at 1,
 because that's what human beings have done for thousands of years.
 Languages in the C family (including C++, Java, Perl, and Python) count from 0
-because that's simpler for computers to do.
+because that's more convenient when indices are computed rather than constant.
 As a result,
 if we have an M&times;N array in Python,
 its indices go from 0 to M-1 on the first axis


jiffyclub:
I agree that the existing explanation is not correct, but I don't think novice students will understand your replacement explanation. What about providing no justification at all?

khinsen:
They may not fully understand it, but they get some idea of why this is not just some historical accident.

The point is not very important, since none of the later material depends on an understanding of this explanation. So yes, we might simply say nothing. But my experience is that people like being given an explanation for surprising facts, even if it is unclear, and in fact even if it is wrong.

abostroem:
I agree that the new explanation is confusing. Can you elaborate on the last sentence of your commit message?

khinsen:
@abostroem By "human convenience" I mean that zero-based indexing reduces the amount of "+1" and "-1" in the source code when indices are computed. This also means that there are fewer occasions for introducing a mistake by forgetting +/-1. I can't offer anything but anecdotal evidence for this, unfortunately. Presumably one could do a study on Fortran/Matlab vs. C/Python code to verify this claim.

There is a [short paper by Dijkstra](https://www.cs.utexas.edu/users/EWD/transcriptions/EWD08xx/EWD831.html) on this subject that also argues for zero-based indexing. It refers to experiments with a language called Mesa from Xerox PARC that support this argument. I haven't seen anything more precise on these experiments.

tbekolay:
I think it's hard to get it across in relatively few words, but in the past I have tried to explain why it's easier by referring to centuries.

Most people are either confused or were confused about centuries, as we say that the years 0-99 are century 1, 100-199 are century 2, and so on. Most people, when they hear a year like 1990, look to the leading two digits (**19**90) to determine the century, and through enough reinforcement, they come to learn the rule that you have to add one to that get the actual century (20). If we had instead started with a century 0, then we would no longer have to learn that rule, that many people still have to explicitly remember, as it is not intuitive. Similarly, with computers, it is often handy to be able to be able to just look at the digits to determine the "century" (or whatever you're figuring out), rather than having to remember to add one all the time.

I'm not sure if this is worth adding to the lesson as a callout or what have you; probably it's best to find a nice reference or blog post and link to that. I can make such a blog post if one doesn't already exist.

ketch:
I like @tbekolay's explanation, though I agree that it's counterproductive to spend too much time on this point.  I'm a mathematician and there have been many times when I've appreciated zero-indexing, but I had to think hard to follow Dijkstra's careful justifications.  Note that he justifies zero-indexing *after* and *because* of the fact that one wants to use half-open intervals (i.e., omit the value referenced by the upper index).

QuLogic:
The most straight-forward explanation of 0-based indexing I've found (though I have no idea how well it works for new learners) is that the index does _not_ refer to the item. The index instead refers to the _space between elements_. That is, in the string `'abc'`, the indices are like so:
```
+---+---+---+
| a | b | c |
+---+---+---+
0   1   2   3
```

If you use a single index (`x[0]`), then it picks the value in the box after that index's space. If you use a slice (`x[0:2]`), then it picks the boxes between those two index's spaces.

Though that explanation isn't really a justification for 0-based indexing, it does help show things like the half-open interval.

iglpdc:
> Though that explanation isn't really a justification for 0-based indexing, it does help show things like the half-open interval.

I think (CS people, correct me if I'm wrong) this is the *actual* technical explanation. The index was actually the shift in memory to get the element from the beginning of the data structure. The first element `x[0]` is right there at the beginning of the structure `x`, meaning that `x[0]` and `x` point to the same memory location.

 To get the `i-th` element of `x`, I have to jump over the `i` elements, so `x[i]` points to `x[0] + i * size_elem`, where `size_elem` is the size in memory of the elements stored in the array.

Maybe in this case telling the real thing is a good way to explain it.

iglpdc:
Actually, to advocate a bit for zero-indexing, I think it's what it makes sense also from a mathematical point of view. 

The index labels things *inside* the structure, so it's a relative index for the elements. This means that if I move the structure as a whole somewhere else the indexing stays be the same. 

Given that, it makes sense to use a significant point in the structure as the "origin" of my indexing system and index the remaining elements with the relative position to this significant point. I'm free to use any value to index this significant point, but the most natural thing is set it to zero.

In maths/physics we do this all the time. For example, we set the origin of coordinates of the Solar System to be the center of the Sun (the center of mass of the whole Solar System), so the Sun's position is indexed as `(0, 0, 0)`.

In a Python (ordered) sequence `x` with any possible number of items, the only privileged position is the first element, so it's natural to choose it as our origin of coordinates and label it as  `x[0]`.

gvwilson:
Mike Hoye from Mozilla dug into this a while back, and the *real* 
explanation is stranger than you'd think: 
http://exple.tive.org/blarg/2013/10/22/citation-needed/ (We should put a 
link to this into the lesson...)

tbekolay:
Merged, thanks @khinsen! I also added a commit linking to Mike Hoye's blog post.

PR #178, opened 2015-11-30
Adding call stack section

c-martinez:
In the `Programming with R` lesson, the [`Creating functions` topic](https://swcarpentry.github.io/r-novice-inflammation/02-func-R.html) contains a section about call stack. The python lesson also has a `Creating functions` topic, which does not contain this section. It would be consistent to teach about call stacks in both languages. This pull request adds such section to the Python lesson.

### DIFF
--- a/06-func.md
+++ b/06-func.md
@@ -9,6 +9,7 @@ minutes: 30
 > *   Define a function that takes parameters.
 > *   Return a value from a function.
 > *   Test and debug a function.
+> *   Explain what a call stack is, and trace changes to the call stack as functions are called.
 > *   Set default values for function parameters.
 > *   Explain why we should divide programs into small, single-purpose functions.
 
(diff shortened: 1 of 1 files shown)


abostroem:
This adds a significant new section to the lessons. What part of the Python lesson would you remove in order to include this section?

c-martinez:
I am not suggesting removing any part of the Python lesson. This new section does make the lesson longer, but it is just as long as [its R counterpart](https://swcarpentry.github.io/r-novice-inflammation/02-func-R.html). If adding this requires removing another section of the lesson, then perhaps a section of the R lesson should be removed as well?

tbekolay:
Hi @c-martinez, we actually did have the mirrored content previously; it is still available in the [discussion page](http://swcarpentry.github.io/python-novice-inflammation/discussion.html). In the end, we found that most instructors teaching the lesson skipped this material, and so we chose to remove it from the main lessons, but left it in the discussion in case anyone still wanted to use it. I would be curious to know if those teaching R actually go over this content or skip it frequently as well...

c-martinez:
Hi @tbekolay, ah -- good to know. So perhaps instead of this PR adding a section to the python lesson, I should send a new PR removing the section from the R lesson, to start the discussion ?

tbekolay:
Sure, that sounds like a good way forward :)

tbekolay:
Looks like https://github.com/swcarpentry/r-novice-inflammation/pull/158 has now been resolved, so closing, thanks @c-martinez!

PR #180, opened 2015-12-04
Added callout for way to find methods in IPython

benlaken:
Made a quick note about the object. tab completion feature in IPython and ? help function. I found this to be helpful in the classes in letting them know how they can fend for themselves more quickly with a feature that they might otherwise not find.

### DIFF
--- a/01-numpy.md
+++ b/01-numpy.md
@@ -403,16 +403,16 @@ minimum inflammation: 0.0
 standard deviation: 4.61383319712
 ~~~
 
-> ## Mystery methods {.callout}
+> ## Mystery methods in IPython {.callout}
 >
 > How did we know what methods data has and how to use them? When you are working
-> on your own data it might be something different to a numpy object, and how will
-> you know then what methods you can use then? Well IPython gives us an easy way
-> of finding this out. If you type the name of your object with a full-stop then
-> you can use tab completion (e.g. type `data.` and then press tab) to see a
-> list of all methods that you can use on that object. After selecting one you can
-> also add a question mark (e.g. `data.cumprod?`) and IPython will return an explanation
-> of the method!
+> on your own data it might be something different to a numpy object: how will
+> you know what methods you can use then? If you are working in the IPython/Jupyter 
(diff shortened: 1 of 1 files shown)


tbekolay:
Currently, the lesson is designed to not require the use of IPython / Jupyter. Would it be possible to reword this callout with other methods? E.g., the use of `dir`, or links to online documentation?

benlaken:
Using dir() and help() in a regular old Python shell could roughly give the same effect, but it is much less interactive/elegant. Would it be enough to expand this with a further sentence? Such as: "As this is an IPython command, it only works in an IPython terminal or the Jupyter Notebook."

tbekolay:
We do have one IPython specific callout, so I guess two isn't that different. Could you reword this callout so that it looks similar to the "Some IPython magic" callout on? I.e., starts with "If you’re using an IPython / Jupyter notebook,..." Then perhaps add "IPython" somewhere in the name of the callout, so people not using IPython can skip over it.

01-numpy.md
```diff
@@ -403,6 +403,18 @@ minimum inflammation: 0.0
...
+> ## Mystery methods in IPython {.callout}
+>
+> How did we know what methods data has and how to use them? When you are working
+> on your own data it might be something different to a numpy object: how will
+> you know what methods you can use then? If you are working in the IPython/Jupyter 
+> Notebook there is an easy way to find out. If you type the name of your object
+> with a full-stop then you can use tab completion (e.g. type `data.` and then press tab)
+> to see a list of all methods that you can use on that object. After selecting one you
+> can also add a question mark (e.g. `data.cumprod?`) and IPython will return an 
+> explanation of the method!
```
QuLogic: Specifically, the explanation is the output of `help(data.cumprod)`. Maybe it would be good to mention that as it would tie in with the section on writing docstrings.

valentina-s:
This looks good, thanks! I will merge and remove the reference to docstring (as they have not learned about creating functions and docstrings at this point).

PR #184, opened 2016-01-04
Slicing Exercise (for instructor training checkout)

tobyhodges:
…e to confirm the learner's understanding of slicing and negative indexing. Follow-up exercises could be written to further test understanding of step sizes and empty start/stop positions.


03-lists.md
```diff
@@ -167,15 +167,49 @@ odds: [1, 3, 5, 7]
...
+> ~~~ {.python}
+> ["h", "e", "l", "l", "o"]
+> ~~~
+> Hint: You can create an empty list like this:
+
+> ~~~ {.python}
+> my_list = []
+> ~~~
+
+Subsets of lists and strings can be accessed by specifying ranges of values in brackets, similar to how we accessed ranges of positions in a Numpy matrix. This is commonly referred to as "slicing" the list/string.
```
tbekolay: Currently, our convention is to have all of the lesson material first, then all of the challenges at the end. Could you move this new material up so that it appears before the challenges?

Also (this is much more nit-picky), could you add some line breaks here? We currently don't do this very consistently, but in most of our `.md` files, we have newlines either after a certain number of characters, or when it makes sense for the sentence (see [this blog post](http://rhodesmill.org/brandon/2012/one-sentence-per-line/) for details). The main reason for this is so that if someone makes a change to this paragraph later on, it will be more obvious what has changed because diffs are done on a line-by-line basis.

03-lists.md
```diff
@@ -167,15 +167,49 @@ odds: [1, 3, 5, 7]
...
 
 ~~~ {.python}
-my_list = []
+date = "Monday 4 January 2016"
+months = ["jan", "feb", "mar", "apr", "may", "jun", "jul", "aug", "sep", "oct", "nov", "dec"]
+day = date[0:6]
+year = date[17:21]
+summer = months[5:8]
+last = months[-1]
+print("day", day)
```
tbekolay: The output right now was a bit confusing on first glance; I think it would help visually to add a colon in each print statement (i.e., `print("day:", day)`, without output `day: Monday`).

tbekolay:
I made a few inline comments about the new material, but in general it looks great! I like the challenge question; it's straightforward, and only tests one thing, which is perfect for this lesson. Would you mind also adding a solution to your challenge? We have been maintaining a list of solutions in [`instructors.md`](https://github.com/swcarpentry/python-novice-inflammation/blob/gh-pages/instructors.md); it would be great to keep that up-to-date if possible.

gvwilson:
Agreed that learners need to understand slicing, but I think we should try to find an example that's more authentic (i.e., something that might actually come up in real science).

tobyhodges:
Ok, I'll try to come up with something more relevant...

tobyhodges:
I find it difficult to judge the line between "something that might actually come up in real science" and obscure, so forgive me if these are still too generic/contrived. Appreciate the feedback!

gvwilson:
thanks - I'm happy if @tbekolay and @valentina-s are happy (though I'll reorganize at merge to put variable, slice, print together for each example to make it easier to follow)

valentina-s:
I will reorganize and merge. Also, I will convert the tuples to lists, we currently do not have introduction to tuples in the lesson. It will show an example of slicing a list of lists which I think will really test their understanding.

PR #185, opened 2016-01-04
Fix formatting for python division example

djw8605:
Add missing `>` which lead to improper formatting.

Before:
image

After:
image


tbekolay:
Thanks for your contribution!

PR #187, opened 2016-01-05
Change glob listing from HTML to CSV

djw8605:
Since the students are working in the data directory, not a html directory, glob the CSV data files.


djw8605:
Fixed in f5aaaf8d063a9d79c48a840ce4826f55d3c69bb6, beat me by a few days!

PR #188, opened 2016-01-05
Add exercise for the instructor training

metajinomics:
Adding exercise for instructor training. It checks that the learner understand command line arguments as well as handle exception that covered previous lessons.

### DIFF
--- a/10-cmdline.md
+++ b/10-cmdline.md
@@ -610,3 +610,10 @@ the program now does everything we set out to do.
 >
 > *   If no filenames are given, it reports the number of lines in standard input.
 > *   If one or more filenames are given, it reports the number of lines in each, followed by the total number of lines.
+
+> ## Generate an Error message {.challenge}
+>
+> Write a program called `check_argument.py` that print usage then exit program if there are no arguments are submitted. (Hint) You can use `sys.exit()` for exit program.
+> ~~~ {.output}
+> usage: python check_argument.py filename.txt
+> ~~~


10-cmdline.md
```diff
@@ -610,3 +610,10 @@ the program now does everything we set out to do.
 >
 > *   If no filenames are given, it reports the number of lines in standard input.
 > *   If one or more filenames are given, it reports the number of lines in each, followed by the total number of lines.
+
+> ## Generate an Error message {.challenge}
+>
+> Write a program called `check_argument.py` that print usage then exit program if there are no arguments are submitted. (Hint) You can use `sys.exit()` for exit program.
```
tbekolay: A few grammatical issues here:

- `print usage` -> `prints usage`
- `exit program` -> `exits`
- `if there are no arguments are submitted` -> `if no arguments are provided`
- `for exit program` -> `to exit the program`

tbekolay: Also, to echo what my comment from #184:

Could you add some line breaks here? We currently don't do this very consistently, but in most of our `.md` files, we have newlines either after a certain number of characters, or when it makes sense for the sentence (see [this blog post](http://rhodesmill.org/brandon/2012/one-sentence-per-line/) for details). The main reason for this is so that if someone makes a change to this paragraph later on, it will be more obvious what has changed because diffs are done on a line-by-line basis.

tbekolay:
Hi @metajinomics, thanks for the exercise! The idea looks great to me, and it's at the right level for the lesson.

I have a few suggestions that I think would improve the exercise (in addition to the inline comments I made). Let me know what you think!

- Usually we like to specify both the input and output that's expected, mostly for clarity. You've specified the output here, but it would be worth stating that you should get this output when you do `$ python check_argument.py`. Look at the "Finding particular files" challenge for an example.
- The exercise tells us what to do when we're unsuccessful, but what should happen on success? It would be good to give the learner a simple thing to do if command line arguments are specified (e.g., print the arguments, or print `"Thanks for specifying arguments!"`).

10-cmdline.md
metajinomics: ...
## Generate an Error message {.challenge}

Write a program called `check_argument.py` that prints usage then exit program if no arguments are provied. (Hint) You can use `sys.exit()` for exits.
~~~ {.output}
 usage: python check_argument.py filename.txt
 ~~~
...

metajinomics: @tbekolay Thank you for your suggestions. It is very helpful.

metajinomics:
@tbekolay Thank you for your kind suggestions. It is a very nice point. Here is the fixed solution. 


usage: python check_argument.py filename.txt

import sys
if (len(sys.argv) != 2):
    print "usage: python check_argument.py filename.txt"
    sys.exit()

print "Thanks for specifying arguments!"

gvwilson:
I'm happy to do some more editing on this once it's merged, but so far, I'm having no luck merging things that depend on patches... @tbekoly if you can merge as-is to create a fresh PR, then ping me, I'll do the rest.

tbekolay:
Merged, and edited the text as discussed above. Thanks @metajinomics!

Issue #136, opened 2015-07-06
Lessons never say to start the notebook


jiffyclub:
The [first lesson](http://swcarpentry.github.io/python-novice-inflammation/01-numpy.html) starts right off showing Python commands and there's never discussion of where someone following along should put those commands. This may be intentional so that the lessons are flexible, but it also leaves someone following along on their own confused as to where they should be typing. I'd recommend adding some preliminary discussion pointing people at the IPython notebook and terminal.

PR #189, opened 2016-01-06
Adding instructions to starting ipython.

djw8605:
This fixes issue #136.

### DIFF
--- a/index.md
+++ b/index.md
@@ -64,6 +64,17 @@ To do all that, we'll have to learn a little bit about programming.
 > $ cd && cd Desktop/python-novice-inflammation/data
 > ~~~
 
+> ## Starting iPython {.getready}
+>
+> We will be using iPython for the lessons.  iPython is an interactive interface for python that is very useful for learning and 
+> debugging python.  In the Setup, you should have already installed Anaconda which includes IPython.
+>
+> To start iPython, open a terminal or git bash and type the command:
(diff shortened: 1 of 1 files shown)


tbekolay:
While it is true that many instructors teach the lesson with the IPython / Jupyter notebook, the lesson materials are designed to be used from any environment (e.g., a bare Python interpreter). So, while we would definitely like to add in some text about how to start the notebook, it would have to be accompanied by text for other ways of starting up Python (specifically, starting a Python interpreter from the shell). If you don't mind adding that in that would be great; otherwise, I'll add in that text in another PR.

Additionally, people installing Anaconda from scratch at the moment are instructed to run `jupyter notebook` instead of `ipython notebook`. Both should be included as we can't guarantee what versions learners will have.

djw8605:
Both `jupyter notebook` and `ipython notebook` work on a new installation of Anaconda.  I wonder if this will be kept for a while?  Anyways, it works right now, and starts the same service.  A consistent command across versions like `ipython notebook` is probably the ideal way to go, for now.

As for using the python interpreter, I could add instructions for that ass well.  Really, it would just be running the command `python` from the shell.

djw8605:
Added Instructions for starting the interpreter in 579f7e3

tbekolay:
Looks good to me, thanks!

Issue #181, opened 2015-12-11
Totally need to add this image


gvwilson:
https://twitter.com/hadleywickham/status/643381054758363136

alexsavio:
In the R programming lessons, maybe? :)

tbekolay:
Done in #192, thanks @alexsavio! Your move, https://github.com/swcarpentry/r-novice-inflammation

PR #192, opened 2016-01-07
Add visual example of list indexing.

alexsavio:
Suggested in #181.

Not sure if I should add a reference to the original author.


tbekolay:
Hi @alexsavio, thanks for the PR! I like the image; feels like a good visual analogy that will help many learners.

A few nitpicky / technical comments:

- Since the HTML files are generated and not edited by hand, their diffs are not very helpful for understanding what a PR is doing. Because of that, we ask that you not commit the HTML files changed by the PR; as you can see in the [commit log](https://github.com/swcarpentry/python-novice-inflammation/commits/gh-pages), we regenerate the HTML after we merge a PR. Do you mind removing the HTML changes?
- Your first commit (2d4bb2d) fixes a problem that should be fixed in master now. If you can remove it from the PR that would be great -- or, I'm happy to omit this commit when merging in the PR, whatever you prefer!
- I think the callout is missing some details at the moment... What is the list `x`? It's important to note that `x` is a nested list of lists -- something that I'm not sure we've done in the lesson to this point. Perhaps the callout should also introduce the idea of nested lists, and then provide the picture as a way to reinforce the concept introduced?
- I think that we should indeed add a reference to the original tweet. It can be as simple as linking to the tweet when one clicks on the image, or it can be more explicit (e.g., "Adapted from @hadleywickham's [tweet](https://twitter.com/hadleywickham/status/643381054758363136) about R lists.")

alexsavio:
Hi @tbekolay, thanks for your comments.

- I can commit only the .md. I didn't know how you were managing this.
- I will redo the commit only with the image and the .md file
- Would you like me to add the concept of nested lists in the lesson?
- I can add that link to the image.

I am not an expert with git, could you please tell me how would be the cleanest way to redo this PR?

tbekolay:
> I can commit only the .md. I didn't know how you were managing this.

Thanks! No worries, the documentation is kind of scattered around different locations.

> Would you like me to add the concept of nested lists in the lesson?

I think it's sufficient to add the concept to the callout. Something like "Since lists can contain any Python variable, it can even contain other lists." then give an example where you might want to do it, perhaps?

> I can add that link to the image.

Sounds good, thanks!

> I am not an expert with git, could you please tell me how would be the cleanest way to redo this PR?

I think the easiest way would be to make a new branch and manually apply the changes you want to keep from this branch to that branch. Then you won't have to worry about modifying your original branch.

If you don't mind trying a few new git things, then you can reset the history of your branch, but keep the changes you've made to the files with:

```
git reset origin/gh-pages
```

You can then make changes as you see fit, then `git add` and `git commit` again. When you're done, because your history has changed, you will have to force push. You can do that with:

```
git push --force origin 1stlssn
```

If this seems confusing, then I would encourage you to make a new PR. Thanks!

alexsavio:
Hi, I am going to close this PR. The new one is #209.
Thanks for the instructions, I was already doing another branch when I received them.

PR #193, opened 2016-01-08
Added 'Ranking Olympians' exercise for lesson on loops.

ikajic:
Added the exercise "Ranking Olympians" in the lesson "2. Repeating Actions with Loops" as a requirement for wrapping up the instructor training.

### DIFF
--- a/02-loop.md
+++ b/02-loop.md
@@ -226,3 +226,37 @@ so we should always use it when we can.
 > Write a loop that takes a string,
 > and produces a new string with the characters in reverse order,
 > so `'Newton'` becomes `'notweN'`.
+
+> ## Ranking Olympians {.challenge}
+> 
+> Given a list of five men alpine skiing finalists at Sochi 2014 Winter Olympics:
+>
+> ~~~ {.python}
+> finalists = ['Viletta SUI', 'Kostelic CRO', 'Innerhofer ITA', 'Jansrud NOR', 'Zampa SVK']
+> ~~~
+> 
+> write a piece of code that prints ranked finalists:
+> 
+> ~~~ {.python}
+> 1 Viletta SUI
+> 2 Kostelic CRO
+> 3 Innerhofer ITA
+> 4 Jansrud NOR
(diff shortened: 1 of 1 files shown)


ketch:
This is a very straightforward exercise, but I like that it uses data that have some meaning.  It seems like a very nice way to introduce the `enumerate` function, which is very useful.  I don't know if it's okay for challenges to add more material to the lesson, though.

ikajic:
There are two things being introduced: enumerate, and together with it the
idea that we can manipulate two loop variables at the same time. I have
seen that some exercises introduce new concepts, to that's why I decided to
include enumerate. I can also remove it from the exercise and leave the joy
of discovery to the learner (or just hint it after the exercise), but then
I am not sure about how exciting the exercise is.

02-loop.md
```diff
@@ -226,3 +226,37 @@ so we should always use it when we can.
...
+> 1 Viletta SUI
+> 2 Kostelic CRO
+> 3 Innerhofer ITA
+> 4 Jansrud NOR
+> 5 Zampa SVK
+> ~~~
+>
+> When working with collections such as lists and strings, it is common to use
+> the index of an element and the element inside the loop at the same time. To
+> make it easier for us, Python provides a handy `enumerate` object which
```
QuLogic: `enumerate` is a function, not an object.

ikajic: The rationale behind saying `enumerate` is an object instead of a function is that I didn't want to give a learner the impression that they can use `enumerate` outside the for-loop as they use it within the loop. They might try to "call" enumerate by passing an iterable and expecting to get a value like they did in a for loop.

QuLogic: Well, you can call `enumerate` outside a loop, but since it returns a generator (which isn't explained elsewhere), perhaps that's not a good idea...

gvwilson:
I like the idea of introducing `enumerate`, but I think we need to find an exercise which has some scientific or research use: numbering strings from a list is unlikely to be compelling.

PR #194, opened 2016-01-09
Add a comparison logic error to the "Errors and Exceptions" topic

brunogrande:
Logic errors aren't formally introduced in the lesson, despite being the most dangerous since they can go undetected. I've added a simple exercise demonstrating a comparison logic error. 

(Pull request done as part of the [Instructor Training Checkout Procedure](http://software-carpentry.org/blog/2015/12/instructor-training-checkout-procedure.html).)

### DIFF
--- a/06-func.html
+++ b/06-func.html
@@ -30,7 +30,7 @@
           <h2 class="subtitle">Creating Functions</h2>
           <section class="objectives panel panel-warning">
 <div class="panel-heading">
-<h2><span class="glyphicon glyphicon-certificate"></span>Learning Objectives</h2>
+<h2 id="learning-objectives"><span class="glyphicon glyphicon-certificate"></span>Learning Objectives</h2>
 </div>
 <div class="panel-body">
 <ul>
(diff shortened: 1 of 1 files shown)


tbekolay:
Hi @brunogrande, thanks for the contribution! I like the idea behind the challenge question, but I'm not sure if it's the right place. We definitely agree that logic errors are the most dangerous type; however, the way we've structured topic 7 is to really focus on exceptions (though we do call them errors throughout -- partly because that's what Python calls them). I.e., we're not talking about all the things that can go wrong, just the things that Python will complain about and give you a traceback that might be confusing to new programmers.

Your challenge, however, fits in perfectly with [topic 8](http://swcarpentry.github.io/python-novice-inflammation/08-defensive.html), which is all about potential errors that Python doesn't tell you about by default. That topic also happens to be light on exercises, so would you mind moving your exercise to that lesson? It might be worth reading that topic over and making sure the text here matches the tone of that topic (you could even add the 'logic errors' intro to the topic itself, leaving the challenge itself as is).

Finally, would you mind also adding a solution to your challenge? We have been maintaining a list of solutions in [`instructors.md`](https://github.com/swcarpentry/python-novice-inflammation/blob/gh-pages/instructors.md); it would be great to keep that up-to-date if possible.

brunogrande:
A new pull request will be made using a separate branch.

PR #196, opened 2016-01-11
added two more exercises and some more detail explaining and testing …

tobyhodges:
…understanding of sequence slicing in steps, and omission of start/end index for the slice. The Bowie reference seems appropriate today - I hope you don't think it's too cute.

### DIFF
--- a/03-lists.md
+++ b/03-lists.md
@@ -167,15 +167,126 @@ odds: [1, 3, 5, 7]
 
 This is different from how variables worked in lesson 1, and more similar to how a spreadsheet works.
 
-## Turn a string into a list {.challenge}
+> ## Turn a string into a list {.challenge}
 
-Use a for-loop to convert the string "hello" into a list of letters:
+> Use a for-loop to convert the string "hello" into a list of letters:
+
+> ~~~ {.python}
+> ["h", "e", "l", "l", "o"]
+> ~~~
+> Hint: You can create an empty list like this:
+
+> ~~~ {.python}
+> my_list = []
+> ~~~
+
+Subsets of lists and strings can be accessed by specifying ranges of values in brackets, similar to how we accessed ranges of positions in a Numpy matrix. This is commonly referred to as "slicing" the list/string.
 
 ~~~ {.python}
-["h", "e", "l", "l", "o"]
+date = "Monday 4 January 2016"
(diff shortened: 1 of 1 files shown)


gvwilson:
I like the Bowie reference, but can you please split the PR into separate chunks (one per exercise) to simplify review?

tobyhodges:
Hi Greg,

No problem.

Cheers,

T

tobyhodges:
Hi Greg,

I’ve submitted three separate PRs, though I don’t believe that the material from each can really be included independent of the some material covering slicing in general. 

This is the first time that I’ve submitted multiple PRs at once. I must confess that I wasn’t sure of the workflow for this: I decided to submit the PRs from separate branches from gh-pages, containing only the changes for a single section exercise each time. By the end I had become convinced that this was the wrong approach. Instead, I now think that I should have created successive branches from each previous one, building on top of the changes added each time, and submitting PRs for each of those branches. 

I’d really appreciate your advice on this - what is the best/most robust approach when submitting multiple PRs for the same file? Sorry if I’ve cause you a headache when reviewing these - I’m still learning, and if you’d like me to try again with a different approach then please just let me know.

Cheers,

T

gvwilson:
HI Toby,

You did the right thing submitting each change from a separate branch.  
If you submit a PR from a branch XYZ, then update *your* copy of the XYZ 
branch and push those changes to your repo, the pull request is 
automatically updated: this is how people respond to code reviews and 
other feedback.  So sending three different PRs from a single branch, 
and having them all in play at once, isn't possible.

What you can do instead if you're brave is:

- create a branch ABC from gh-pages, make the first change, push ABC to 
your repo, submit the PR
- create a second branch DEF *from ABC* (rather than gh-pages), make the 
second change, push DEF to your repo, submit the PR
- create a third branch GHI *from DEF*, etc.

This preserves the chaining order, and also allows the PRs to be 
separate.  @wking may have an even better workflow - Trevor?

Thanks,
Greg

gvwilson:
This is good stuff, but I don't think we have enough time in this lesson to cover list slicing.  We *will* try to include it in https://github.com/swcarpentry/python-novice-gapminder, and if you'd like to redirect this text there, we'd be very grateful. Thanks!

tobyhodges:
No problem. I was planning to work on material for the Gapminder branch this evening, so I will take care of this too.

PR #197, opened 2016-01-11
Initial Jupyter Notebook conversion for lesson

davidmam:
I've converted the lessons from markdown to Jupyter notebooks and they run fine at [Mybinder.org](http://mybinder.org/repo/davidmam/python-novice-inflammation )
Apart from an almost direct translation, I found one bug in the 04-files section. `glob.glob` may not return a sorted list of file names so I added a bit to ensure the correct files were selected for the subsequent processing. I've not put that back into the Markdown file.


jduckles:
@davidmam, could you add these to a stand-alone repository in your account and we can link to it as supplementary material? We've been down the road of supporting ipynbs in our lessons repositories and it caused more problems than it solved. We're not opposed to revisiting it in the future, but for right now I think keeping it standalone is best.

tbekolay:
As @jduckles said, we'd be happy to see this material in a separate repository, but for the time being we don't require Jupyter for this lesson. Additionally, we're moving to a new lesson template that uses only Jekyll, so we can't incorporate a manual `nbconvert` step in our workflow. I'm going to close this PR, but if you do start up a repository with this material, please let us know on the [discuss mailing list](http://lists.software-carpentry.org/listinfo/discuss)!

PR #199, opened 2016-01-12
Omitting slice indices

tobyhodges:
Includes sections covering slicing of sequences, with examples and an exercise to test a grasp on simple slicing. The second section then covers slicing from the beginning or to the end of a sequence by omitting the first or second range index respectively.


tbekolay:
I think this will need to be cleaned up to flow better with the rest of the topic before the lesson release, but still a useful addition. Thanks @tobyhodges!

PR #200, opened 2016-01-12
Slicing MCQ

tobyhodges:
Includes a new multiple choice question to test the learner's understanding of how to slice sequences in Python. Inclusion of the MCQ assumes that slicing and step size have been covered already.

Admittedly, the MCQ isn't perfect e.g. it tests the learner on more than one single concept at once. If you would prefer that I create a few MCQs to cover each concept separately (providing index ranges, zero-based indexing, slicing in steps/step start position, blank range indices, ...) I will be happy to do so. In this case, a single MCQ was included as I had already written other exercises to cover the individual concepts separately.

### DIFF
--- a/03-lists.md
+++ b/03-lists.md
@@ -87,24 +87,24 @@ TypeError: 'str' object does not support item assignment
 does not.
 
 > ## Ch-Ch-Ch-Changes {.callout}
-> 
+>
 > Data which can be modified in place is called [mutable](reference.html#mutable),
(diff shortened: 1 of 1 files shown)


ketch:
This looks like a good MCQ, but I think it doesn't fit in the lesson currently since things like `x[::2]` haven't been taught yet.

valentina-s:
Thank you for the suggested exercise. We have already merged several slicing exercises, and we do not want to add additional new concepts as the lesson is quite long.

PR #201, opened 2016-01-15
Add MCQ for modifying data in lists

derekhoward:


iglpdc:
I think the challenge is nice, but a bit confusing, because it contains two questions. You ask about what the operator `*` does with list and also about in-place modification. I think I'd be better if you address these two questions into two separate parts of your challenge. (I'm not the maintainer of the lesson, just giving my opinion.)

About the first part, we don't talk much about how copies are done in Python (this is more advanced material, I think), so I'd avoid  "copies of the references". As they have seen in the lesson that `+` concatenates the strings, you could ask what they think the operator `*` should do. (This is a good lesson on keeping standard meanings of operators consistent when overloading them).

Your example could exploit the similarity with the same expression in `numpy` (multiplying an array by a number makes an element-wise multiplication) to stress operator overloading. My first guess is maybe that `repeats = counts * 2` makes `repeats` equal to `[4, 8, 12, 16, 20]`. So if you add that to your choices A and B, you kinda cover that  operators in Python can be overloaded to do very different things. 

The second part of the challenge could go on modifying `counts` and asking what happens to `repeats`.

derekhoward:
Thanks for the feedback. I simplified the exercise to focus on the first part of operator overloading. I think you're right that it may be best to avoid talking about creating copies of references with the * operator and the second question might be too subtle of a problem to include.

PR #202, opened 2016-01-16
added new question for lesson6: factorial

maikia:
Question added as a part of instructor training requirements.

### DIFF
--- a/06-func.md
+++ b/06-func.md
@@ -695,11 +695,11 @@ the second parameter in the list.
 > What should be A and B so that the function below returns the factorial?
 >
 > ~~~ {.python}
-> def factorial(n):
->     if n == 1 or n == 0:
->         return A
->     elif n > 1:
->         return n * factorial(B)
+> def factorial(A):
+>     result = 1
+>     for i in range(2, n+1):
+>         result *= i
+>     return B
 > ~~~
 >
-> Test the function for different values of `n`.
+> Test the function for different arguments.
--- a/06-func.md
+++ b/06-func.md
@@ -687,3 +687,19 @@ the second parameter in the list.
 >
 > print(k)
 > ~~~
+
+> ## Factorial {.challenge}
+>
+> Factorial of n (where n is non-negative integer) is the product of all positive integers equals or less to n. For example, factorial of n = 4 is equal to 4 * 3 * 2 * 1 = 24. With an exception that factorial of 0 equals 1
+>
+> What should be A and B so that the function below returns the factorial?
+>
+> ~~~ {.python}
+> def factorial(n):
+>     if n == 1 or n == 0:
+>         return A
+>     elif n > 1:
+>         return n * factorial(B)
+> ~~~
+>
+> Test the function for different values of `n`.


gvwilson:
Recursion is famously hard to teach in computer science, so introducing it as part of a challenge question is pretty brave.  Can you think of a way to re-do this using iteration instead?

gvwilson:
p.s. could you please mail Greg to tell him who this GitHub user ID belongs to?  We need to map it back to a real human being. Thanks!

tbekolay:
I am a bit concerned right now that this exercise is too similar to other exercises that we have in this lesson. Essentially, the goal is to write a function, but we do that in the first three exercises in this lesson. Originally, there was recursion, which would have made this exercise unique, but as @gvwilson pointed out, recursion is a tough thing to teach, and doesn't have many use cases that can't instead be done with iteration.

So, we need something in this exercise to make it different from other questions. I like how the question is set up with a bunch already done, but with some blanks to fill in. @maikia, what do you think about restructuring the question so that the unique part is that someone has given you a function and it's your job to finish it? Then, we can have either the name of the function, or the docstring be a third thing that has to be filled in.

Does that make sense to you @maikia? I think if we restructure the question in that way, it would be good to make it the first exercise in this lesson, as filling in the blanks of this function is a good way to lead into writing your own functions (which happens in the subsequent exercises).

maikia:
Thanks Trevor.

You are right about the place. It certainly is too easy of a question to place it behind all the others. However, the idea of the question now (after the change from recursion) was to check the understanding of the arguments passed to the function and values returned by it. 

I could place it in the end of 'Composing Functions' section and made it as MCQ. However, in the python lessons (in contrast to some other lessons) there are no questions in the middle and there are no MCQs so I am not sure if it really fits in to have one out of a blue.

What do you think?

tbekolay:
That makes sense to me! While we don't have any examples of MCQs in the middle of topics yet, I support adding one. Adding one will also hopefully encourage people to write more!

@abostroem, do you have any reservations with adding a MCQ in the middle of the functions topic? I think having all the challenges at the end was part of the `tools/check.py` script at one point, but that must have changed at some point since other lessons have challenges in the middle.

abostroem:
I don't think I was part of the conversation to put exercises at the end. My justification was always that if you want to put an exercise in the middle then maybe you should be breaking the lesson into smaller pieces. Although I'm not entirely opposed, I'd like to hear from someone about why we made that decision and how it has evolved in the other lessons (@gvwilson @jiffyclub?).

Other options to think about:
I could see making this a callout (maybe I'm misusing those) and calling a quick check.

Alternately, it isn't entirely clear to me what we plan to do with all of the MCQ. This could be a good one for a novice audience - even if it is at the end.

gvwilson:
On 2016-01-20 5:55 PM, Azalee Bostroem wrote:
>
> I don't think I was part of the conversation to put exercises at the 
> end. My justification was always that if you want to put an exercise 
> in the middle then maybe you should be breaking the lesson into 
> smaller pieces. Although I'm not entirely opposed, I'd like to hear 
> from someone about why we made that decision and how it has evolved in 
> the other lessons (@gvwilson <https://github.com/gvwilson> @jiffyclub 
> <https://github.com/jiffyclub>?).
>
Our first style guide insisted that exercises go at the end, and said 
that if you want to put an exercise in the middle, it means you should 
split the topic in two.  Many people have ignored this; we'll revisit 
this spring.

gvwilson:
Thanks again for the exercise - I think it duplicates others that we've now merged, but we would welcome other contributions.

PR #203, opened 2016-01-17
Fix rendering of last loops exercise solution

andreww:
We need to add a blank line before the solution
to the 'reversing a string' solution in the instructors
page so that it properly displays as code in the html.
While we are at it, mark the Python code {.python}, not
{.output}.

### DIFF
--- a/instructors.md
+++ b/instructors.md
@@ -141,11 +141,12 @@ After discussing these challenges could be a good time to introduce the `b*=2` s
 ## [Storing Multiple Values in Lists](03-lists.html)
 
 Solutions to exercises:
+
 >## Turn a string into a list {.challenge}
 >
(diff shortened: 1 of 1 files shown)


andreww:
We also need to do similar things elsewhere in the file. Second commit does this too.

tbekolay:
Apologies for the delay; we recently pushed some fixes for a bunch of formatting issues in the `instructors.md` file. I'm pretty sure the [rendered version](http://swcarpentry.github.io/python-novice-inflammation/instructors.html) is good now, but I would appreciate you taking a look and confirming that the issue is fixed @andreww :)

andreww:
I thought I had looked for other patches fixing these. Anyway, the rendered version looks OK now.

PR #204, opened 2016-01-17
add exercise for the 'functions' lesson

vdhelm:


tbekolay:
Thanks for your contribution @vdhelm! I like the idea behind this question -- it's quite a gotcha that you can't mix positional and keyword arguments freely. Even though I know you can't do it, every now and then I do it anyway and get met with a SyntaxError. I also like that there are some interesting design questions that can come up for discussion with more advanced groups (there's sometimes a clear order to the arguments, but the default values you have don't always line up with that order), but the exercise is still interesting for a classroom that won't get to that point.

I'm not sure about the presentation of this question, though. Specifically, the only way for a student to get this exercise correct is to know an admittedly esoteric Python fact a priori, or to try it out for themselves. To me, the argument list fact that we're teaching is not intuitive, so asking people to intuit it is more likely to make them worse rather than better; I would think that the most likely occurrence is for a student to guess 1., then be disappointed when they're wrong. Failure can be a good tool for learning, but in my opinion it should always be avoidable. On other hand, the ability to try things out is one thing I stress as being empowering when teaching, as once you get over that hump and get comfortable trying things, learning happens much faster. So I would err on the side of having learners just try it and then work together to figure out what's wrong, rather than "tricking" them into being wrong, then telling them why they were wrong. (Note: I know that you're not intentionally trying to trick them, I'm attempting to express how some hypothetical students might feel when attempting this exercise.)

I do think that it's useful to teach students this fact though, so what do you think about restructuring the question so that they try to run the code snippet, then pose deeper questions like "Why do you think that Python does not allow default values anywhere in the argument list? Can you rewrite the function so that it will run?"

vdhelm:
Hi @tbekolay,

Thanks for your feedback, I totally see your point, I was indeed expecting most students to 'get it wrong' the first time. I always notice how much you can learn from mistakes, which is what I was going for, getting an 'emotional' reaction so that you actually remember what happened.

PR #206, opened 2016-01-19
Make output consistent with input in lesson 5.

ketch:
Fixes a tiny inconsistency between input and output messages.


tbekolay:
Good catch, thanks @ketch!

PR #207, opened 2016-01-19
Add an exercise (counting vowels) to lesson 5.

ketch:
This exercise combines concepts from the lessons on loops and conditionals.

This submission is required as part of my instructor training.

### DIFF
--- a/05-cond.md
+++ b/05-cond.md
@@ -239,3 +239,7 @@ and false statements. However, they aren't the only values in Python that are tr
 > using in-place operators.
 > Do you think the result is more or less readable than writing the same without in-place operators?
 
+> ## Counting vowels {.challenge}
+>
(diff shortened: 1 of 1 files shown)


tbekolay:
Hi @ketch, thanks for the exercise! Looks nice and straightforward, and I like that it combines `if` statements and `for` loops.

Two minor notes:

- There may be some unnecessary confusion here about the letter `'y'`, as I grew up with the understanding that it's "sometimes" a vowel. It's definitely not a trivial problem to determine when it's used in what way, so I'd recommend being explicit that `'y'` is never considered a vowel for the purpose of this exercise.
- Would you mind also adding a solution to your challenge? We have been maintaining a list of solutions in [`instructors.md`](https://github.com/swcarpentry/python-novice-inflammation/blob/gh-pages/instructors.md); it would be great to keep that up-to-date if possible.

PR #208, opened 2016-01-19
Update 01-numpy.md

antolikjan:
MCQ added, as an entry for the SC discussion session.

### DIFF
--- a/01-numpy.md
+++ b/01-numpy.md
@@ -380,6 +380,23 @@ tripledata:
  [ 6.  6.  3.  3.]]
 ~~~
 
+> ## MCQ: Understanding numpy array operations {.challenge}
+>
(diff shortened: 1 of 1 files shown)


ketch:
This is a very nice MCQ, but in my opinion the title is not appropriate.  The question deals with Python lists, not numpy arrays.  In fact, the most likely reason to give a wrong answer is to confuse the two.  It's tricky (perhaps confusing) to refer to the lists as arrays also in the variable names.

antolikjan:
You are completely right, I have no idea why I said numpy arrays, it was silly.

gvwilson:
Thanks for the PR - we definitely need more on list slicing, but we also need exercises that show how it's used on authentic problems.  Can you please think about whether something like this could be put into https://github.com/swcarpentry/python-novice-gapminder instead?

PR #209, opened 2016-01-20
Add example of list of lists indexing.

alexsavio:
A corrected PR from #192.


ketch:
> we could represent a matrix as a list of lists

While one *could* do this, it's probably not a good idea to promote it to students, since a 2D numpy array will probably serve them better.

alexsavio:
Hi @ketch, I totally agree.

tbekolay:
I like that `x[0][0]` is `'pepper'` ;) This looks great to me, so I'm going to merge it. I worry a bit that instructors will feel that they have to teach nested lists, so I made issue #210 to brainstorm ideas to denote that not everything needs to be taught in every workshop.

PR #211, opened 2016-01-21
Added challenge to 01-numpy.md regarding slicing a vector vs an array

metavee:
Illustrates one of the differences between NumPy and MATLAB/intuition. Depending on the way you slice something, you may get back a 1D (N,) or 2D (Nx1) array, which can cause weird errors if you aren't expecting it.

### DIFF
--- a/01-numpy.md
+++ b/01-numpy.md
@@ -704,21 +704,43 @@ the graphs will actually be squeezed together more closely.)
 >
 > Modify the program to display the three plots on top of one another instead of side by side.
 
-> ## Converting between an array and a vector {.challenge}
+> ## Stacking arrays {.challenge}
 >
-> In some cases in NumPy, an vector of size N behaves differently than an Nx1 array.
-> The vector will have a shape (N,), while the array will have a shape (N,1).
->
-> The code below generates a 5x5 array of random numbers `A`,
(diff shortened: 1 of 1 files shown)


01-numpy.md
```diff
@@ -703,3 +703,22 @@ the graphs will actually be squeezed together more closely.)
 > ## Moving plots around {.challenge}
 >
 > Modify the program to display the three plots on top of one another instead of side by side.
+
+> ## Converting between an array and a vector {.challenge}
+>
+> In some cases in NumPy, an vector of size N behaves differently than an Nx1 array.
```
QuLogic: I don't think this challenge really explains _how_ it might behave differently, or why that might even be a problem?

metavee: Fair point. In my experience, it generally results in strange-looking broadcasting errors.

Perhaps I could reformat this as a small code fragment that results in such an error, and the challenge would be to fix the error. Do you think that would work better?

metavee:
I updated the challenge to try and guide the participants to the error in an example.

I'm not totally satisfied with how it turned out, but I found it quite challenging to come up with a clear and simple example. It would be great to get feedback on whether or not this is salvageable.

valentina-s:
That looks good! Thank you for the contribution!

PR #212, opened 2016-01-21
Add exercise on extracting elements from a list (+ small formatting fix)

pvanheus:
* Add exercise on extracting elements from a list
* Fix formatting bug from "Turn a string into a list" exercise

### DIFF
--- a/03-lists.md
+++ b/03-lists.md
@@ -223,9 +223,30 @@ This is different from how variables worked in lesson 1, and more similar to how
 > ["h", "e", "l", "l", "o"]
 > ~~~
 > Hint: You can create an empty list like this:
+>
 > ~~~ {.python}
 > my_list = []
 > ~~~
+
+> ## Extracting elements from lists {.challenge}
+> Earlier we saw that you can extract elements from NumPy arrays, for example, using the two-dimension
+> array `data` that we saw before:
+>
+> ~~~ {.python}
+> print(data[0, 0])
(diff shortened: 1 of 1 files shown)


03-lists.md
```diff
@@ -223,10 +223,31 @@ This is different from how variables worked in lesson 1, and more similar to how
...
+> ~~~
+>
+> This also works with lists. How would you extract the first number from this list of numbers? How would
+> you extract all the odd numbers?
+>
+> ~~~ {.python}
+> numbers = [1, 2, 3, 4, 5, 6, 7, 8, 9]
+> ~~~
+>
+> Could you use a loop to extract all the odd numbers? Try it!
```
tbekolay: Is it possible to do this without using an if statement? We don't present those until topic 5.

tbekolay:
I like this exercise, and how it links back to what we went over in the first topic!

One thing that isn't clear to me is what you mean by "all the odd numbers". Do you mean all the odd numbers in the list? Or do you mean all the odd-numbered indices? The former requires if statements, which haven't been introduced at this point in the lesson. The latter is quite straightforward, though it would need to be made explicit to prevent confusion.

gvwilson:
Thanks for the PR - we definitely need more on list slicing, and we're going to add that material to https://github.com/swcarpentry/python-novice-gapminder - could you please redirect the exercise there?

PR #215, opened 2016-01-22
Added an exercise to give practice in composing functions.

apdavison:
This PR is part of the requirements for instructor training.


gvwilson:
Thanks for the PR - I think this one covers the same ground as others we have now merged, but additional PRs against https://github.com/swcarpentry/python-novice-gapminder would be very welcome.

PR #216, opened 2016-01-23
Add challenge about floating-point accuracy.

QuLogic:
This is a wide topic, but I hope the description in the instructor's guide is sufficient to convey the general issue.

I'm a bit conflicted about the location. On the one hand, some of the lesson in `06-func` mentions this issue, but on the other hand it's not really relevant to functions. I think maybe it could be fit into `08-defensive` as something you need to be aware of when writing tests.

This PR is for Software Carpentry's Instructor training.

### DIFF
--- a/06-func.md
+++ b/06-func.md
@@ -685,3 +685,18 @@ the second parameter in the list.
 >
 > print(k)
 > ~~~
+
+> ## Floating-point accuracy {.challenge}
+>
+> What does the following code print? Is it what you expect?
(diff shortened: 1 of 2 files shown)


gvwilson:
We definitely need material on floating point precision, but I agree that the section on functions isn't the right place. Can you please look at https://github.com/swcarpentry/python-novice-gapminder and see if it belongs in there somewhere (perhaps with the material on testing)?

PR #217, opened 2016-01-27
Add blank line to fix code block formatting

adamobeng:
The generated HTML doesn't display 226--229 as a code block. It looks like pandoc requires a blank line before code blocks, but Github's builtin markdown renderer doesn't... Standards!


tbekolay:
Thank you!

PR #219, opened 2016-01-27
Remove reference to size of indent

adamobeng:
I think referring to the size of indents at this point could cause confusion. Given that it's not the first time we refer to indents, a reader might think that there is something about functions in particular that means that means they should be indented by four spaces. Also, this is already better dealt with in lesson 7.

### DIFF
--- a/06-func.md
+++ b/06-func.md
@@ -39,8 +39,7 @@ The function definition opens with the word `def`,
 which is followed by the name of the function
 and a parenthesized list of parameter names.
 The [body](reference.html#function-body) of the function --- the
-statements that are executed when it runs --- is indented below the definition line,
-typically by four spaces.
+statements that are executed when it runs --- is indented below the definition line.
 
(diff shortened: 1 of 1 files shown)


06-func.md
```diff
@@ -39,8 +39,7 @@ The function definition opens with the word `def`,
 which is followed by the name of the function
 and a parenthesized list of parameter names.
 The [body](reference.html#function-body) of the function --- the
-statements that are executed when it runs --- is indented below the definition line,
-typically by four spaces.
+statements that are executed when it runs --- is indented below the definition line.
```
iglpdc: I'd change "when it runs" -> "when the function is called".

Also why not adding "similarly to loops and conditionals" after "below the definition line".

tbekolay:
I agree; it is in part a personal preference, and if people are wondering what the standard is, they can look at the many code examples in the lesson. Merging!

PR #222, opened 2016-01-27
Small changes from review

QuLogic:
After reviewing the lesson for instructor training, I've put together several small changes that fix minor issues like typos, etc..

### DIFF
--- a/01-numpy.md
+++ b/01-numpy.md
@@ -207,7 +207,7 @@ print(type(data))
 <class 'numpy.ndarray'>
 ~~~
 
-The output tells us that `data` currently refers to an N-dimensional array created by the NumPy library. These data correspond to arthritis patients' inflammation. The rows are the individual patients and the columns are there daily inflammation measurements.
+The output tells us that `data` currently refers to an N-dimensional array created by the NumPy library. These data correspond to arthritis patients' inflammation. The rows are the individual patients and the columns are their daily inflammation measurements.
 We can see what its [shape](reference.html#shape) is like this:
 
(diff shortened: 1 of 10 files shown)


QuLogic:
It looks like I have fixed #217 and #218 in here too.

iglpdc:
Great! +1 (maybe merging first #217 and #218)

tbekolay:
This all looks great! I am going to skip the changes to `python3` in 1366e046e8d0048c760e744e4bbba3bdf3803a17 because 1) this breaks things for me because I use a virtual environment and 2) eventually `python` should reference `python3` and the special case will be `python2`.

QuLogic:
> 1) this breaks things for me because I use a virtual environment

Hmm, sounds like another thing broken with `virtualenv`. Conda and pyvenv seem to do this correctly?

> 2) eventually python should reference python3 and the special case will be python2.

Unfortunately, PEP0394 doesn't have much of a timeline for it, _but_ it does suggest writing a cross-compatible script if using the `python` reference, so everything should be working right now.

tbekolay:
> Hmm, sounds like another thing broken with virtualenv. Conda and pyvenv seem to do this correctly?

Yeah, this is mostly an issue with my setup. I'm using a python2 virtualenv; were I using a python3 virtualenv it would work fine. Another issue that I should have mentioned is that the `Makefile` is shared across all lesson repositories, so changing it might make it more difficult to pull in changes from [lesson-template](https://github.com/swcarpentry/lesson-template/blob/gh-pages/Makefile) in the future.

PR #223, opened 2016-01-27
Clarify expression of error types

adamobeng:
Without this change, it could be interpreted to mean that using the "read" flag instead of the "write" flag can cause a `FileNotFoundError`, whereas it's a different type of error entirely.


iglpdc:
Looks good to me.

tbekolay:
Me too, thanks!

PR #224, opened 2016-01-27
Merge two distinct references to argparse

adamobeng:


tbekolay:
Thanks, this is a lot better IMO as it keeps all of the argparse references in the callout, rather than the main lesson. Merging!

PR #227, opened 2016-02-06
Update numpy.md, files.md

joannadiong:
Modification to numpy.md
I instructed a python workshop and only introduced the magic command `% matplotlib inline` _after_ getting participants to run the section to show the heat map. I learned that on some computers (Macs, and a few Windows), the ipython notebook hangs if the magic command is not specified - a handful of participants were forced to shut down the notebook and restart. It would be helpful to include the magic command earlier when plotting the graphs to prevent notebooks from potentially hanging.

Modification to files.md
My machine runs on Linux Mint 17.2 (I am new to Linux). The `print(glob.glob('inflammation*.csv'))` command executed but produced a list of inflammation filenames that were out of numeric order. I checked with participants but it seems this behaviour only occurred on my machine. Everyone in that workshop was either running Windows or Mac so the problem may be OS dependent. It was easily fixed with by adding the `sorted()` option. This change could be considered for inclusion but it is not a major problem.

Thanks

### DIFF
--- a/01-numpy.md
+++ b/01-numpy.md
@@ -511,6 +511,7 @@ and use two of its functions to create and display a heat map of our data:
 
 ~~~ {.python}
 import matplotlib.pyplot
+% matplotlib inline
 image  = matplotlib.pyplot.imshow(data)
 matplotlib.pyplot.show(image)
 ~~~
--- a/04-files.md
+++ b/04-files.md
@@ -24,7 +24,7 @@ while `?` matches any one character.
 We can use this to get the names of all the HTML files in the current directory:
 
 ~~~ {.python}
-print(glob.glob('data/inflammation*.csv'))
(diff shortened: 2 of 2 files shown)


tbekolay:
For the `%matplotilb inline` change, could you instead make a separate callout that gives some context on it and how it is used? Currently this lesson does not require the use of Jupyter notebooks, so we would like to keep Jupyter notebook-specific instructions in callouts so that people not using notebooks can quickly see that instructions are notebook specific and skip them.

#246 also contains the `sorted` fix; there have been several comments there about how best to explain why `sorted` is useful and when is best to mention it, so could you look there and update this PR accordingly? Alternatively, if you prefer, you can take the `sorted` part of this PR out and leave it to #246.

PR #230, opened 2016-02-10
Add debugging challenge (part 09)

davidjb:
This exercise has the aim of ensuring learners are able to step through
unseen code with unexpected output to locate issues.  The issues present
are that the loop is not being utilised correctly (height/weight always
set as the first patient's data) and that the height/weight variables
are reversed.

### DIFF
--- a/09-debugging.md
+++ b/09-debugging.md
@@ -7,6 +7,7 @@ minutes: 30
 > ## Learning Objectives {.objectives}
 >
 > *   Debug code containing an error systematically.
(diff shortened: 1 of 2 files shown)


valentina-s:
Thank you for your contribution! Could you also, please, add the above description to the instructors manual (instructors.md). This way instructors will know what are the issues with the code. Thanks!

davidjb:
@valentina-s done, added notes and more to the instructors doc.

PR #231, opened 2016-02-10
Lesson 10:-cmdline Explanation for silent failure is wrong for readings-04.py. 

sschmeier:
Changed explanation.

The original explanation "If `action` isn’t one of the three recognized flags, the program loads each file but does nothing with it (because none of the branches in the conditional match). Silent failures like this are always hard to debug." is incorrect.

$ python readings-04.py --blah test.txt
will throw exception as `values` is not defined through one of the flags

However,
$ python readings-04.py test.txt
$ python readings-04.py --max
will run and this produce a silent failure.

We only get a silent failure if we do not specify at least two additional arguments on the command-line, one for the **flag** and one for the **filename**, but only one, the program will not throw an exception but will run. It assumes that the file list is empty, as `sys.argv[1]` will be considered the `action`, even if it is a filename.

### DIFF
--- a/10-cmdline.md
+++ b/10-cmdline.md
@@ -403,11 +403,15 @@ but there are several things wrong with it:
 
 1.  `main` is too large to read comfortably.
 
-2.  If `action` isn't one of the three recognized flags,
-    the program loads each file but does nothing with it
-    (because none of the branches in the conditional match).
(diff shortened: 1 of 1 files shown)


tbekolay:
Right you are. Thanks!

PR #232, opened 2016-02-11
Added loops diagram

tcford:
As part of SWC teacher training, I (Talitha Ford) created a diagram to help illustrate how loops work. Hope its useful!


tbekolay:
Hi @tcford, thanks for the PR! A few minor things:

- I think you forgot to add the image to the commit. Could you put it in the `img` folder and add it to this PR? Note that we usually prefer SVG diagrams over PNG, so if you happen to have an SVG you can use that instead :)
- We prefer Markdown syntax over HTML tags. You can include an image in Markdown with `alt text`. Instead of setting the width in the HTML tag, you can change the image itself to be the desired width.

tcford:
Sorry @tbekolay, 
I have include both .png and .svg images in the fig directory. Im not familiar with svg and can not work out why some of the text in the image is missing. 
I have changed the syntax to Markdown, though the image is quite large and might need resizing.
Did the new pull request work?

PR #233, opened 2016-02-11
New Question for 05-cond (Instructor Training Checkout)

amantero:
New Question for 05-Cond for Instructor Training Checkout, I thought introducing the idea of the while loop plus having them construct it using concepts from the just covered sections could be useful.


gvwilson:
Thanks very much for the PR - I agree that programmers should know about `while` loops, but the lesson is already full to bursting, so we're trying not to introduce new concepts.  I hope there will be room for this in the full-day novice lesson https://github.com/swcarpentry/python-novice-gapminder.

PR #234, opened 2016-02-12
Added exercise in conditionals section on using the break statement.

mcmero:
This is for Software Carpentry instructor training. The exercise is aimed at integrating concepts from the loops and conditionals modules, and introducing how loops can be terminated early using conditionals with the break statement.


gvwilson:
Thank you for the PR - the `break` statement is very useful, but we'd prefer an example that showed it doing something people actually do in real life (most of us don't calculate Fibonacci sequences).  Can you please re-submit with a different motivating example?

PR #235, opened 2016-02-15
Contribution to Python 02 Loop lesson

j-a-t:
This is my contribution to the Python looping lesson as part of my training as an instructor.


gvwilson:
Thank you for the PR - it's hard to talk about nested structures without talking about aliasing, which is something we've delayed to later lessons (because it's really hard to wrap your head around).

PR #236, opened 2016-02-16
Allegravia patch 1

allegravia:
This is an update of 06-funct.md that contains a challenge (exercise) on composing functions.


gvwilson:
Thanks very much for the PR - it's a good example.

PR #237, opened 2016-02-17
Added link to NumPy Documentation

OKCody:
Having the link to the NumPy documentation might be useful for people exploring its capabilities.


valentina-s:
That is a good idea, thank you!

PR #239, opened 2016-02-22
Update 01-numpy.md

gedankenstuecke:
made it %whos, as required for ipython 3.1.0 (and above?)


valentina-s:
With automagic on (%automagic 1, which was default for me with anaconda 3.5.1), one does not need % in front of the magic commands. I am merging your PR, so that we make sure it works for everybody. Thank you for your contribution!

gedankenstuecke:
Thanks for merging! My older Anaconda version apparently didn't enable it by default 😃

PR #240, opened 2016-02-22
Update 01-numpy.html

gedankenstuecke:
made it %whos, as required for ipython 3.1.0 (and above?)


nsoranzo:
You should modify only the `.md` file, not the .`html` . The latter will be updated by the repository maintainer once the pull request is accepted.

gedankenstuecke:
Ah, thanks, I quickly changed both as I was using the material in a course I was giving, so I just manually updated the html for a quickfix, but that makes sense :smiley:

PR #241, opened 2016-02-22
New self-documenting code exercise

stuckyb:
I created a new exercise to introduce the concept of self-documenting code and give learners a chance to reflect on their coding style and practice reviewing code with a partner.

### DIFF
--- a/06-func.md
+++ b/06-func.md
@@ -685,7 +685,6 @@ the second parameter in the list.
 > print(k)
 > ~~~
 
-
 > ## Self-documenting code {.challenge}
 >
 > What do these two functions do?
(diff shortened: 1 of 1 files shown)


gvwilson:
This is a useful example, but I think the bulk of it should be *outside* the challenge - the before/after example and accompanying text are lesson, and only the last paragraph is challenge.  Can you please refactor and resubmit?

valentina-s:
I like this example a lot: I even used it today in a SWC workshop to motivate the discussion for documentation. I think inserting it before the Testing and Documentation section would be good. I only worry that the term 'self-documenting code' may create the perception that as long as the names are human readable we do not need documentation, so I suggest removing the sentence which mentions it.

tbekolay:
I merged and made the changes suggested by @gvwilson and @valentina-s. Thanks for your contribution @stuckyb!

PR #242, opened 2016-02-24
Include a blueprint for a function image

wd15:
Improve the function lesson by including an SVG image of a function
with annotation. Image helps to break up the text and more easily
explain the parts of a function than just reading the text.

 * An explanation about how to reproduce the image is here,
https://gist.github.com/wd15/2b4ffbe5ce0d0ddb8a5b.

 * This addition is part of the instructor training for Daniel Wheeler.


06-func.md
```diff
@@ -35,6 +35,11 @@ def fahr_to_kelvin(temp):
     return ((temp - 32) * (5/9)) + 273.15
 ~~~
 
+![The blueprint for a python function](fig/python-function.svg)
+
+<!--- see https://gist.github.com/wd15/2b4ffbe5ce0d0ddb8a5b to
+regenerate the above figure --->
+
 The function definition opens with the word `def`,
```
iglpdc: Maybe you could use "keyword" instead of "word".

06-func.md
```diff
@@ -35,6 +35,11 @@ def fahr_to_kelvin(temp):
...
 which is followed by the name of the function
```
iglpdc: I'd remove the "which is": "which is followed" -> "followed"

iglpdc:
Looks a good idea to me, but for some minor details. I usually explain this when teach and it's nice to have it in the lesson.

Apart from the comments above, you use "signature" for the "def" part. But the signature is defined taking into account the arguments (and the return type in some other languages at least) [1]. 

I think better alternatives could be:

- label "def" as "def keyword" or something like that, and/or
- label the whole first line as "function prototype" or something similar. I don't know whether there's a better name in Python for this, but "function prototype" is the name used in other languages.

[1] https://www.python.org/dev/peps/pep-0362/

wd15:
@iglpdc, thanks for the feedback -- will update the pull-request soon.

wd15:
@iglpdc, updated without adding "function prototype", but used "def keyword" and modified the associated sentence.

valentina-s:
Looks good! Thank you for your contribution.

PR #243, opened 2016-03-06
Update 01-numpy.md

jstapleton:
Clarifications


valentina-s:
Thanks for adding the clarifications!

PR #246, opened 2016-03-10
glob.glob returns paths in arbitrary order.

nsoranzo:


abostroem:
@nsoranzo Thanks for your contribution. The use of sorted overlaps with #227. Commenting inline on the text changes you made.

04-files.md
```diff
@@ -16,8 +16,8 @@ The only thing that's missing is a library with a rather unpleasant name:
 import glob
 ~~~
 
-The `glob` library contains a single function, also called `glob`,
-that finds files whose names match a pattern.
+The `glob` library contains a function, also called `glob`,
+that finds files and directories whose names match a pattern.
```
abostroem: To me this makes it sound like it will return directories in the list as well (like ls) rather than files and their paths. Can you come up with a clearer way to say this?

04-files.md
```diff
@@ -28,12 +28,12 @@ print(glob.glob('data/inflammation*.csv'))
...
 
 ~~~ {.output}
-['data/inflammation-01.csv', 'data/inflammation-02.csv', 'data/inflammation-03.csv', 'data/inflammation-04.csv', 'data/inflammation-05.csv', 'data/inflammation-06.csv', 'data/inflammation-07.csv', 'data/inflammation-08.csv', 'data/inflammation-09.csv', 'data/inflammation-10.csv', 'data/inflammation-11.csv', 'data/inflammation-12.csv']
+['data/inflammation-05.csv', 'data/inflammation-11.csv', 'data/inflammation-12.csv', 'data/inflammation-08.csv', 'data/inflammation-03.csv', 'data/inflammation-06.csv', 'data/inflammation-09.csv', 'data/inflammation-07.csv', 'data/inflammation-10.csv', 'data/inflammation-02.csv', 'data/inflammation-04.csv', 'data/inflammation-01.csv']
 ~~~
 
 As these examples show,
-`glob.glob`'s result is a list of strings,
-which means we can loop over it
+`glob.glob`'s result is a list of file and directory paths in arbitrary order.
```
abostroem: I would add a comment about the use of the sorted function.

04-files.md
```diff
@@ -16,8 +16,8 @@ The only thing that's missing is a library with a rather unpleasant name:
 import glob
 ~~~
 
-The `glob` library contains a single function, also called `glob`,
```
abostroem: Do you disagree with single or think it is redundant? I like the emphasis is places.

04-files.md
nsoranzo: `glob.glob()` **does** return both files and directories, not sure what you are asking me to clarify here.

04-files.md
nsoranzo: The `glob` library (or more precisely module) does not contain only the `glob()` function, but also the `iglob()` iterator and the `escape()` function, see https://docs.python.org/3/library/glob.html .

04-files.md
nsoranzo: Do you prefer it here or after the next example, in which the function is used?

tbekolay:
This change was also proposed in #153; from that thread:

@gvwilson:

> Please modify to show glob.glob first (with items not sorted), then introduce sorted in a callout box with a note saying "you should only bother to sort if you really care, and you usually only really care as the last step of processing".

04-files.md
nsoranzo: @abostroem @tbekolay I have added a short description of the `sorted` built-in function.

tbekolay:
Could you make the comment as suggested by @gvwilson above, in a callout box?

nsoranzo:
@tbekolay I like it as it is know, I see no need for a callout box.

tbekolay:
Hm, reading it over again it does look like a pretty minimal set of changes to introduce both `glob` and `sorted`, so I'm :+1: on this. Does anyone object to merging this? I'll give the rest of the day for people to respond and merge tomorrow morning if there are no objections.

tbekolay:
Thanks for your contribution @nsoranzo!

PR #249, opened 2016-03-16
Exercise to create a new function and ins and outs of string concatenation

ghumurtaza:
The exercise explain 4 major concepts through one example:

1. Simple Function creation and returning a value from a function
2. Immutability of strings and how concatenation is really costly because of this
3. join operation and why it is faster
4. Extended slicing feature of python

This exercise would be used as an example while teaching one of these concepts. It could also be used as exercise at the end of a module where trainer could work with students on the exercise.

### DIFF
--- a/06-func.md
+++ b/06-func.md
@@ -684,3 +684,29 @@ the second parameter in the list.
 >
 > print(k)
 > ~~~
+
+> ## Write a function to swap a given string.
+>
+> There are multiple ways to achieve this:
+
+def reverse(given):
+    s = ''
+    for i in range(0, len(given)):
+        s = s+given[len(given)-i-1]
+    return s
+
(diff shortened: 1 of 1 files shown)


gvwilson:
Thanks for the exercise - can you please:

1. Format it as an indented block using `>` in the same way that other exercises are formatted.
2. Give the title the `{.callout}` style (similar to the other exercises).
3. Write a fuller explanation of what the learner is being asked to do - it wasn't until I read the code that I understood what "swap a string" meant.

I like the multiple solutions, but I worry that novices will find them overwhelming - we don't yet have a good way to record "notes to the instructor" for exercises, so perhaps one way to address this would be present one of the other solutions and ask the learner whether it always produces the same answer, why or why not?

PR #251, opened 2016-03-24
fixed two inconsistencies in the stdin part of Command-Line Programs

timtroendle:


tbekolay:
Good catch! I think it is especially ironic that we made the exact error that we say is a common error to make ;) Though the script works either way, which masked the error. Anyhow, thanks for the contribution, merging!

timtroendle:
Hehe, that's true. Happy to be able to contribute!

PR #252, opened 2016-03-24
removed notion of dictionaries

timtroendle:
Dictionaries haven't been introduced at that point and aren't used
outside this lesson. So they have been replaced by a container that
has been introduced and is used heavily: strings.


valentina-s:
Thank you for the contribution!

PR #253, opened 2016-03-25
Added multiple choice question on slicing

hachterberg:
This adds a simple multiple-choice question on slicing, created to assess possible confusions students could have on the topic. This follow the idea of informative multiple-choice questions taught during the instructor training, for which this is my belated homework.

### DIFF
--- a/01-numpy.md
+++ b/01-numpy.md
@@ -325,6 +325,26 @@ small is:
  [ 2.  2.  1.  1.]]
 ~~~
 
+> ## Check your understanding: Slicing arrays or lists {.challenge}
+>
+> Can you slice an array to get the correct part?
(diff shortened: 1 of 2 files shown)


valentina-s:
Thank you for the suggested exercise. We already have one exercise on slicing with a simple string 'oxygen'. Adding a more motivating example with multiple choice will be a good addition. For example, something along the lines extracting specific string from a filename. Would you like to add such context to the exercise?

tbekolay:
I also believe [this exercise in the lists topic](http://swcarpentry.github.io/python-novice-inflammation/03-lists.html#non-continuous-slices) has significant overlap with this exercise, so I'm going to close this. Thanks for your contribution regardless @hachterberg!

PR #257, opened 2016-03-30
alternative lesson 8 using doctests under development

claireverity:
Alternative/new lesson on test driven development under development using doctests instead of assertions. Building tests into docstrings is a gentler intro to testing that combines what learners already know about docstrings from the module on error checking and using 'help'. Not yet functional since some of the doctest modules changed under migration from ipython notebook to jupyter and I haven't worked out how to access the new doctest functions in jupyter.

### DIFF
--- a/08_alt_def.ipynb
+++ b/08_alt_def.ipynb
@@ -1,0 +1,366 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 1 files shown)


gvwilson:
I like the idea, but I think it would fit better into https://github.com/swcarpentry/python-novice-inflammation (the new novice Python lesson).  We are also using .md instead of .ipynb for source.

PR #259, opened 2016-04-01
Minor changes to python novice pages

marylhaley:
index.html - changed the "ipython notebook" reference to "jupyter notebook" since jupyter is mentioned earlier in the page

01-numpy.html - fixed a minor spelling error


valentina-s:
Thanks for catching those errors! Keep in mind for the future when you submit a pull request to modify the .md files not the .html ones. We generate the .html ourselves to avoid issues with formatting.

PR #261, opened 2016-04-05
Update 03-lists.md

gfucile:
Proposed exercise for check-out session

### DIFF
--- a/03-lists.md
+++ b/03-lists.md
@@ -249,3 +249,68 @@ This is different from how variables worked in lesson 1, and more similar to how
 >
 > Do they always do the same thing?
 > Which do you find easier to read?
+
+> ## Populate a list from an external file {.challenge}
+> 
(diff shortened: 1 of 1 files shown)


valentina-s:
Thank you for suggesting this exercise. I think there are too many new concepts introduced in one exercise. I don't find the introduction of opening files crucial, unless they work with non-csv data, which this lesson does not discuss.  I think if it is added, it should be incorporated in an exercise/callout which does something with the data in a  way that is not possible with loadtxt (the simplest example that comes to my mind is when the data set is too big and dont want to load it in memory, but want to calculate some statistics). Although useful, given the short length of the lesson and difficulty explaining what IN really is, I suggest leaving it out.  A simple exercise on splitting/joining strings could be added to the string discussion. Does @tbekolay agree?

tbekolay:
I agree, @valentina-s. A simple exercise on splitting / joining strings would be useful, as it's something that definitely comes up a lot! I'm going to label this with `revise-and-resubmit` and close it for now; we would be happy to have the splitting / joining exercise @gfucile :) But for now we want to clear the PR queue before changing to the new lesson template.

PR #265, opened 2016-04-12
List comprehensions

wgharib:
This exercice introduces the list comprehension concept in python

### DIFF
--- a/03-lists.md
+++ b/03-lists.md
@@ -251,11 +251,11 @@ This is different from how variables worked in lesson 1, and more similar to how
 > Which do you find easier to read?
 > ## Nested loops in list comprehension {.callout}
 >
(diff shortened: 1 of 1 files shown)


valentina-s:
Thank you for suggesting this exercise. I think list comprehensions are hard to understand for beginners. Unless @tbekolay disagrees, I will leave them out of this lesson, which already has a lot of new material.

tbekolay:
I agree... as much as I love list comprehensions (and set and dict comprehensions), students need to have enough experience with the non-list comprehension version of this kind of code to realize whey they're so great.

We don't have much in the way of an intermediate Python lesson yet @wgharib ([this one](https://github.com/swcarpentry/python-intermediate) is still in the planning stages), but when there is one, I would encourage you to submit these exercises there :)

PR #266, opened 2016-04-26
Update 02-loop.md

marylhaley:
Made a minor change to the wording that describes "range", to indicate that the range iteration stops *before* the second argument, rather than *at* it.


tbekolay:
I agree, _at_ is more ambiguous than just before. Thanks for your contribution!

PR #269, opened 2016-06-13
Finding type of elements of a numpy array

Mahdisadjadi:
Add a few lines to find out the type of elements (e.g floats, ints) in numpy arrays after mentioning `shape` and attributes of an array.

### DIFF
--- a/01-numpy.md
+++ b/01-numpy.md
@@ -217,15 +217,24 @@ print(data.shape)
 (60, 40)
 ~~~
 
-This tells us that `data` has 60 rows and 40 columns. When we created the
-variable `data` to store our arthritis data, we didn't just create the array, we also
-created information about the array, called [members](reference.html#member) or
-attributes. This extra information describes `data` in
-the same way an adjective describes a noun.
+This tells us that `data` has 60 rows and 40 columns. We can also find out
+the type of elements in `data`:
+~~~ {.python}
+print(data.dtype)
+~~~
+~~~ {.output}
+dtype('float64')
+~~~
+so elements are [floating-point numbers](reference.html#floating-point number).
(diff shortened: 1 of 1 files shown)


tbekolay:
Currently, we don't talk about the distinction between integers and floating point numbers in the first NumPy topic, though we do talk about it in a later lesson. What do you think the benefit of introducing this early on is? The NumPy topic already as a lot of material in it, many have argued too much, so I would be much more keen on removing material from this topic than adding material to it.

Mahdisadjadi:
Yes that's right we talk about them in `Creating Functions` lesson but I think among all attributes `shape` and `dtype` all more important and useful. So I think the emphasis should not be on the distinction between integers and floating point numbers but on how to find the type of an array.

tbekolay:
But given that at this point in the lesson we have not introduced the difference between integers and floating point numbers, is the type of the array meaningful to students at this point? Perhaps discussing the `dtype` would be better done as a callout box in the Creating Functions lesson? Like I mentioned before, the first topic is already way too long, so I don't think that I am okay with adding anything without taking anything away.

Mahdisadjadi:
Actually Creating Functions lesson talks about `dtype` in passing parameters to `loadtxt`. Do you think adding a callout box right after that is useful? 
Alternatively, we can talk about loading a file with integers using `loadtxt` (non-default value for `dtype`).

tbekolay:
Hmm, perhaps the callout box makes more sense right after the "Integer division" callout? That's the only place where we're really explicitly talking about integers and floats.

tbekolay:
OK, I think I'm convinced that this is the right place since this is where we use the `type` function. I've merged this now, and rearranged it slightly to flow better with the rest of the lesson (also made this a callout).

Issue #271, opened 2016-06-13
overlap in first figure 05-cond


medelero:
overlap in first figure of the 05-cond Conditionals.

btel:
That's what it looks like in Firefox 46 on Linux:


![conditionals](https://cloud.githubusercontent.com/assets/41565/16021301/2e6e50a6-31b2-11e6-81e6-cd00a7a19561.png)


see also #270

gvwilson:
Is this still an issue?

valentina-s:
This has been resolved by replacing with .png's.

PR #273, opened 2016-06-13
link PNGs instead of SVGs (closes #270 and #271)

### LINKED ISSUE #270: overlap in figure: Operation Across Axes
There is an overlapping in the figure "Operation Across Axes" of the section 01-Numpy.

btel:
It replaces links to SVG files with links with respective PNGs. This solves the problem with differences of rendering of SVGs on different platforms and browsers (mainly related to fonts - see #270 and #271).

### DIFF
--- a/01-numpy.md
+++ b/01-numpy.md
@@ -455,7 +455,7 @@ next diagram on the left), or the average for each day (as in the
 diagram on the right)? As the diagram below shows, we want to perform the
 operation across an axis:
 
-![Operations Across Axes](fig/python-operations-across-axes.svg)
+![Operations Across Axes](fig/python-operations-across-axes.png)
 
 To support this,
 most array methods allow us to specify the axis we want to work on.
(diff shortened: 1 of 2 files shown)


tbekolay:
Could you remove the changes to the HTML files in this PR? Right now there's a ton of noise for a small change in the markdown files. Let me know if you want help with this!

gvwilson:
Hm - we generally prefer SVG to Markdown because SVG rescales better on 
small screens, but on the other hand, if the fonts aren't rendering...

btel:
@tbekolay sorry for that. I fixed the PR.

@gvwilson I agree it would be better to link SVGs instead of PNGs. The problem appears mostly in FF. On Chrome there are only minor formatting issues (the monospace font does not render).

tbekolay:
> sorry for that

No problem :) I'm 👍  on this PR but will wait for at least one other :+1: before merging (@gvwilson?)

gvwilson:
Go ahead - we'll figure it out.

btel:
BTW An alternative solution would be to replace text by paths. This way we keep SVG format, but we loose the possibility to edit texts.

tbekolay:
Thanks @btel! PS: I'm a big fan of svg_utils, used it for a lot of figures :)

PR #275, opened 2016-06-13
use 'import' a library instead of 'load'

btel:
"load" and "import" are used interchangeably in 01-numpy. This PR replace all instances of 'load' with 'import' for consistency. It also adds some explanation what it means to import a library (=open).


aflaxman:
This seems like a good idea, all changes make sense to me.

tbekolay:
LGTM too, thanks @btel!

PR #278, opened 2016-06-30
lists.md: heterogeneous data and nested lists

gabayae:


gabayae:
This example is meant to demonstrate that a list can contain heterogeneous data, and that the way we refer to an element in the list remains very basic of the corresponding type of structure.

tbekolay:
Hi @gabayae, thanks for your contribution!

I think that the idea of heterogeneous types of data in lists is a useful concept to introduce, but I don't think that it's imperative enough to have it in the main lesson. I think that a callout box would be appropriate, as we've done for lists of lists.

I also think that the example introduces more complexity than is necessary for the novice audience that we're targeting. We currently do not talk about dictionaries or classes or objects at all in the lesson, so those references should be removed. We also talk about lists of lists elsewhere, so I think for this lesson all that is necessary is to mix integers, floats, and strings, which are three types that we've introduced at this point. I think a simple example like `sample_age = [2, 5.5, "Unknown"]` gets the point across, and is something that a scientist could see being useful.

gabayae:
Hi **@tbekolay**, thanks for the comment. 

I think I agree with you regarding the complexity. The example you gave is surely more appropriate.   I actually noticed later on that " dictionaries" were not discussed at all in the lesson.

maxim-belkin:
Closing this PR but the discussion is useful. I'll create an issue with a request for a PR

PR #279, opened 2016-07-02
Given a list A, we have showed that B=A assigns A memory but not A va…

jmf-mas:
…lues


tbekolay:
Hi @jmf-mas, thanks for your contribution!

The HTML files in this repository are automatically generated from the source files, which are in markdown. In order to review your changes, could you make them to `03-lists.md` instead, and leave the HTML file unchanged?

jmf-mas:
Ah okay thanks.

PR #280, opened 2016-07-02
Copying list values

jmf-mas:
This part of code makes a difference between copying list values and sharing its memory. We show how to do in either copying list values or pointing to its memory location.

### DIFF
--- a/03-lists.md
+++ b/03-lists.md
@@ -86,6 +86,54 @@ TypeError: 'str' object does not support item assignment
 
 does not.
 
+> ## Do I share list memory or list values? {.callout}
+>
+> One thing we should know in python, while assigning a list A to another list B, the instruction B=A does assign A values to B, rather it shares A memory to B.
+> In other words, A and B have the same memory, which means any change in A concerns B and vice versa.Let see it through the following example:
+>
+>~~~ {.python}
+>A = [1, 4, 9, 25, 36]
+>B = A
+>print('A is :', A)
+>print('B is :', B)
+>~~~
+>
+>~~~ {.output}
+>A is: [1, 4, 9, 25, 36]
+>B is: [1, 4, 9, 25, 36]
+>~~~
+>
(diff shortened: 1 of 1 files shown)


03-lists.md
```diff
@@ -86,6 +86,54 @@ TypeError: 'str' object does not support item assignment
...
+>
+>~~~ {.output}
+>A is: [10, 4, 9, 8, 36]
+>B is: [10, 4, 9, 8, 36]
+>~~~
+>
+>All depends on the motivation of doing it. But if the motivation is to have a new list B with its own memory location, one of the ways is to do as follows: 
+>
+>~~~ {.python}
+>B=A[0:]
```
dpshelio: Wouldn't be the following the same? 
```python
B = A[:]
```
Is it easier to understand it using `[0:]`? I think neither of these notations is shown in the lesson before... maybe we should do:
```python
B = A[0:5]
```

jmf-mas: Hi #dpshelio

Yeah,

B = A[:] and B = A[0:] give the same result,  B = A[0:5] as well.

I did not see where it was used before in "Programming with Python"
lesson that I read.
Can you show me the place, please?

Anyway, my contribution is in a sense that, compared to other
programming languages, when we assign
a variable to another, this creates a new location memory. But it is
not the case with Python.

Best regards.

dpshelio: No, I didn't mean this was already thought, but the concept of indexing inside the list. Nowhere before this it has been shown that by not indicating the first or the last element means the first or the last respectively, and therefore I would keep that by teaching it as: `B = A[0:5]` instead of `A[0:]` or `A[:]`

maxim-belkin:
Closing this beause the callout box `Ch-Ch-Ch-Ch-Changes` talks about the topics discussed in this PR.

Thank you @jmf-mas  for your contribution and @dpshelio for the review!

PR #282, opened 2016-07-08
Corrected a typo on the lesson.

JUpani:
This is my ever first pull request, so if I did not do everything right or did not say the right things, it was not intentional. Thanks


gvwilson:
Thanks very much for the pull request - could you please make the change to the Markdown file rather than to the HTML? We generate the latter from the former.

JUpani:
You are welcome Greg
I think now it's done. Is that what you wanted?

tbekolay:
Thanks for your contribution @JUpani!

PR #283, opened 2016-07-08
moved making choices section to before loops section

jrosner:
imho teaching the idea of scope and program flow is easier to understand if you start with the most simple example, i.e. if condition
once this is understood, loops should be easier to comprehend.


gvwilson:
Thanks for the PR - it's what we need for instructor training.  Our experience with novices, though, is that it's hard to motivate conditionals unless you have multiple values to process: `if x > 0: do something` seems kind of pointless if you have just one value for `x`.  Processing multiple values, on the other hand, is instantly appealing, and putting a conditional inside the loop to process different values different is then a natural extension.

PR #44, opened 2015-02-16
Putting in glossary references.

gvwilson:
We no longer use **bold** for glossary terms, but instead mark them as `[word](reference.html#label)`.  This pull request puts all the references in that form, but does *not* add all the terms to the glossary.

PR #284, opened 2016-07-08
Merging new styles and documentation

gvwilson:
This PR brings in all of the 2016-06 template, along with documentation on building lessons.


valentina-s:
Thanks Greg for converting the lesson! One remark I have is regarding the times allocated to the lessons: as of now I think the schedule is overly ambitious. Here we usually cover material up to functions including command line programs in a full day. I think this varies widely on the students, instructor, number of exercises... Maybe we should explicitly state somewhere in the lesson that these times can be variable: I don't want instructors to start rushing for materials just because we fixed a lesson to 20 min. That I think applies to all lessons.

There are still a few typos and inconsistencies which were left after the merges, so I put a list of all of them in one issue #287 (including some of my ideas for improvement). I could fix the simple stuff after the merge and leave more major modifications for the future.

gvwilson:
Hi Valentina,

Pull requests to make the times more realistic are very welcome - what's 
in there now is mostly out-of-date guesswork.  And yes, there should be 
a note somewhere that times are provisional and advisory - what and 
where do you think best?

Cheers,
Greg

valentina-s:
We could add a footnote to the title Schedule, and below the actual schedule state something like

The suggested schedule will vary depending on topics and exercises covered.

?

gvwilson:
Sure - file a PR in the lesson-example repo and I'll fold it into the docs?

valentina-s:
Filed as #295 within this lesson. Also submitted a PR to the lesson-example #44.

gvwilson:
Merged.

PR #285, opened 2016-07-13
2 small changes

jangari:
Additional callout for arbitrary loop variable names, and importing numpy as is, and as np.


valentina-s:
Thank you for your suggestions! I agree that the loop callout is helping understand the loops better. I think having both np and numpy imported is a bit confusing. Instead I suggest to say something along the lines that if they work with other people's code it is good to agree on the convention.

valentina-s:
I modified the final commit and merged your PR. Thanks!

PR #290, opened 2016-07-20
Update 01-numpy.md

devendra1810:
I feel we need to show students the basic steps first as to create a variable etc. and then tell them about libraries and how to import them and whats the use of them.

### DIFF
--- a/_episodes/01-numpy.md
+++ b/_episodes/01-numpy.md
@@ -29,83 +29,7 @@ keypoints:
 - "Use the `pyplot` library from `matplotlib` for creating simple visualizations."
 ---
 
-Words are useful,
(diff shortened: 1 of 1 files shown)


valentina-s:
I actually had the same feeling when teaching both Python and R, but I can see how this approach aims to motivate the students from the beginning. Let's get input from others.

tbekolay:
Hm, initially I was skeptical but I do often opt for this ordering, so I'm :+1: on this. The only change I think is required before merging is to add links to words like "variable" to their definition in the glossary. We try to do this the first time these jargon are used; with the reordering, the first usage changes. Linking them both is perfectly fine too!

PR #294, opened 2016-08-04
A more intuitive solution to exercise in "loops".

rueuntal:


tbekolay:
Hi @rueuntal, can you make separate branches for each change? Right now since you're adding more commits to the `gh-pages` branch, those commits are also a part of this pull request, as the pull request is for a live branch, not for a static snapshot of the branch at the point you make the pull request. It's easier to review pull requests that make one change rather than several.

rueuntal:
Hi @tbekolay - sorry about that! How can I fix it?

tbekolay:
Hey, sorry for the delay! There are a few ways to go about it, depending on your familiarity with git... probably the easiest way forward is to make a separate feature branch for each of the changes. Some generic info can be found [here](https://bocoup.com/weblog/git-workflow-walkthrough-feature-branches), but specifically in this case I would do something like this:

```bash
git remote add swc https://github.com/swcarpentry/python-novice-inflammation.git
git fetch swc
git reset swc/gh-pages
```

This will reset your history back to where swcarpentry's gh-pages branch is. Your changes will still be in all of the files, but your commits (and your commit messages) will be gone (you might want to copy your commit message text for this reason).

Then, for each feature:

```bash
git checkout -b featurename
git add -p
git commit
git push origin featurename
git checkout gh-pages
```

`git add -p` will allow you to choose a subset of your changes to be included in the next commit. In the end, you should have separate branches for each change, which you can open as separate pull requests.

Let me know if I can help with any of those steps!

rueuntal:
Thanks for walking me through! I've created two new PRs for changes in the two lessons (#300 and #301). Please let me know if there are still problems!

PR #295, opened 2016-08-08
Adding a footnote to the schedule that times will vary.

valentina-s:


tbekolay:
Is this something we should add on this lesson specifically, or should this be done in the more general https://github.com/swcarpentry/styles ?

valentina-s:
This one was merged in the lesson template, so closing.

PR #296, opened 2016-08-08
Update switcheroo example

jrleeman:
The switcheroo example in the functions episode had all of the possible answers on one line and designated with a, b, c, d, unlike the other examples of 1, 2, 3, 4. This was also confusing as the variables in the example are a and b. I updated the formatting of the question and possible answers. I also added the solution with an explanation of why that is the correct answer. (SWC instructor training qualification)


valentina-s:
Thank you for your contribution!

PR #300, opened 2016-08-19
A more intuitive solution for exercise in "Loops".

rueuntal:


valentina-s:
Thanks for the contribution! This is indeed cleaner!

PR #304, opened 2016-08-25
Fixing references to .html pages

gvwilson:


tbekolay:
Checked this out locally and I no longer get 404s, so that's good! However, the anchor links in the reference page no longer work. This is because the references page no longer makes `id` attributes on the definition list elements.

I'm going to merge this now anyhow, since it's a clear improvement over the 404s. I made #305 to track the anchor link issue.

PR #307, opened 2016-08-27
new solutions to exercises in 03-lists.md

btel:


valentina-s:
Thanks a lot for all the solutions! Just one request: for the solution exercise 'Non-continuous Slices', could you provide the solution with all the indices and then add :: syntax (since at that stage they have not learnt yet about dropping indices)

btel:
Sure, I updated the PR. Thanks!

valentina-s:
Thanks!

PR #308, opened 2016-08-27
BUG: Fix broken link to setup

dopplershift:
This needs to be a relative link to setup, not absolute.


tbekolay:
Thanks for the quick fix!

PR #309, opened 2016-09-05
Notebooks with blanks

plstonge:
With the Python workshop, typing everything takes too much time. Participants are concentrated to copy what we type, and they do not listen to explanations. With these notebooks, I want them to focus on the new thing they are learning. For example, if I want to teach the for loop, all print statements are already there.

Exercises are no longer at the end of the chapter. I tried to apply the principle of "doing an exercise after learning up to 7 new things". This is why exercises are distributed in the appropriate sections of each chapter.

In the same "notebooks" folder, we can find notebooks with blanks (0x-yyy.ipynb) and their solution (soln-0x-yyy.ipynb).

Now, regarding _episodes/03-lists.md, I have improved the primes/odds examples. I have also fixed a print statement.

EDIT: I would like to use these notebooks for the event of September 8 in Montreal (Canada). Thanks!

### DIFF
--- a/_episodes/03-lists.md
+++ b/_episodes/03-lists.md
@@ -413,7 +413,7 @@ print("With known last position:", sond)
 sond = months[8:len(months)]
 print("Using len() to get last entry:", sond)
 sond = months[8:]
-("Omitting ending index:", sond)
+print("Omitting ending index:", sond)
 ~~~
 {: .python}
 
--- a/_episodes/03-lists.md
+++ b/_episodes/03-lists.md
@@ -220,15 +220,16 @@ If we make a list and (attempt to) copy it then modify in place, we can cause al
 ~~~
 odds = [1, 3, 5, 7]
 primes = odds
-primes += [2]
+primes[0] = 2
+odds += [9]
 print('primes:', primes)
 print('odds:', odds)
 ~~~
 {: .python}
 
 ~~~
-primes: [1, 3, 5, 7, 2]
-odds: [1, 3, 5, 7, 2]
+primes: [2, 3, 5, 7, 9]
+odds: [2, 3, 5, 7, 9]
 ~~~
 {: .output}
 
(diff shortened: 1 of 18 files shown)


valentina-s:
Thanks for preparing all these notebooks: they are very useful resource! We have discussed before whether to include notebooks in the lesson, but the biggest problem with this is keeping them in sync with the main lesson (as changes are continuously made), so in general we prefer them to be separate. It would be useful to have a way to link to them. @tbekolay is there some place to put supplementary materials? We can create a final 11-resources.md file which contains useful links?

tbekolay:
> is there some place to put supplementary materials

I think adding something to the Extras menu at the top would be good; I believe that any valid markdown file we place in the `_extras` folder will be added to that list, so we could add an "external resources" page or something like that. The notebooks would have to be placed somewhere easily (and ideally permanently) accessible; any thoughts on where these could go? We won't add them to this repo due to the difficulty in keeping them in sync with the lessons.

plstonge:
In my humble opinion...

I believe notebooks should come with all the necessary data samples. Proposed notebooks are in the "notebooks" folder and are consistently using the "../data/" prefix for input files. Participants are then learning how to separate their code from their data.

That being said, having to download and extract one archive file for the data and one extra archive file for notebooks, some participants may find a way to put the data in the wrong parent folder, which will lead to extra problems to debug in the classroom. IMO, we do not have time to lose on this kind of problem. We better keep data samples together with notebooks, either in python-novice-inflammation or in a new project (inflammation-material, for example).

Synchronizing the online material with notebooks should not be a big problem as all instructors should always review and test, before any event, what they are about to teach because all lessons are continuously getting updated. And if there is any important discrepancy between the online material and notebooks, any certified instructor that wants to use those notebooks should already have the knowledge to do corrections through GitHub.

Finally, the Python lesson has already too much material to fit in 3.5 hours, even if using notebooks. Instructors are already skipping subsections and exercises in the very updated online lesson. Having notebooks that are slightly outdated should not be a big problem if subsections and exercises are identified/labelled such that instructors and participants can follow.

Thanks!

PR #310, opened 2016-09-16
Update 01-numpy.md on dot notation.

rozmanov:
The use of the dot notation in accessing properties and methods of an object are not clearly stated in the lesson. I added a diagram and a brief discussion to address this issue.

This is the fist step of my graduation procedure.


tbekolay:
This PR contains the diagram, but I don't see the brief discussion, perhaps it wasn't `git add`ed?

Anyhow, I worry about the cognitive load involved in introducing the idea of "objects"... we try to gloss over this topic deliberately as it's quite deep and we have too much material as is. What do you suggest that we remove to make room for a discussion of objects?

rozmanov:
Yes, I did not add the updated text. Sorry about it. I updated the text and pushed it to my repository. Should it be enough for you to see the changes? The discussion is really brief, but in my opinion it is important to touch upon the issue. At least this should be better than some partisan introduction of an idea, I guess.

valentina-s:
I find this concept too complicated for the beginning of the lesson, and I don't think we should make the lesson any longer. Maybe putting the example in a callout or an optional exercise will be more appropriate? The exercise can have a simple question: name one property/method of a numpy array: eg shape, max().

Issue #287, opened 2016-07-15
Overall review of lesson 2016-06 by @valentina-s


valentina-s:
- [ ] (01-numpy) Typo in Data Type section: We can also find the out the type of the data contained in the NumPy array. 
- [ ] (01-numpy) Issue with sentence: 'More convenient when indices are computed rather than constant' - I don't understand the ending - seems unfinished. 
- [ ] (01-numpy) Multiple assignment - move it to later, maybe when discussing creating own functions 
 maxval, minval, stdval = numpy.max(data), numpy.min(data), numpy.std(data)
- [ ] (01-numpy) Change 'What if we need the maximum inflammation for all patients' to 'inflammation for each patient over all days'
- [ ] (01-numpy) Typo in sentence: 'The mathematician Richard Hamming once said ... (or course) ...'
- [ ] (01-numpy) Exercise – Drawing Straight Lines (Why are we expecting the lines to be straight?)
- [ ] (02-loops) Explicitly say that the collections is the word = collection of characters.
- [ ] (02-loops) Collection is not ordered - use sequence instead of collection.
- [ ] (02-loops) Range does not produce a list in Python 3, so students cannot see the output of range.
- [ ] (03-lists) State 'individual values and characters' in 'There is one important difference between lists and strings: we can change the values in a list, but we cannot change the characters in a string.'
- [ ] (03-lists) Why not use Darwin instead of Bell?
- [ ] (03-lists) Substitute equally with similarly in 'And equally, you can omit the ending index in the range to take a slice to the very end of the sequence'?
- [ ] (03-lists) Some formatting issue before Tuples and Exchanges exercise
- [ ] (03-lists) Tuples and Exchanges - since tuples have been removed, title should be changed, also I do not think the goal of the exercise is very clear, maybe relate to multiple assignment
- [ ] (03-lists) Typo 'the the' in: If we want to start by analyzing just the first three files in alphabetical order, we can use the sorted built-in function to generate a new sorted list from the the glob.glob output.
- [ ] (05-cond) Type (double Not Not) in 'That's Not Not What I Meant'
- [ ] (05-cond) Repetition of Tuples and exchanges: move or remove
- [ ] (06-func) Type in: 'min, mean, and and max of centered data are' 
- [ ] (07-errors) Decide on what to do with the error python code.
[...]

PR #311, opened 2016-09-18
Replace 'collection' with 'sequence' in for-loop explanation

MichaelConnell:
Fixes one of the items in Issue #287:
 [02-loops] Collection is not ordered - use sequence instead of collection.

Replaced three instances of the word 'collection' with the word 'sequence.'


valentina-s:
Thanks!

PR #312, opened 2016-09-22
Fix final defensive programming challenge so it initializes from data

DamienIrving:
At the end of the defensive programming lesson it says that an important rule of programming is to *always initialize from data*. In the solution to the final challenge of that lesson the highest and lowest value are initialized using `numpy.inf`. I've changed this so they are initialized from the data instead.


tbekolay:
Thanks for this! It's definitely good to practice what we preach :)

PR #315, opened 2016-09-27
Adding auto-generated IDs

gvwilson:
The new version of Kramdown will add IDs to glossary entries (definition lists) if instructed to do so.

### DIFF
--- a/reference.md
+++ b/reference.md
@@ -5,6 +5,7 @@ permalink: /reference/
 
 ## Glossary
 
+{:auto_ids}
 additive color model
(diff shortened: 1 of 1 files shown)


tbekolay:
Awesome! For anyone trying this out locally, to update Kramdown do `gem update kramdown`. With this, IDs are generated and reference anchor links now work when using the latest Kramdown, and will presumably start working on Github pages once they update their Kramdown version (current version is listed at https://github.com/github/pages-gem/blob/master/lib/github-pages/dependencies.rb#L15 for reference).

PR #320, opened 2016-10-06
fix quotation, indentation

stragu:
Challenge is not currently appearing on https://swcarpentry.github.io/python-novice-inflammation/05-cond/


valentina-s:
Thanks for the contribution!

Issue #267, opened 2016-04-28
Item Error (lesson 7) should read Index Error?


rdtarvin:
I think the subheading Item Error in the Errors and Exceptions lesson should be "Index Error", unless you were trying to be more general.

PR #322, opened 2016-10-10
Fix issue #267

wligtenberg:
I agree with the reporter.
Since the only errors that are mentioned are IndexErrors it makes sense to just call it Index Errors.
If at some point other errors regarding containers are added, we might want to call it Container Errors or something like that.


valentina-s:
I think Index Error is good as it is a standard type of error in Python. Thanks!

Issue #276, opened 2016-06-14
Introduce enumerate function


gvwilson:
#193 introduced `enumerate`, which is very useful and ought to be included in exercises, but which needs a better motivating example.

dstndstn:
it looks like this issue was addressed in PR #323 which has been merged -- so this issue should be closed.

FelixLiu-SF:
I think the current enumerate explanation in Computing the Value of a Polynomial can be improved by explaining all of the variables - not just i and x. 

Underneath the example,
for i, x in enumerate(xs):
    # Do something with i and x

It currently reads: "The loop above assigns the index to i and the value to x." 
It may be better to say something like: "The code above loops through xs, assigning the index to i and the value to x."

valentina-s:
Thanks for the suggestion. Would you like to submit a PR to add this modification?

PR #323, opened 2016-10-14
Added exercise for enumerate in loops lesson

paalmbj:
This is a maths-based example for `enumerate` in reference to issue #276. I found it a bit difficult to find a use for `enumerate` which wasn't better served with `zip` or slicing. An obvious one is to report errors on line / column numbers, but that usually requires `if`, which hasn't been introduced yet.

[This is my instructor training checkout contribution]

### DIFF
--- a/_episodes/02-loop.md
+++ b/_episodes/02-loop.md
@@ -299,3 +299,47 @@ so we should always use it when we can.
 > > {: .python}
 > {: .solution}
 {: .challenge}
+> ## Computing the Value of a Polynomial
+>
+> The built-in function `enumerate` takes a sequence (e.g. a list) and generates a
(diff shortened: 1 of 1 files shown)


gvwilson:
Thanks very much for the PR - can you please mail gvwilson@software-carpentry.org so that we can match this GitHub account to a person?

tbekolay:
I like this exercise a lot! The solution is nice and short which makes it doable in a short time period, as well as shows off the power of for loops.

One issue is that you need a new line separating your exercise from the other exercises around it. I'll fix this in the merge, but keep that in mind if you make more exercises in the future.

Thanks!

paalmbj:
I'll remember to add a new line next time :)  Hope the exercise is useful to someone, thanks for merging.

PR #324, opened 2016-10-17
Adds solution to challenge `Stacking Arrays` in episode one

itcarroll:
This is my first PR for a Software Carpentry lesson, so please don't hesitate to correct my procedure.

The way I interpreted the challenge, it is necessary for the trainee to avoid removal of singleton dimensions that can result from indexing. I don't think the solution gets ahead of future lessons, but perhaps an maintainer could say?

### DIFF
--- a/_episodes/01-numpy.md
+++ b/_episodes/01-numpy.md
@@ -1094,8 +1094,9 @@ the graphs will actually be squeezed together more closely.)
 > > A 'gotcha' with array indexing is that singleton dimensions
 > > are dropped by default. That means `A[:, 0]` is a one dimensional
 > > array, which won't stack as desired. To preserve singleton dimensions,
-> > the index itself can be an array. For example, `A[:, [0]]` returns
-> > a two dimensional array with one singleton dimension.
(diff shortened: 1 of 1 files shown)


valentina-s:
Thank you for the contribution! 
Your solution is ok (I suspect the array indexing would look a bit like a magic to a novice, but I do not see much better alternative). Another option is to explain that the stacking functions treat the one dimensional arrays as row vectors, so to obtain the desired result we need 

`numpy.vstack([A[:,0],A[:,-1]]).T`

Unfortunately, I realized transposing has not been introduced in this part yet, so it would require a brief explanation. So if you wish you can add it as an alternative.

itcarroll:
I guess array indexing had only been used with slices in the lesson, so I replaced the `A[:, [0]]` with the equivalent `A[:, :1]` and likewise for the last column. That shouldn't be introducing anything new into the challenge solution. (Transposing, as you note, would be a new topic.)

valentina-s:
I like this solution! Could you just replace the above appearance of A[:,[0]] with A[:,:1], and I will merge.

valentina-s:
Thanks!

PR #327, opened 2016-10-21
`FileNotFoundError` supersedes `IOError`

davis68:


valentina-s:
Thanks! Seems also when writing to a file opened for read only, Python 3 returns an UnsupportedOperation Error. Could you make this change too? Apparently, in Python 3 IOError has been merged with OSError, so even if a function explicitly raised an IOError it returns an OSError message. For example, pandas.read_csv returns OSError.

davis68:
Here's a note expanding on your comment and attempting to clarify the situation in a general way, @valentina-s.

valentina-s:
Thanks!

PR #332, opened 2016-11-28
Add solution to Mixing Default and Non-Default Parameters exercise

ldko:
This adds a solution for the Mixing Default and Non-Default Parameters exercise in the Creating Functions lesson.


valentina-s:
Thanks for the contribution!

PR #333, opened 2016-11-29
wording adjusted for consistency in description of loop operation

corinnebosley:
The characters in a word were described in two different ways here (in L140 and L141, the characters were described as variables in a collection, and in L149, the word was described as the variable).

I have changed the wording on lines 140 and 141 to be a little less confusing.  Open to suggestions for different wording, but I think it should be consistent.


tbekolay:
This makes sense to me, thanks for your PR!

PR #336, opened 2016-12-16
Show how to define a list before using strings/chars to illustrate looping

gboushey:
I noticed that the chapter on loops in python introduces the concept of iterating through a list through strings and characters.  I thought that the concepts might be easier to follow if the learners first explicitly defined a list.  This way, a learner isn't encountering two ideas at the same time - first, that variables can be lists as well as single value, second that strings aren't ordinary variables even if they seem that way at first.

I added this to the start of the lesson on python loops, then transition into the notion that strings can also be treated as lists, iterated over, accessed through index, and so forth.  This may lead to a little bit of repetition, similar loops over strings that a learner already did over a list of integers, but I think the slightly different angle might help reinforce the concepts.  

This all does makes the lesson a little longer, but doesn't introduce any new concepts or complexity.

### DIFF
--- a/_episodes/02-loop.md
+++ b/_episodes/02-loop.md
@@ -25,14 +25,14 @@ We have a dozen data sets right now, though, and more on the way.
 We want to create plots for all of our data sets with a single statement.
 To do that, we'll have to teach the computer how to repeat things.
 
-So far, we've seen how to create variables that hold a single value.  For example, we can create a variabe, "a", that holds the integer value 10.
+So far, we've seen how to create variables that hold a single value.  For example, we can create a variable, "a", that holds the integer value 10.
 
 ~~~
 a = 10
 ~~~
 {: .python}
 
-We've also shown how to use these variables to create new values.  For instance, we can create a new variable, "b", that holds the value of the first variable multiplied by ten.
+We've also shown how to use these variables to create new variables.  For instance, we can create a new variable, "b", that holds the value of the first variable multiplied by ten.
 
 ~~~
 b = a * 10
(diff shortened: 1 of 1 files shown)


gboushey:
Note - this is for instructor training.

gboushey:
I made an additional change to indicate that strings are not actually a list of chars, there are some differences.  

I'm not sure if this is the right time to talk about these differences, might be overload.  It could be an interesting example for the class to mention that while they can be treated similarly in some ways, other operations that work on lists will fail, and to see if anyone can come up with a bit of code that will work on a list but not on a string.

tbekolay:
I like the new text; it's well written and easy to follow! However, the content itself is largely a duplicate of episode 3, which is all about lists (see `03-lists.md`). As you mention, the two concepts are pretty closely connected (lists and for loops). The reason we used strings in this episode is so that we could talk about for loops without the cognitive load of lists. There's definitely a case to be made that lists should be discussed before loops, but I think in either case we want to talk about them separately. In my opinion, for loops are a more fundamental concept in programming so I like to get to it quickly -- it's a powerful construct! Whereas lists are not necessarily as mind blowing -- they kind of follow from the idea of looping over something. This is something we do a lot in this lesson, is give the big important idea first, then fill in the details later. [python-novice-gapminder](http://swcarpentry.github.io/python-novice-gapminder/) takes the opposite approach, where we start small and build up to these ideas (you'll notice there that lists are introduced before loops). I think that, because python-novice-gapminder now exists, we should keep the for loops first in this lesson.

What do you think about moving your text to `03-lists.md` to give some more context to the list introduction? Also, when you do this, please add new lines to your markdown source; if you look at the rest of the files, there are newlines inserted manually at reasonable places. The reason for this is so that when people make changes to the text in the future, the diff is easier to read, as only the lines that are changed get marked as changed (whereas if it's all in one line, then we have to hunt within that line to figure out what's changed).

gboushey:
I agree, this is very similar to the gapminder approach (all well as chapter 3 in this series).  There's lot of value in keeping the written lessons from getting too verbose.   I'll take a look at lesson 3, but I'm hesitant to add too much since I do like how compact it is.   Since these lesson plans are designed for workshops with instructors and teaching assistants, there's plenty of opportunity to bring up these as part of a related discussion, so I'm thinking what I wrote might be more useful as instructor notes?  Not sure where we'd put something like that right now...

Issue #329, opened 2016-11-13
Possible inaccuracy in 'Defensive Programming'


dndln:
https://swcarpentry.github.io/python-novice-inflammation/08-defensive/

Section: Testing Assertions
```
def running(values):
    assert len(values) > 0
    result = [values[0]]
    for v in values[1:]:
        assert result[-1] >= 0
        result.append(result[-1] + v)
        assert result[-1] >= result[0]
    return result
```

> The second assertion checks that the first value in the list is positive. Input such as [-1,0,2,3] will make it fail.

The second assertion is actually checking if the last value of the 'result' list is positive on every iteration. On the first iteration it does check that the first value is positive, since result[0] == result[-1], but on subsequent iterations it does not.

tbekolay:
Yes, you're quite right, a PR clarifying that would be great!

I don't recall when we added this exercise, but it strikes me as an odd one... why would we want to constrain the function to only accept monotonically increasing positive numbers? It's also not at all clear what would happen if we ran `running('abc')` with the 'defensive' code, as `'a' >= 0` is not obvious.

Perhaps the whole exercise should be replaced with something more straightforward?

ldko:
This has been resolved.

PR #337, opened 2016-12-16
Fix issue #329

dndln:
Referencing issue #329

### DIFF
--- a/_episodes/08-defensive.md
+++ b/_episodes/08-defensive.md
@@ -472,26 +472,26 @@ This violates another important rule of programming:
 
 > ## Testing Assertions
 >
-> Given a sequence of values, the function `running` returns
-> a list containing the running totals at each index.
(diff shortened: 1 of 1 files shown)


gvwilson:
Thanks very much - if this is for instructor training, can you please mail lessons@software-carpentry.org to let us know?  We don't have this GitHub ID in our records yet.
Thanks,
Greg

dndln:
This isn't for instructor training (I don't know what that is!). 

I just read through the tutorial to learn, and noticed this as I went.

valentina-s:
Thank you for working on fixing this exercise! It might be still confusing to people why the values should be positive and integer. I think if you rename it as something like get_total_count  and add a sentence in the beginning that these are counts, would make it more clear.

dndln:
Updated!

valentina-s:
Thanks!

PR #339, opened 2016-12-21
Python3 fixes

pletzer:
Output of the print commands are inconsistent with what either python2 or python3 would return. 
The output would be consistent if one were to use python2 and print without the parentheses. However, the Setup tab (http://swcarpentry.github.io/workshop-template/#setup) clearly requires learners to install python 3.x.  This change mixes the advantages of print with commas and print with format.

### DIFF
--- a/_episodes/01-numpy.md
+++ b/_episodes/01-numpy.md
@@ -133,7 +133,7 @@ print('weight in pounds:', 2.2 * weight_kg)
 {: .python}
 
 ~~~
-weight in pounds: 121.0
+('weight in pounds:', 121.00000000000001)
 ~~~
 {: .output}
 
--- a/_episodes/01-numpy.md
+++ b/_episodes/01-numpy.md
@@ -140,11 +140,23 @@ print('weight in pounds:', 2.2 * weight_kg)
 As the example above shows,
 we can print several things at once by separating them with commas.
 
+For a more readable output, one can use the format function:
+
+~~~
+print('weight in pounds: {}'.format(2.2 * weight_kg))
+~~~
+{: .python}
+
+~~~
+weight in pounds: 121.0
+~~~
+{: .output}
+
 We can also change a variable's value by assigning it a new one:
 
 ~~~
 weight_kg = 57.5
-print('weight in kilograms is now:', weight_kg)
+print('weight in kilograms is now: {}'.format(weight_kg))
 ~~~
 {: .python}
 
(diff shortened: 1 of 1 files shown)


tbekolay:
Thanks for your PR! It definitely needs to happen as the print statement output wasn't updated correctly for Python 3. However, I feel that the `format` method and the `{}` syntax used by `format` are too complicated for novices. I think the most straightforward way to do this is to use `+` for string concatenation; for example, the line

```python
print('weight in pounds:', 2.2 * weight_kg)
```

would become

```python
print('weight in pounds: ' + str(2.2 * weight_kg))
```

This would require the introduction of the `str` function, but that function is much easier to explain ("it converts any variable to a string") than the `format` method ("well, it allows you to build a string with another string that is kind of like a template for what the final string will be once you fill in certain special characters with the values of variables..."). It also means we don't have to introduce the `{}` syntax used by the `format` method.

I would also just use this form of printing at the start, rather than using a bad form of printing and then saying "oh, actually let's use this other thing." We're doing the reader a disservice by introducing the first form of printing only to then tell them they should do something else.

valentina-s:
`print('weight in pounds:', 2.2 * weight_kg)`
The current print statements work for me on Python 3.5.2. Am I missing something?

_episodes/01-numpy.md
```diff
@@ -133,18 +133,30 @@ print('weight in pounds:', 2.2 * weight_kg)
 {: .python}
 
 ~~~
-weight in pounds: 121.0
+('weight in pounds:', 121.00000000000001)
```
QuLogic: This is what happens in Python 2, not 3.

valentina-s:
I see the issue. It is a bit unfortunate (or not?) that the numerical error shows up in the beginning of the lesson. 
`print('weight in pounds: ' + str(2.2 * weight_kg))` still shows the error for me.

One way to fix it is to say that the computer introduces a small error and we cannot have precision for very small values. We can fix the output using round:
`print('weight in pounds',round(2.2*weight+0.1,13))`

I think round is pretty easy to understand.

An alternative which avoids this discussion in the beginning is to set weight to 65. Then 2.2*65 shows as 143. But the students will observe this behavior sooner or later so might as well explain it.

What do you guys think?

tbekolay:
Hm, I think the original issue was also related to the fact that in Python 3 `print` is a function, whereas in Python 2 since it is not a function, when you do `print('weight in pounds:', 2.2 * weight_kg)` this is interpreted as "please print this tuple", instead of "call the print function with these two arguments". Some of the proposed changes in this PR were using Python 2, rather than Python 3.

Regarding the rounding, though, if we did happen to use a number that doesn't print well, then I would be in favor of changing the weight such that it prints well. The original choice was, I imagine, arbitrary also, so no harm in changing it!

PR #340, opened 2017-01-03
[Instructor trainning submission] Added new example to debugging episode with overlapping errors

panchyni:
From the instructor training, I remember talking about editing R markdown files to contribute to lessons,
but I did not see any .rmd files in this repo and the CONTRIBUTING.md was not clear on this. As such, I made changed to the .md files. If this is an error, please let me know so that I can correct it. Thank you.

### DIFF
--- a/_episodes/09-debugging.md
+++ b/_episodes/09-debugging.md
@@ -282,3 +282,74 @@ not more.
 > >   `calculate_bmi(...)`
 > {: .solution}
 {: .challenge}
+
+> ## Interacting Begs for Interacting Genes
+>
+> One test to check if products of two genes interact
+> is if effect (phenotype) of deleting both genes(AB) is
+> different from the sum of the effect of deleting each
+> gene individuals (A+B)
+>
+> In the same way, bugs in your code may interact with one
+> another. The effects of one error may not be obvious until
+> another is fixed. This is why is making changes one at a
+> time with a clear goal in mind is so important.
+>
+> To debug the following code, first determine how many
+> interacting genes are in the phenotype. Then address the
+> bugs you encounters one at time. Which errors masked each
+> other and how?
+>
+> ~~~
+> #Phenotypes:   A    B    AB
+> phenotypes = [[0.5, 0.2, 0.9],\
+>              [0.3, 0.3, 0.1],\
+>              [0.3, 0.7, 1.0],\
+>              [0.3, 0.5, 0.8],\
+>              [0.5, 0.4, 0.5]]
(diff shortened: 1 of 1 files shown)


tbekolay:
Hi @panchyni thanks for your pull request! You're correct that this lesson doesn't have R markdown files, so we edit the .md files.

I like the idea behind this exercise, and it would be great to have more exercises in the debugging section! My main concern with how it is currently written is that it might be difficult for students who are not in genetics to understand what the correct solution should look like. For example, is it important to use the term "phenotype" when a more generic term would suffice? Why do you use A and B to refer to the two genes? What do the numbers in "phenotypes" mean, how should a student conceptualize these numbers? (I.e., what does it mean for Gene A to have phenotype value 0.5?)

Code-wise, it seems a little confusing to me that the `count` variable is passed into the function. Was this done to give something that's debuggable, or is this a common pattern that you've encountered? In these types of examples, it's a tough balancing act to write code that is buggy but still readable. It's all good if you've seen this often; in the code that I see more often, I would expect `count` to be created inside the function. Also, there are several style issue -- we try to adhere to [PEP8](https://www.python.org/dev/peps/pep-0008/) in code snippets, though we don't do any automated testing of this at the moment.

It would be great if you could address these concerns above as I think this could be a great exercise! Let me know if you want any help with rewordings or style.

panchyni:
Thanks for the feedback! 

Sorry if there was too much jargon in the genetics example. I was actually trying to use as much generic language as possible, but based on your comments, I think this made things more confusing. Instead of "phenotype" and "deletion", I could reword the example to talk about how often an organism dies because of a mutation. The same principle would apply: if two mutations cause the organism to die more/less often than the sum, that would indicate interaction. Perhaps this would be a little clearer? Ultimately, the language is flexible, but I do like how the theme of gene interaction lines up with the lesson.

As to the use of `count`, it is something I have seen before, but it was a stylistic choice on my part. I have the habit of being (perhaps overly) explicit with my examples, because I expect students to forget things, like scope, in the courses I normally teach. This is one of many things I am going to have unlearn when writing for a more focused course. I will fix this and conform the code style to PEP8 for my next submission.

I will have an update for you (hopefully) soon.

panchyni:
Hello @tbekolay,

I have rewritten my example based on your suggestions. I think the intro and code should be more readable now, but I am not as confident about the style. I went through the PEP8 specifications a couple times, but I'm still not that familiar with it, so I don't have a good sense of if I caught everything that needs to fixed.

PR #341, opened 2017-01-07
Devendra1810 gh pages #290

valentina-s:
This is the merge of PR #290. I moved the reference for variable  and assign to the beginning. I also added some motivational sentence in the beginning and some transitions in between the two parts. Since this was a bit more major reorganization I am submitting it as PR so that @tbekolay and @devendra1810 have a look at it and see whether it is consistent with what they had in mind. In the future maybe this initial example could be related to converting some inflammation values so that it sounds like a continuous story.


valentina-s:
Merging this PR after resolving a conflict.

Issue #288, opened 2016-07-18
Missing scripts?


weaverbel:
In the Python lesson sections, **Errors and Exceptions** and **Command Line programs**,  scripts, e.g. readings_04.py, are referenced but not provided in the initial data download. When preparing the class, @LeahRoberts had to track down the scripts on GitHub and then convert them to Python 3 so they would work. Has anyone else had this problem or are the Python 3 scripts available somewhere we don't know about? If so, can the lesson be updated so they are easier to find.

jangari:
I just ran this for myself for the first time the other day, and I think this section (command line programs) would actually work better if we had to write our programs from scratch. I was confused by the material jumping forward to a different readings_0?.py script. It seems to allow the students to not write the scripts but simply run them from the command line.

I would further suggest that one could port the code from iPython notebooks (or wherever they've done their inflammation work) into a command line script, and make the necessary modifications to output the plots as PNG files. From [here](http://stackoverflow.com/questions/9622163/save-plot-to-image-file-instead-of-displaying-it-using-matplotlib-so-it-can-be) it looks as simple as a .savefig method on the figure object.

maxim-belkin:
Resolved in #344

PR #344, opened 2017-01-12
Add zip file of 'code' directory and refer to it on the Setup page

mrawls:
- This addresses Issue #288
- Learners will now have both 'data' and 'code' directories
- These codes are used in the 10-cmdline lesson (Command-Line Programs)


valentina-s:
Thank you for these instructions!

PR #345, opened 2017-01-14
Fix callout formatting instruction in loops lesson.

sechilds:
For the "What's in a name?" callout block in lesson 2, the block was
not being rendered as a callout. This change fixes that formatting.


valentina-s:
Thanks for the fix.

PR #346, opened 2017-01-15
Use list instead of tuple in variable swapping exercise

tv3141:
> right, left = left, right
> "...we pack right and left into a tuple and then unpack it again into left
> and right."

This example for variable value swapping introduces three new concepts:
    1. the tuple data type
    2. special tuple syntax, omitting parentheses
    3. container unpacking in an assignment

I think using a list is much clearer, and only introduces the concept of
container unpacking.

right, left = [left, right]


valentina-s:
Thanks, I agree that using a list is more clear (I think these were remnants from the old lessons which introduced tuples).

Issue #60, opened 2015-04-08
Add link to data files?


BernhardKonrad:
I'm not sure of the intention of the html version, but wouldn't it make sense include a direct link to the csv data files? That way students can work on this tutorial at their own pace without having to clone the repository?
See also #9.

r-gaia-cs:
> I'm not sure of the intention of the html version

Because we use GitHub Pages to host the HTML version and GitHub Pages can't compile Markdown using Pandoc.

> wouldn't it make sense include a direct link to the csv data files?

Yes. I thought that I did fixed it once but couldn't find the commit. =(

> That way students can work on this tutorial at their own pace without having to clone the repository?

+1

iglpdc:
What I do usually to avoid the students having to clone the repo is to make a zip with the files needed in each lesson and put a link in the workshop website (see [here](http://iglpdc.github.io/2015-04-01-wcmc/), at the end of the "General info" section). Maybe we could provide a script that makes the zip and adds the links, so it's faster if the instructor needs/wants to do it.In my experience, specially for the shell lesson, it's tricky to have git working in all computers.

wking:
On Wed, Apr 08, 2015 at 08:19:36AM -0700, Ivan Gonzalez wrote:
> What I do usually to avoid the students having to clone the repo is
> to make a zip with the files needed in each lesson…

You can create a Git repository and have GitHub (or whatever hosting
site) build the zip automatically.  For example, see
efran/2015-04-09-UW#16, [1], and [2].  We're not taking this route for
UW, but I still think it's the best-of-both-worlds approach.

[1]: https://github.com/wking/2015-04-09-UW-data
[2]: https://github.com/wking/2015-04-09-UW-data/archive/master.zip

BernhardKonrad:
Thank you for these answers. You are right, for workshops the instructor can always provide alternative ways (git clone or zip download). Making this easier and more obvious for instructors may be worthwhile in its own right.

I was more thinking about the use of the lecture material outside or after a workshop as a learning resource. For starters, maybe a link to the corresponding `data` folder would be sufficient: https://github.com/swcarpentry/python-novice-inflammation/tree/gh-pages/data?

wking:
On Wed, Apr 08, 2015 at 10:22:53PM -0700, Bernhard Konrad wrote:
> I was more thinking about the use of the lecture material outside or
> after a workshop as a learning resource. For starters, maybe a link
[...]

PR #347, opened 2017-01-17
Clarifies location of working directory and data files, updates code accordingly.

BernhardKonrad:
Simplifies the text and location of files to make it easier for self-studying. Somewhat related to #60.


valentina-s:
Thanks for suggesting these clarifications. Although, I agree that in real life we should not work in data folder, based on an older discussion: #9  it was decided that SWC is taught from the data folder to avoid the complication coming from paths on different operating systems and different terminals (at least in the beginning, it still requires path-changing when using the code folder...). This is also consistent with how the R lesson is. So will close  for now until we decide to change this policy. Does @tbekolay agree?

tbekolay:
Yes, I think for the time being working in the data folder is easiest. As for changing the policy, perhaps opening an issue with a rationale and soliciting votes would be the way forward.

PR #349, opened 2017-01-24
Fixed incorrect line numbers in 08-defensive.md

walternathan6754:
This commit fixes the line numbers referred to in the normalize_rectangle function of defensive programming were incorrect. My guess is the comment line was added later, and the line numbers were not updated.

Also added the line numbers for the post-conditions "The post-conditions on lines 17 and 18...". I thought this would clarify to learners which assertions were preconditions and which ones were post-conditions.

### DIFF
--- a/_episodes/08-defensive.md
+++ b/_episodes/08-defensive.md
@@ -131,7 +131,7 @@ def normalize_rectangle(rect):
 ~~~
 {: .python}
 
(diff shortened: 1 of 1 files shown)


valentina-s:
Thanks for the correction! Maybe we should also add the line numbers to the code excerpt?

PR #352, opened 2017-02-04
Updating mailmap and AUTHORS before release

twitwi:


twitwi:
(some context about this PR)

The authors file is used to list all authors of the lesson, which currently means "people having commited to the repo". This file is used for the release of the lessons (where we get a DOI from Zenodo, publish on our site [1], etc).

We generate the authors file from `git log` by using the commiter names. However, for a same person we have commits showing different emails, different writings of the name, a pseudo instead of the name, etc. Git has the `.mailmap` mechanism [2] exactly to solve this issue by describing how commits should be patched. This file used to be filled manually (leading to useless entries, missing ones, errors) and it is now assisted by a script, leading to some notable cleanup.

[1] https://software-carpentry.org/lessons/previous/
[2] https://git-scm.com/docs/git-shortlog#_mapping_authors

.mailmap
```diff
@@ -1,114 +1,53 @@
...
+Ryan Neufeld <raeneufe@uwaterloo.ca> <metavee@users.noreply.github.com>
+Sarah Clayton <sarah.c.clayton@gmail.com>
+Sigrid Klerke <SigridK@users.noreply.github.com>
+Timothée Poisot <t.poisot@gmail.com> <tim@poisotlab.io>
+Toby Hodges <tbyhdgs@gmail.com>
+Trevor Bekolay <tbekolay@gmail.com>
+Valentina Staneva <valentina-s@users.noreply.github.com>
+Valentina Staneva <val_is_everywhere@yahoo.com>
+Willem Ligtenberg <willem@wligtenberg.nl>
+YourName <you@example.com>
```
kynan: suppose that one's a mistake

twitwi: There is actually a commit from "Your Name <you@example.com>".
I added this rewrite so that it can be detected later in the release process ("YourName" is single-word so it won't pass the "firstName lastName" check and will ring a bell in the release process).

twitwi:
Thanks for the ordering comment. I made this choice to limit the diff on some repositories where the AUTHORS were not fully ordered.
I'll update the PR to order them, it indeed makes more sense.

twitwi:
It is now sorted.
I'm taking the liberty to merge it for the release.

PR #353, opened 2017-02-06
Add an intermediate step in explaining how to call a function

lingling-CB:


valentina-s:
Thank you for the addition!

PR #355, opened 2017-02-18
Formatting in 02-loop.md in the "What's in a name?" callout is messed up

jarrah42:
The formatting in this callout doesn't render for me. I've updated it to what I think it should be.


tbekolay:
Thanks @jarrah42!

PR #359, opened 2017-03-08
Fix code zip

msleigh:
Fixes to make tutorials consistent with setup instructions. *code.zip unpacks into ./code dir rather than into ./, and paths consistently assume user is in python-novice-inflammation/data.


evanlinde:
+1 for this. This brings things into line with the setup page.

valentina-s:
Thanks for the synchronizing the paths!  I wonder if `python ../code/readings_04.py` works on Windows (it might depend on the terminal used). If not, we might need to add some clarification around it.

PR #361, opened 2017-03-16
Tweaking make help message

roopchansinghv:
Tweaking make help message to clarify when it is appropriate / necessary to rebuild figure page.


tbekolay:
Definitely makes more sense this way. Thanks!

PR #363, opened 2017-03-22
Update 02-loop.md

rulisp:
Add a bit of prose to the motivation for needing a loop construct. Remove the exponentiation example and replace with a loop over patient data similar to the previous exercise.


maxim-belkin:
Thanks, @rulisp! Could you please break this PR into a bunch of smaller ones? It will help us merge appropriate changes quicker.

Issue #366, opened 2017-03-28
06 - composing functions - refactor to build up a more complicated function


dstndstn:
Currently, the fahr_to_celsius function is composed of fahr_to_kelvin and kelvin_to_celsius, both of which involve the "magic number" 273.15 -- this seems like bad form; and fahr_to_kelvin includes both conversion functions, so the composition doesn't really make sense, since it "undoes" a conversion that was already done.

tbekolay:
You're right that the composed `fahr_to_celsius` isn't the most efficient way to do the conversion (both computationally and conceptually). But the point of the topic isn't to present efficient code, it's to introduce the concept of function composition. It's relatively easy to motivate composition when presenting this lesson by appealing to laziness -- "Who remembers the formula to convert from fahrenheit to celsius? How confident are you in that formula? Rather than looking it up to be sure, why don't we use the functions we've already made to save us the time of looking it up." This follows from the general idea that your time as a programmer is more valuable than wasting a few CPU cycles on a function call that could be avoided.

Currently we don't talk about "magic numbers" in the lesson. IMO, this is a bit of a premature optimization when instructing novice programmers -- yes, as you scale up, using variables for these numbers will make your code easier to read and easier to modify. But is that a necessary anxiety to give a novice programmer when they're just getting their feet wet? Further, 273.15 isn't an egregious example of a magic number as it will never need to be changed to something else. On the other hand, there is some precedent about talking about code style in callouts in the "What's in a name" callout in episode 2. I wouldn't be opposed to a callout in episode 6 about magic numbers, but I don't think I would change the functions that are there already.

dstndstn:
I'm concerned about the conceptual complexity, not the computational.  The way the functions are currently laid out, it's like 2-steps-forward-1-step-back to do fahr to celsius via fahr to celsius to kelvin plus kelvin to celsius.  Sure, I can see the "laziness" appeal, but I don't see this as a good example of building up complexity.

Let me put together a quick PR just to see what it looks like.  I think it'll be cleaner, but I've never taught this lesson so I could be wrong!

PR #367, opened 2017-03-29
06-func.md: simplify function composition example

dstndstn:
build fahrenheit-to-kelvin out of fahrenheit-to-celsius and celsius-to-kelvin

This also updates the figure illustrating the parts of a function definition.

This is for issue #366


valentina-s:
I like this simplification. I have to admit that I think I have taught it this way (without realizing) since it is more natural: as a European in US I always have to convert from F to C first... I will let @tbekolay decide whether it is worth updating.

maxim-belkin:
I like this change.
@dstndstn if you could rebase this PR on top of the current `gh-pages` branch that would be very much appreciated.

```
git rebase gh-pages composing-functions
```

maxim-belkin:
Let me, however, check if we can clean up SVG code first

maxim-belkin:
OK, let's work on SVG figures in a different PR.
Thanks, @dstndstn, for improving to the lesson!

PR #368, opened 2017-04-02
Generate Composite Statistics - solution

digitaltxtlab:
Added solution to Generate Composite Statistics (instructor check out procedure)


valentina-s:
Thanks a lot for adding this solution!

PR #369, opened 2017-04-03
More explanation for why we import individual libraries.

Matthew-Whitlock:


valentina-s:
Thank you for the addition.

PR #370, opened 2017-04-12
Update markdown references from relative SVG to PNG

louisvernon:
Github does not render relative SVGs: github/markup#556. This PR changes exiting markdown references from SVGs to PNGs, which are then rendered correctly. There was no PNG equivalent for "python-function.svg" so one was added to the repo.


rgaiacs:
@louisvernon Thanks for the pull request.

> Github does not render relative SVGs

This is true.

screenshot_2017-04-13_08-40-52

But our target is GitHub Pages and not GitHub. And SVG is render in GitHub Pages.

screenshot_2017-04-13_08-41-08

We want SVG because, primarily, they look good in high resolution displays and, secondly, they are easy to edit. So when we have a SVG we are going to use it!

A side note is that if we follow your logic that "GitHub does not render relative SVG" we should also change the markup that we are using because GitHub does not render

    ~~~
    weight_kg = 57.5
    print('weight in kilograms is now:', weight_kg)
    ~~~
    {: .python}

properly.

louisvernon:
Thank you for the detailed response. Your comment regarding SVGs working correctly within the target interface is well taken. I understand your prioritization of SVGs on GH Pages over working figures within the regular GitHub markdown renderer.

I do think that having a code block not formatted as desired outside of GH Pages is not as impact full as un-rendered figures. While investigating solutions I found several undesirable workarounds to get the SVGs rendered correctly, for example linking via https://rawgit.com/.

Hopefully GitHub will change this behavior in the future.

PR #374, opened 2017-04-26
Added R as another example of a 1-based indexing

timothymillar:
Added R to list of programming languages using 1-based indexing.
It seems logical to point out this difference for R in particular because it's such a commonly used and taught language.


valentina-s:
Yes, R will be more familiar to some folks. Thanks!

timothymillar:
@valentina-s could I please get this tagged with instructor-training, Thanks

valentina-s:
Done!

PR #375, opened 2017-05-04
Post-workshop fixes

nsoranzo:
The following fixes result from running this tutorial during a [Software Carpentry Workshop at the Earlham Institute](http://earlham.ac.uk/software-carpentry-workshop-2017) together with @froggleston and @tonietuk.

See individual commits for details.

### DIFF
--- a/_episodes/02-loop.md
+++ b/_episodes/02-loop.md
@@ -356,7 +356,7 @@ so we should always use it when we can.
 > ## Computing the Value of a Polynomial
 >
 > The built-in function `enumerate` takes a sequence (e.g. a list) and generates a
-> new sequence of the same length. Each element of the new sequence is a pair composed by the index
+> new sequence of the same length. Each element of the new sequence is a pair composed of the index
 > (0, 1, 2,...) and the value from the original sequence:
 >
 > ~~~
--- a/_episodes/02-loop.md
+++ b/_episodes/02-loop.md
@@ -336,7 +336,8 @@ so we should always use it when we can.
 
 > ## Reverse a String
 >
-> Write a loop that takes a string,
+> Knowing that two strings can be concatenated using the `+` operator,
+> write a loop that takes a string
 > and produces a new string with the characters in reverse order,
 > so `'Newton'` becomes `'notweN'`.
 >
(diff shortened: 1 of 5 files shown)


_episodes/08-defensive.md
```diff
@@ -531,11 +531,13 @@ This violates another important rule of programming:
 > >
 > > def range_overlap(ranges):
 > >     '''Return common overlap among a set of [low, high] ranges.'''
+> >     if not ranges:
```
tbekolay: What do you think about `if len(ranges) == 0` instead? I know experienced Python users like the simplicity of this, but to a novice it feels more clear that an empty list has length 0.

_episodes/02-loop.md
```diff
@@ -355,8 +356,8 @@ so we should always use it when we can.
 > ## Computing the Value of a Polynomial
 >
 > The built-in function `enumerate` takes a sequence (e.g. a list) and generates a
-> new sequence of the same length. Each element of the new sequence contains the index
-> (0,1,2,...) and the value from the original sequence:
+> new sequence of the same length. Each element of the new sequence is a pair composed by the index
```
tbekolay: Nitpick: "composed by" -> "composed of"

_episodes/08-defensive.md
nsoranzo: On the other hand `not ranges` also handles the case where `ranges` is `None`.

tbekolay: This is true, but not a situation I suspect our novice students will run into as I'm not sure we ever talk about `None` in the lesson. The more likely situation is they call it and pass nothing, which raises a `TypeError`. That said, if you strongly prefer to keep it this way I can probably live with it.

nsoranzo: We talk about `None` a bit before this exercise:
```
`None` is built into Python, and means "nothing here".
```

What about I add a comment:
```python
    if not ranges:
        # ranges is None or an empty list
        return None
```

tbekolay: > What about I add a comment

Yeah, that sounds great, thanks!

tbekolay:
Thanks for the improvements!

PR #377, opened 2017-05-08
Add code examples to ch-ch-ch-changes

n0mjs710:
Demonstrate two variables assigned to the same list where changing one
changes both. Demonstrate making a copy of a list with a new variable
name.


tbekolay:
Thanks, I think the addition of an example here will help understanding!

PR #383, opened 2017-06-19
Fix code block color

rgaiacs:
We use box color to help learners to know which block they should **type** and run on their Jupyter Notebook or Python console. This pull request fix the color of one block that learners should not type.

# Before

screenshot_2017-06-19_14-34-09

# After

screenshot_2017-06-19_14-35-48


valentina-s:
Thanks for catching that!

PR #386, opened 2017-06-20
Add example diagram on zero index arrays to 01-numpy python lesson

TabletopWhale:
I added an example diagram to the first discussion of zero indexing in python. There are other diagrams later on in different lessons, but I thought it might be helpful to have a simple visual representation of the concept when it is first introduced.

python-zero-index


valentina-s:
Thanks a lot for creating this diagram!

PR #387, opened 2017-06-28
Better solution for pre- and post-conditions

### LINKED ISSUE #330: Error in 08-defensive post-condition exercise: min < mean < max
In the "Pre- and Post-conditions" exercise, the solution isn't quite correct:

```python
assert numpy.min(input) < average < numpy.max(input), 'Average should be between min and max of input values'
```
This doesn't hold when all input values are equal.

The correction is easy (I can PR if folks agree):
```python
assert numpy.min(input) <= average <= numpy.max(input), 'Average should be between min and max of input values'
```

Side note: since `input` is python built-in function, I'd suggest renaming the variable (say to `values`).  It's not wrong as is, but it's bad style.

pllim:
Post-condition now takes account of average of a list with single value. Changed "input" to "input_list".

Fix #330 

c/c @bmcfee

### DIFF
--- a/_episodes/08-defensive.md
+++ b/_episodes/08-defensive.md
@@ -462,9 +462,9 @@ This violates another important rule of programming:
 > > ## Solution
 > > ~~~
 > > # a possible pre-condition:
-> > assert len(input) > 0, 'List length must be non-zero'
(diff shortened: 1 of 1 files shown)


valentina-s:
Thanks for the fix!

Issue #281, opened 2016-07-07
Check/remove unexpected images


gvwilson:
`make lesson-check` reports:

Unexpected image files: 03-loop_4_0.png, ave-inflammation-over-time.png, combined-inflammation-2.png, combined-inflammation-per-day.png, initial-heat-map.png, ipython-notebook.png, loop-inflammation-01.png, loop-inflammation-02.png, loop-inflammation-03.png, loops_image.svg, max-inflammation-over-time.png, min-inflammation-per-day.png, python-call-stack-01.png, python-call-stack-02.png, python-call-stack-03.png, python-call-stack-04.png, python-call-stack-05.png, python-call-stack-06.png, python-call-stack-07.png, python-flowchart-conditional.svg, python-flowchart-nested-loops.png, python-flowchart-nested-loops.svg, python-operations-across-axes.svg, python-overlapping-ranges.png, python-sticky-note-variables-01.png, python-sticky-note-variables-02.png, python-sticky-note-variables-03.png, software-carpentry-banner.png

PR #393, opened 2017-07-30
Remove unexpected image files as reported by make lesson-check in iss…

Doomslang:
…ue #281: Check/remove unexpected images #281. Verified that none of the reported files are referenced in any of the lessons.


valentina-s:
Thanks for checking and cleaning all these files!

PR #396, opened 2017-08-06
Add input example to normalize_rectangle

rrlove:

---

Currently the normalize_rectangle function is somewhat confusing. Putting an example of the input format, as described in the text, in the docstring is both an example of good programming practice and makes the exercise more straightforward.

---


valentina-s:
That is a good clarification. Thanks!

PR #397, opened 2017-08-16
Change the swap into a tuple unpack

ianhbell:
Personally, I always write the swap as ``v1,v2 = v2,v1``, and I don't see the need here for creating a list; it complicates things without reason.  Besides, strings are immutable, so perhaps also better to go with a immutable container.

### DIFF
--- a/_episodes/03-lists.md
+++ b/_episodes/03-lists.md
@@ -496,7 +496,7 @@ Omitting ending index: ["sep", "oct", "nov", "dec"]
 > Compare it to:
 >
 > ~~~
-> left, right = [right, left]
+> left, right = right, left
 > ~~~
 > {: .python}
 >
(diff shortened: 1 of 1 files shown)


tbekolay:
I agree that in the real world tuple unpacking is more common, though it's mostly for syntax reasons (no bracket is better than bracket!) We actually used to use tuples in this example, but switched to lists in #346 because we had not talked about tuples in the lesson up to this point. Better to introduce one topic (unpacking) instead of two (unpacking, oh and also tuples). For that reason, I'm going to close this PR, but thanks for the contribution 

Also, to get uber technical / nitpicky, the creation of the list is no more effort than the creation of a tuple in `v1, v2 = v2, v1`. Python happens to automatically call `tuple` in that line, but you could argue that rather than complicating things with the list, we are instead uncomplicating things by being explicit about what is happening.

Speed difference is negligible:

```ipython
In [6]: def swap_tuple():
   ...:     a, b = 2, 3
   ...:     a, b = b, a
   ...: 

In [9]: %timeit swap_tuple()
73.6 ns ± 0.587 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [10]: def swap_list():
    ...:     a, b = [2, 3]
    ...:     a, b = [b, a]
    ...:     

In [11]: %timeit swap_list()
71.9 ns ± 0.384 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

Anyway, for the reasons above I'm going to close this PR unmerged. Thanks for the PR regardless!

ianhbell:
Fair enough I guess.  I see your point about not introducing something that we had not already discussed…

PR #399, opened 2017-08-20
Add More Learning objectives for Episode 03 on Learning Python Lists

dddragons:
---
This contribution is for instructor training.
---
After reading through the episode 03 about python lists, I feel that the episode teaches much knowledge about using lists. Some topics covered by the episodes are not reflected by the current learning objectives. Therefore, several measurable learning objectives are added to the episode.


valentina-s:
Thank you for the contribution! Yes, this list of objectives is more thorough.

PR #407, opened 2017-09-25
Use a descriptive name for sd plot example

mawds:
Change the name of the variable for the standard deviation plot example, so that it reflects what it contains.


valentina-s:
Good catch! Thanks for the fix!

Issue #404, opened 2017-08-28
Unwanted __MACOSX folder in python-novice-inflammation-code.zip


pllim:
I followed instructions at http://swcarpentry.github.io/python-novice-inflammation/setup/ and downloaded `python-novice-inflammation-code.zip` to my Linux machine. Then, I unzipped it as follows:
```
unzip python-novice-inflammation-code.zip
```
Now, I see the following in my working directory:
```
code/
__MACOSX/
python-novice-inflammation-code.zip
```
Here is the content of `__MACOSX`:
```
code/
  ._argv_list.py
  ._arith.py
  ._check.py
  ._count_stdin.py
  ._errors_01.py
  ._errors_02.py
  ._gen_inflammation.py
  ._line_count.py
  ._my_ls.py
  ._readings_01.py
  ._readings_02.py
  ._readings_03.py
  ._readings_04.py
  ._readings_05.py
  ._readings_06.py
  ._readings_07.py
  ._readings_08.py
  ._readings_09.py
  ._rectangle.py
  ._sys_version.py
._code
```
Is there a reason you include this `__MACOSX` folder in your distribution?

rgaiacs:
`__MACOSX` is a hidden folder used by macOS but visible on other OS. Ideally the zip file would be generated by a continuous integration system but we don't have this flexibility with GitHub Pages.

valentina-s:
I think I removed it in this version: https://github.com/swcarpentry/python-novice-inflammation/blob/fix_zipfile/code/python-novice-inflammation-code.zip

Can you double check if it cleaned (I am on Mac so it is invisible to me)?

amjames:
@valentina-s the `__MACOSX` folder only appears when the archive is unpacked at the command line.  If you unpack the zip file using the `unzip` command line utility, even on OSX it will appear, if you unpack by clicking on the zip in finder it does not. However creating an archive without the `__MACOSX` folder does not prevent you from unpacking it in this way. The same behavior is observed in the updated version you linked the only difference is that the `code` directory has slightly different permissions.

Actually the `__MACOSX` folder only appears in the archive when it is created in finder in the first place. If you create the archive using the `zip` utility at the command line it can be extracted at the command line on OSX/linux without the extra folder appearing *and* can still be opened by clicking on it in finder. I have no idea why doing the compression with finder puts it there but that is what I found. I have submitted a PR (#409) that fixes this.

amjames:
[...]

Issue #405, opened 2017-08-28
code and data ZIP files have different permissions


pllim:
After I unzipped both `python-novice-inflammation-code.zip` and `python-novice-inflammation-data.zip` on my Linux machine, I noticed that they have inconsistent file permissions:
```
drwx--S---. 2 username usergroup 2048 Jul 31 04:27 code/
drwxr-sr-x. 2 username usergroup 1024 May 10  2015 data/
```

valentina-s:
Thanks for spotting this problem. I updated the permissions of the code folder (to all). Can you check if it works on Linux: https://github.com/swcarpentry/python-novice-inflammation/blob/fix_zipfile/code/python-novice-inflammation-code.zip

I can also add more permissions for the data folder.

pllim:
Thanks, but unfortunately now I cannot even unzip it. `unzip` command on Linux now gives me this error:
```
Archive:  python-novice-inflammation-code.zip
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
unzip:  cannot find zipfile directory in one of python-novice-inflammation-code.zip or
        python-novice-inflammation-code.zip.zip, and cannot find python-novice-inflammation-code.zip.ZIP, period.
```

The same command worked on the original archive.

valentina-s:
Not sure what is wrong with that one as it opens on my laptop. Can you try this one submitted by @amjames https://github.com/swcarpentry/python-novice-inflammation/pull/409/files before I merge it. It removes the MACOSX folder and the permissions after unzipping on my laptop are drwxr-xr-x for the folder and -rw-r--r-- for each file. What permissions do you think we should aim for?

pllim:
@valentina-s , yes, the version from @amjames works fine and has no extra `__MACOSX` folder. Thank you!

PR #409, opened 2017-09-27
Removes the unnecessary __MACOSX directory from the code zip archive.

amjames:
The zip archive containing code for this lesson contains an unneeded `__MACOSX` folder which is unpacked alongside the `code` directory when the zip archive is unzipped at the command line.

By extracting the archive before and after it's last change in 57f69c73e4b282576c8aa9bd301642634dcdeadd I suspect that the archive in that commit was created using finder, and not by using the `zip` utility. See discussion about this [on SO](https://stackoverflow.com/questions/10924236/mac-zip-compress-without-macosx-folder)

The `__MACOSX` directory is added to the archive when it is created from finder in MacOS, however an archive that does not contain it (ie one created using the `zip` command line utility) can still be extracted by double clicking the `.zip` file in finder without issue. 

The only change introduced by this PR is that linux users (and osx users who use `unzip` in the terminal) won't see the problem described in #404.


pllim:
This fix both #404 and #405 for me. Thanks!

valentina-s:
Great! They will get closed automatically when I merge it.

PR #412, opened 2017-10-02
Fix output of %whos, numpy is not imported yet

gjbex:
The output of the `%whos` magic shows that the `numpy` module is loaded, while that actually only happens in the next section.  In this branch, I've deleted the output line that shows the module.


tbekolay:
This makes sense, thanks!

PR #414, opened 2017-10-04
Suggest changing list example to string example in  02-loops.md, sect…

wpk-nist-gov:
Suggest changing list example to string example in  02-loops.md, section "What's in a name"

The original 02-loop section "What's in a name" uses an example using a for loop over a list.
As lists aren't introduced until section 03-lists, I suggest changing this example to a loop over
a string.  Note that there is an excercise "Computing the Value of a Polynomial" in 02-loops that also
uses a list, that may be moved to 03-lists as well.

### DIFF
--- a/_episodes/02-loop.md
+++ b/_episodes/02-loop.md
@@ -157,25 +157,7 @@ command to signify the end of the loop body (e.g. `end for`); what is indented a
 > ## What's in a name?
 >
 >
-> In the example above, the loop variable was given the name `char` as a mnemonic; it is short for 'character'. 'Char' is not a keyword in Python that pulls the characters from words or strings. In fact when a similar loop is run over a list rather than a word, the output would be each member of that list printed in order, rather than the characters.
->
-> ~~~
-> word = 'oxygen'
-> for char in word:
->    print(char)
-> ~~~
-> {: .python}
->
-> ~~~
-> o
-> x
-> y
-> g
-> e
-> n
-> ~~~
-> {: .output}
->
+> In the example above, the loop variable was given the name `char` as a mnemonic; it is short for 'character'. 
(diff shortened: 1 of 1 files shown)


tbekolay:
Hm. So, the PR as it is now doesn't make much sense because the callout in which the example resides explicitly talks about how you can loop over a list just as you can a string -- looping over the string `'oxygen'` is already done earlier in this episode, so doing it a second time doesn't seem to add much.

You could argue that the callout should therefore be moved from this episode to 03-lists. I think that would be okay, but it doesn't flow as well as it does here. I'm okay with moving it if it can still flow well in the other episode, but that will require some thought about where to put it and how to transition to it.

While I agree that the ideal case is that we do not use something that has not yet been introduced, I feel that we can be a little more relaxed on that ideal within callouts. I do not think of callouts as being core elements of each episode; I might not say them out loud when teaching. I see them as useful tidbits of extra information or additional context. Given that interpretation, I'm okay with the callout referencing something that will be more fully introduced in a later episode, assuming that the example is not super confusing without the introduction. In this case, I think it's relatively easy to see what's going on, but like I said, if the consensus agrees that this is confusing then I would support moving it to episode 03-lists.

wpk-nist-gov:
Sorry, I failed to change the text above the first example in the callout.  I fixed this.

Right now, the callout introduces two ideas: 

1. looping over lists pulls out items of the list
2. the name of the the loop variable is up to the user.  

I'd argue that (2) is the important point in this section.  (1) is introduced at the beginning of 03-lists already.

tbekolay:
Right you are, I like shortening this callout. Merging now, thanks!

wpk-nist-gov:
Thank you!

PR #415, opened 2017-10-11
Updated Python Loop Lesson to Address Indexing

rspavel:
Previous version of python loop lesson introduced ambiguity in how to
index into an array/list/string in Python. Updated to make it clear that
Python indexes from zero so as to not introduce this ambiguity


valentina-s:
Thank you for the suggestion @rspavel. Since the numbers in the diagram refer to the loop cycle(not the index) I think it is correct for them to be 1-6.

annefou:
@valentina-s I think you are correct too. Shall we close this PR?

maxim-belkin:
@rspavel is this PR part of your instructor checkout process?

maxim-belkin:
By the way, I concur with @valentina-s and @annefou.

maxim-belkin:
I think the point here is that the **first** **\*index\*** in an array is **zero**  but if you want to number iterations of a **\*loop\*** you would still start from **one** (first iteration, second iteration, ...)

I think a PR to make the word "loop" bold would be reasonable. I also think it would be better to say `iteration of a loop` rather than just a `loop`.

Thank you, @rspavel!

PR #417, opened 2017-10-11
Modify to replace initial section regarding Python variable assignmen…

markcmiller86:
- Replace initial section regarding Python variable assignment with review of some shell commands to browse contents of inflammation data files and orient learners with that data.
- Introduce concept of Python variable and assignment as part of the numpy.loadtxt() command.
- Explain that Python is powerful both as a *language* and as a set of community developed *libraries*.
- Explain that lesson begins with *libraries* and not *language*.

Longer rationale...the lesson begins with somewhat arbitrary step of defining and assigning a python variable for `weight_kg` and `weight_lb`. There is significant space used elaborating on Python variables and assignment here but without any connection to the dataset to be analyzed. I felt the concept of varaible assignment is easily handled at the time `numpy.loadtxt()` is being introduced. I include a couple of lines elaborating that concept there.

Beginning the lesson with a review of shell commands makes the lesson dovetail nicely with shell lesson (which is often taught ahead of this python less in the shell, python git trio). More importantly, it allows learners to underestand and orient themselves with the data in the files they are loading into Python so that once there, they cann connect the dots back to what they see through `head` shell commands, etc.

Finally, previous revision offered little in the way of explanation that this lesson really starts in the _deep end_ of python in the sense that learners are first being exposed to _library_ functionality before they learn much of the basic _language_. I felt it was important to explain this and motivate it slightly.

### DIFF
--- a/_episodes/01-numpy.md
+++ b/_episodes/01-numpy.md
@@ -28,130 +28,102 @@ keypoints:
 - "Use `numpy.mean(array, axis=0)` or `numpy.mean(array, axis=1)` to calculate statistics across the specified axis."
 - "Use the `pyplot` library from `matplotlib` for creating simple visualizations."
 ---
-In this lesson we will learn how to manipulate the inflammation dataset with Python. But before we discuss how to deal with many data points, we will show how to store a single value on the computer.
+In this lesson we will learn how to manipulate the inflammation dataset with Python.
 
-
-The line below [assigns]({{ page.root }}/reference/#assignment) the value `55` to a [variable]({{ page.root }}/reference/#variable) `weight_kg`:
-
-~~~
-weight_kg = 55
-~~~
-{: .python}
-
-A variable is just a name for a value,
-such as `x_val`, `current_temperature`, or `subject_id`.
-Python's variables must begin with a letter and are [case sensitive]({{ page.root }}/reference/#case-sensitive).
-We can create a new variable by assigning a value to it using `=`.
-When we are finished typing and press Shift+Enter,
-the notebook runs our command.
-
-Once a variable has a value, we can print it to the screen:
+First, lets familiarize ourselves with the raw,
+[comma-separated-values (csv)](https://en.wikipedia.org/wiki/Comma-separated_values)
+data files by reviewing some shell commands to browse their contents. From a shell window,
+ensure you are in the `python-novice-inflammation` directory...
 
 ~~~
-print(weight_kg)
+$ cd ~/Desktop/python-novice-inflammation/
+$ pwd
+/Users/nelle/Desktop/python-novice-inflammation
+$ ls
+code	data
+$ cd data
+$ ls
+inflammation-01.csv	inflammation-04.csv	inflammation-07.csv	inflammation-10.csv	small-01.csv
+inflammation-02.csv	inflammation-05.csv	inflammation-08.csv	inflammation-11.csv	small-02.csv
+inflammation-03.csv	inflammation-06.csv	inflammation-09.csv	inflammation-12.csv	small-03.csv
+$ wc -l infl*
+60 inflammation-01.csv
+60 inflammation-02.csv
+60 inflammation-03.csv
+.
+.
+.
 ~~~
-{: .python}
+{: .shell}
 
-~~~
-55
-~~~
-{: .output}
-
-and do arithmetic with it:
-
-~~~
-print('weight in pounds:', 2.2 * weight_kg)
-~~~
-{: .python}
-
-~~~
-weight in pounds: 121.0
-~~~
-{: .output}
-
-As the example above shows,
-we can print several things at once by separating them with commas.
-
-We can also change a variable's value by assigning it a new one:
-
-~~~
-weight_kg = 57.5
-print('weight in kilograms is now:', weight_kg)
-~~~
-{: .python}
+Notice that the output from the word-count command, `wc -l`, shows that each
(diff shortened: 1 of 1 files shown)


valentina-s:
Thanks a lot for the proposed modifications.

The shell commands provide a good review but I feel like they will take a lot of time, and this lesson is already extremely long. I think a quicker approach to this is to ask the students to look at the file (with cat, nano or their favorite editor) so that they have an idea of what is inside. This could be done when the .csv files are introduced. Thus we will keep the prerequisites of the lesson to minimum (in case it is taught without a shell lesson). 

I am reluctant about completely removing the simple variable discussion, although I agree it is disconnected from the dataset. It may really hurt the people who have never programmed before. I have personally seen beginners struggling with the later exercises because they do not fully understand what is stored where. I think the example with the relationship between the units aims to explain the difference with variables in Excel (which get linked by the assignment operator).  It will be great if we have an example of converting some units related to the inflammation dataset: but what are the units there??? Would you be interested in coming up with such an example? Maybe a biologist can help out here. 

I am ok with adding explanation of the approach to teach Python. Agree that the word analogy might not be that clear, but I would leave some short explanation of what a library is  geared toward complete beginners (a toolbox of specialized tools?).  Some of the motivation can be added to the intro section: http://swcarpentry.github.io/python-novice-inflammation/. I think the existence of powerful Python tools can be a bullet point by itself. What do you think? 

I will mark the PR for discussion for now to get suggestions from others. It is best to submit individual changes in separate PRs.

tbekolay:
One thing that I worry about in regards to this change is that Software Carpentry (as far as I remember) does not require that you teach the shell lesson before this lesson; and, even if it did, we do not list the shell lesson in our prerequisites:

> Learners need to understand the concepts of files and directories (including the working directory) and how to start a Python interpreter before tackling this lesson. This lesson references the Jupyter (IPython) Notebook although it can be taught through any Python interpreter. The commands in this lesson pertain to Python 3.

or on [our setup page](http://swcarpentry.github.io/python-novice-inflammation/setup/).

Further, I see this lesson as also being a useful resource for people learning online; it should be able to stand on its own as a useful introduction to Python without necessarily needing to know a lot about the shell.

I do agree that we do not introduce the inflammation data well, so more explanation of it is warranted, but perhaps not at the expense of variable assignment, though I agree that it is a slightly awkward thing to discuss in this topic.

It would be great to have all these changes in separate PRs so that we can discuss them separately :+1:

markcmiller86:
Regarding shell as a prerequisite for python, I agree thats an issue with this PR as it winds up making shell a prerequisite where before it may not have been. That said, in almost all of the contexts in which I have participated in SC teleconferences so far, I always see the trio, Shell, Git and Python packaged together. I know that isn't always the case but it seems pretty common, at least to the casual observer. In that context, it seems like a really good fit. On a related point, while I don't have a feel for this myself, I wonder how many python learners are likely to be shell-savvy? 

Is there any way to support optional, parallel _scenes_ in an _episode_? If this or the original content were available as separate, selectable intro _scenes_, then instructors would have the option of choosing which might be the best for their training.

Regarding putting all these changes in _sparate PRs_...I think the challenge there is understanding what everyone might regard as a _separable_ change. I mean, is a saparate change an individual line, a paragraph, a whole sub-section or what? Another related issue is whether the merged result can still make sense if only some of those PRs are actually accepted? In general, I think it would be hard to craft this change into multiple PRs such that any combination of those accepted still winds up making sense and good flow. Finally, that represents a non-trivial amount of additional work for someone like me who is trying to contribute and submit such changes.

If we forgoe the shell-based intro, I almost think it makes the most sense to still start the lesson talking about python as a language and python as a set of libraries and then begin the concept of variable assignment with `numpy.loadtext()` storing the result of an operation performed by a library and maybe _after_ that, saying a few more details about variables with some examples.

tbekolay:
I agree that the Python lesson almost always comes after the shell lesson in SWC workshops, but that doesn't help those that are looking at the lesson online outside of an SWC workshop context. While it might be hard to believe, I would guess that far less than half of Python users are shell-savvy. Most people use Windows, which requires some special effort to install a Bash shell, and even more effort to use that Bash shell effectively since most Linux programs can't be easily installed. I would guess that most people use Python through an IDE or only in the notebook environment through Anaconda.

I don't believe there's any precedent for alternative parallel paths through an episode... it sounds like a cool idea if it were implemented well, but I do wonder if it's just as effective to fork this lesson as another way to provide an alternative.

markcmiller86:
> @tbekolay I don't believe there's any precedent for alternative parallel paths through an episode... it sounds like a cool idea if it were implemented well, but I do wonder if it's just as effective to fork this lesson as another way to provide an alternative.

Can you point me to examples of where providing alternatives is handled? I mean, is there precedent for this already and if so, how is it handled? And, are such alternatives _affiliated_ with SWC or something else? Finally, can you suggest any alternative lighter weight than maintaining a separate fork of the entire lesson?

valentina-s:
Unfortunately, we do not have an existing infrastructure for handling alternative versions.  We have discussed allowing for adding extra examples or discussions within the extras section:
https://github.com/swcarpentry/python-novice-inflammation/tree/gh-pages/_extras.

One possibility is to create a supplementary.md file and put the shell-to-python transition section in it. We can link to it from the instructors note as additional resources. Clearly, it will be hard to ensure that this will stay synchronized with the lesson, but it can be aimed at instructors who might decide that this approach is more appropriate for their audience.

PR #418, opened 2017-10-22
Update 02-loop.md with enumeration description

FelixLiu-SF:
Updated text describing enumeration with a clearer description.


valentina-s:
Thanks!

PR #421, opened 2017-10-25
02-loop.md: add a new exercise on loops

anuj2054:
I added a for loop question in there.

### DIFF
--- a/_episodes/02-loop.md
+++ b/_episodes/02-loop.md
@@ -285,6 +285,27 @@ so we should always use it when we can.
 > {: .solution}
 {: .challenge}
 
+
+
+
+> ## Understanding the number of loops
+>
+>word = 'oxygen'
+>for char in word:
+>    print(char)
+>In the above code, how many times is the loop repeated . 
+>A. 3 
+>B. 4 
+>C. 5
+>D. 6 
+>
+> ## Solution
+> 6
+> {: .solution}
+{: .challenge}
+
+
+
 > ## Computing Powers With Loops
 >
 > Exponentiation is built into Python:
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Thanks for your contribution, @anuj2054 ! I think this is a simple and important exercise. Is this PR part of your instructor checkout process?

@MikeTrizna what are your thoughts on this one?

MikeTrizna:
@anuj2054 I really like how you labelled the different possible answers for this additional Multiple Choice Question, and what each answer tells the instructor. Unfortunately I think your proposed question is too close to the material immediately before the question section. It even uses the same variable name and word. Maybe try changing both of those to ensure that students are able to transfer the lessons they learned to a new example?

_episodes/02-loop.md
```diff
@@ -285,6 +285,33 @@ so we should always use it when we can.
...
+
+> ## Understanding the number of loops
+>
+> ~~~
+> word = 'oxygen'
+>for char in word:
+>    print(char)
+> ~~~
+> {: .python}
+>In the above code, how many times is the loop repeated ?
```
maxim-belkin: Do not add space before the `?`.

annefou:
I removed the blank before ? @maxim-belkin  Should we merge this contribution now?

maxim-belkin:
Yes. I also like the explanation of possible misconceptions that Anuj has in the first commit:

> A. 3 - probably a guess, indicative of confusion
B.4 - thinks 'word' is subject of print
C. 5- possible confusion about index starting at 0
D.6  - probably understands the loop process

Let's merge this one as is and consider adding the above text somewhere in the instructor notes...

Issue #221, opened 2016-01-27
'Tabs and Spaces' example does not actually include tabs


adamobeng:
The "Tabs and Spaces" callout in lesson 7 includes a code example where supposedly "the first two lines are using a tab for indentation, while the third line uses four spaces". In fact, all three lines use spaces for indentation.

Someone copying this example would not be able to reproduce the error.

I'm opening an issue rather than just fixing it, because:
1. this might be a deliberate decision (the markdown file also has spaces, so this is not just a conversion problem)?
2. fixing it is non-trivial, given that pandoc converts tabs to spaces by default (the '--preserve-tabs' option would have to be added to the pandoc call in the Makefile).

gvwilson:
Yeah, this is hard to fix given that we're using Markdown...

yeemey:
I typed out the code example as described in Jupyter, and cannot reproduce the ```TabError```. Trying this out in the command line gives me an ```IndentationError``` instead. 

Either we revise the example to reproduce the error (I tried some random combinations of tabs/spaces, and couldn't get it), or remove it and just note that it's not good practice to mix tabs and spaces?

From the [Python 3 docs](https://docs.python.org/3/reference/lexical_analysis.html#indentation):
> Tabs are replaced (from left to right) by one to eight spaces such that the total number of characters up to and including the replacement is a multiple of eight (this is intended to be the same rule as used by Unix).
> ...
> Indentation is rejected as inconsistent if a source file mixes tabs and spaces in a way that makes the meaning dependent on the worth of a tab in spaces; a TabError is raised in that case.

maxim-belkin:
Hey @yeemey!
Could you please be more specific as to which line in which file you're talking about. Thank you!

yeemey:
Hi @maxim-belkin,
It's the [episode 7 lesson on errors and exceptions](http://swcarpentry.github.io/python-novice-inflammation/07-errors/).

It's ~line 182 in my forked repo with 07-errors.md.

tbekolay:
> I typed out the code example as described in Jupyter, and cannot reproduce the TabError

You have to copy/paste the example, not type it out manually.

tbekolay:
And actually this issue should be fixed now as I did change the HTML to have explicit tabs in there; I just copy/pasted and it worked for me.

PR #422, opened 2017-10-26
Issue #221 - add tab to 'Tabs and Spaces' example

tylerkelly13:
Based on info from this site: https://daringfireball.net/projects/markdown/syntax#html
I have replaced the markdown code block with html code (copied the html source from http://swcarpentry.github.io/python-novice-inflammation/07-errors/ and added a tab). The indentation has been increased on all lines as html tabs are 8 characters wide, not 4 and it is hard to change the html tabs to 4 characters.

### DIFF
--- a/_episodes/07-errors.md
+++ b/_episodes/07-errors.md
@@ -192,12 +192,11 @@ it *always* means that there is a problem with how your code is indented.
 > the first two lines are using a tab for indentation,
 > while the third line uses four spaces:
 >
-> ~~~
-> def some_function():
->     msg = "hello, world!"
->     print(msg)
->     return msg
-> ~~~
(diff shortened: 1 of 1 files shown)


tbekolay:
Thanks, this is helpful! I realized when looking at the diff, however, that Python 3 (which we are using in this lesson) no longer allows you to mix tabs and spaces. So I'm going to merge this, then reword the callout to indicate that this is no longer possible in Python 3. I'll keep the code sample though :)

tylerkelly13:
If the code is used pasted into a .py then run, the error will still come up (I'm pretty sure). Command line will run it with no error, and some editors will convert the tabs or spaces on save to prevent the error.

Issue #358, opened 2017-03-06
Key points in python inflammation lesson not aligned with content


stevehadd:
The key points (KPs) in several sections of the Programming with Python (Inflamation data) lesson do not align with the content of the section.
- 03: Storing Multiple Values in Lists
 - "Lists are indexed and sliced in the same way as strings and arrays." Learners will often use the keypoints as a "cheatsheet" to remind themselves of what they have learnt. It would be more useful to include an example of how to index and slice a list rather than refering them to another lesson.
 - The section mentioned that lists can contain any python object, including lists. This is probably worth including as KP. 
- 05: Making Choices
 - "Nest loops to operate on multi-dimensional data." Not found in this section (or any section!), should be removed.
 - "Put code whose parameters change frequently in a function, then call it with different parameter values to customize its behavior." Not found in this section, should be removed as dealt with in section 6,
 - As the literals True and False are introduced here, they should be included as a KP.
- 06: Creating Functions
 - "Define a function using def name(...params...)." Would be more useful when consulted by a learner if this look like a valid function definition e.g. 'def function_name(parameter1,parameter2)'
 - Call a function using name(...values...). Would be more useful when consulted by a learner if this look like a valid function call e.g. 'function_name(value1,value2)'

valentina-s:
Thanks! It will be great if the key points get revised to reflect the content. PR's welcome!

yeemey:
Hi, I opened a PR for this, addressing all the suggested changes except one re: True/False booleans in ep. 05. This appears in the challenges section, and instructors sometimes skip some of them. IMO, might not be necessary in Key Points, but I defer to @valentina-s. I'm happy to add that change if you think it's better.

valentina-s:
Thanks! Looking now at the lesson there is no explicit statement really explaining that something like 1 > 0 returns the value True. I think it is worth adding a line about that. Then we can also add the keypoints.

vinisalazar:
Apparently this change is now in Episode 7, it was introduced by 9ff6b19.

vinisalazar:
I expanded the example of True and False by explaining that 0 and None are evaluated as False. That's a helpful Python feature to know about, IMO.

Best,
V

paulmaxus:
I've found this issue under the _help wanted_ label, but it appears that all the original suggestions for corrections of the key points sections have been adopted. Should we remove the label and close the issue?

noatgnu:
[...]

PR #425, opened 2017-11-01
Revised KPs in eps. 3, 5, 6 (Issue #358)

yeemey:


valentina-s:
@yeemey Thanks a lot for correcting the key points! Could you also add the KP "Put code whose parameters change frequently in a function, then call it with different parameter values to customize its behavior." as a key point to the function lesson (06)?

yeemey:
thanks, @valentina-s !

PR #427, opened 2017-11-09
[FIX] the list of lists image

Shotgunosine:
The image in python's list tutorial isn't a great representation of a list of lists. The physical representation of a list isn't consistent (pepper shaker, pepper packet). I made a new image using a table as the physical representation of a list, so a list of lists is three tables stacked on a table.


valentina-s:
Thank you for suggesting to improve the representation. I have been thinking that the last picture of the pepper example might have been better if it were just one pepper crumb. 

Could you also add a short caption for your explanation? I first did not see the table underneath so I was confused why the [x[0]] was two crossed tables. I even think that maybe skipping [x[0]] will make things more clear for the beginner. I will label the PR for discussion so we hear what others think about this representation.

maxim-belkin:
I'm closing this as this is a rather old PR and there was no discussion. We've updated the image since then.

Thanks, @Shotgunosine!

Issue #328, opened 2016-11-11
Python 3 floating point rounding


hpourreza:
When I run `2.2*55` in Python 3, I get `121.00000000000001` instead of `121.0` that is on the course website (01_numpy). I think the output on the course website is from Python 2 that prints `121.0`.

rgaiacs:
@hpourreza Thanks for the issue. Could you send a pull request to fix it?

neon-ninja:
On my machine, I get the same result in python2 and python3. I think the difference occurs depending on whether you're running a 32-bit or 64-bit version of python - see https://docs.python.org/3/tutorial/floatingpoint.html

python 2.7.10 on ideone: http://ideone.com/WYcuze = 121.0
python 2.7.10 on repl.it: https://repl.it/F26E/0 = 121.00000000000001
python 3.4 on ideone: http://ideone.com/DCQgI6 = 121.00000000000001

PR #429, opened 2017-11-12
01-numpy.md: change weight to avoid numerical errors

joshkyh:
Proposed fix to floating point print by through native function format #328 

---
_edited by @maxim-belkin to remove PR template_

### DIFF
--- a/_episodes/01-numpy.md
+++ b/_episodes/01-numpy.md
@@ -67,7 +67,7 @@ print('weight in pounds:', weight_lb)
 {: .python}
 
 ~~~
-weight in pounds: 120.0
+weight in pounds: 132.0
 ~~~
 {: .output}
 
--- a/_episodes/01-numpy.md
+++ b/_episodes/01-numpy.md
@@ -60,7 +60,10 @@ print(weight_kg)
 and do arithmetic with it:
 
 ~~~
-print('weight in pounds:', 2.2 * weight_kg)
+weight_lb = 2.2 * weight_kg
+weight_lb = format(weight_lb, '.2f')
+
+print('weight in pounds:', weight_lb)
 ~~~
 {: .python}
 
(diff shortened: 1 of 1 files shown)


MikeTrizna:
I agree with @maxim-belkin that it would be simpler to change `weight` from 55 to 60 (or any other integer that doesn't produce this behavior). 

I've personally run into this behavior early in a Python workshop, and having to explain what was happening caused a major delay AND left a few participants concerned that they would be better off continuing to do calculations in Excel...

maxim-belkin:
Looks like @tbekolay suggested the same solution a while back here: https://github.com/swcarpentry/python-novice-inflammation/pull/339#issuecomment-339103601

@joshkyh can you update your pull request to change the weight from 55 to 60?

maxim-belkin:
Also, is this part of instructor checkout?

joshkyh:
@maxim-belkin done - pull request approach changed to simply changing the weight to 60 :) Not sure about your question on "is this part of instructor checkout". Is this question meant for me?

joshkyh:
Hi @maxim-belkin Apologies, I should have looked through the rest of the file in detail. I have changed the rest of the code that referenced the original weight and changed it accordingly.

I did some search with reference to "instructor checkout" procedure, I am not part of any instructor training programme and does not require this commit to be logged against my progress - just a member of the public. :)

Thanks,
Joshua

_episodes/01-numpy.md
```diff
@@ -53,19 +53,21 @@ print(weight_kg)
...
-print('weight in pounds:', 2.2 * weight_kg)
+weight_lb = 2.2 * weight_kg
+
+print('weight in pounds:', weight_lb)
 ~~~
 {: .python}
 
 ~~~
-weight in pounds: 121.0
+weight in pounds: 120.0
```
tbekolay: I believe this should be `132.0.`

joshkyh:
@tbekolay yes you're right, my bad. Thanks for pointing out.

PR #430, opened 2017-11-15
Add general tool intro to clarify how to use a Jupyter Notebook

chris-p-orth:
This PR includes some of the information from the [python-novice-gapminder](https://swcarpentry.github.io/python-novice-gapminder/01-run-quit/) course for a brief introduction to some concepts regarding Python and the use of Jupyter Notebooks before the first lesson begins.

### DIFF
--- a/_episodes/01-numpy.md
+++ b/_episodes/01-numpy.md
@@ -69,6 +69,8 @@ In this lesson we will learn how to manipulate the inflammation dataset with Pyt
         to tell a complete story of the analysis.
 
 
+## Storing values as variables.
+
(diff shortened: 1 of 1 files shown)


annefou:
It is unclear to me if we should add information about jupyter notebook in the lesson or in the setup only.
My feeling is that this python lesson can easily be independent on whether we use jupyter notebooks or not.

knarrff:
This PR is old, I know - but it is still open, thus:

We do mention that jupyter is one way to use Python and we do support those that use it, even in class. However, we do show the ipython prompt instead and most do follow this example. In a setup like ours and with the proposed addition, we would have to skip over a main course material element which would look odd. Because of that reason I agree that while I think it would be valuable addition, it should not go into the main course material.

Issue #426, opened 2017-11-09
Explanation of print function


dseuss:
While teaching the first part "Analyzing Patient Data" for the instructor checkout, we noted that the `print` function was never really explained. Since this is the first lesson in python, it would probably be good to add a short paragraph about the `print` function and especially that it accepts an arbitrary number of arguments in the very beginning.

rpep:
I've made a pull request which I hope addresses this. Does anyone have any comments?

mialondon:
Thanks for that, looks good to me!

PR #433, opened 2017-12-06
01-numpy.md: better introduction for the print function

rpep:
Add slightly more information about the print function to address issue #426

### DIFF
--- a/_episodes/01-numpy.md
+++ b/_episodes/01-numpy.md
@@ -45,7 +45,7 @@ We can create a new variable by assigning a value to it using `=`.
 When we are finished typing and press Shift+Enter,
 the notebook runs our command.
 
-Once a variable has a value, we can print it to the screen:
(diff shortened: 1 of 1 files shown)


dseuss:
I think the biggest problem with `print` for beginners is that it can take an arbitrary number of parameters.

_episodes/01-numpy.md
```diff
@@ -45,7 +45,7 @@ We can create a new variable by assigning a value to it using `=`.
 When we are finished typing and press Shift+Enter,
 the notebook runs our command.
 
-Once a variable has a value, we can print it to the screen:
+Once a variable has a value, we can print it to the screen. We do this using a function, 'print', which can take variables containing both text or numeric data:
```
gvwilson: Use back quotes around `print` instead of normal quotes so that it's set as a computer term. Other than that, :+1: to merge.

maxim-belkin:
I believe this PR can be closed as we updated the way `print` function is introduced and now we do mention that it can accept several arguments.

Issue #424, opened 2017-10-31
Key points in 06-func mentions stack frame/call stack, but these are not covered


shwina:
> Each time a function is called, a new stack frame is created on the call stack to hold its parameters and local variables.

> Python looks for variables in the current stack frame before looking for them at the top level.

harpolea:
I agree that these two points should be reworded - in the lesson, there is an example which talks about variables defined inside the function not being 'known' outside of the function. This sort of language would be more appropriate than 'stack frame'.

annefou:
I close this issue as it looks like it has been addressed. Please re-open if I am wrong.

PR #434, opened 2017-12-06
Issue #424 - Updated key points in 06-func.md to remove references to…

harpolea:
… call stack / stack frame


valentina-s:
@harpolea Thanks! Simplifying the keypoints is great! What do you think about adding an extra point  along the lines: 'If a variable is not defined within the function it is used, Python looks for a definition before the function call'?

harpolea:
Sounds good! I've added that to the keypoints

valentina-s:
Thanks!

PR #435, opened 2017-12-14
Fixed filename in "Moving Plots Around" challange

dn80:
Fixed the filename of the input from data/inflammation-01.csv to inflammation-01.csv to be consistent with the rest of the content in this session, assuming that the data/ is the working directory (this also avoids conflicts between unix- and windows-style path descriptions).


valentina-s:
Thank you for this fix!

PR #437, opened 2017-12-20
Improving the python-operations-across-axes image

### LINKED ISSUE #321: Prose error describing numpy rows and columns relative to axis number...
The following line shows rows with axis 0:

"If we ask for the average across axis 0 (rows in our 2D example), we get:"
    
and here columns are show with axis 1:

"If we average across axis 1 (columns in our 2D example), we get:"

The diagram above correctly shows axis 0 as columns and axis 1 as rows.

duncanwp:
This minor change to the image is to use the numpy.function approach, and better match the text description. Fixes #321


valentina-s:
Thank you for updating the figures! The old example had max on the left and mean on the right, which was consistent with the text. You have mean on both plots: could you make the left max? Or maybe even better, make both of them max and modify the text accordingly, and then add the example of mean (without a plot) to show that many functions have an option to add an axis.

Issue #436, opened 2017-12-20
Add setup instructions for Windows users


dafnevk:
It would be nice to have setup instructions for opening an interpreter or Jupyter Notebook with Anaconda, without using the Unix shell (especially useful for Windows users).
As I understand, the Unix shell is no prerequisite for this lesson and with Anaconda navigator you can open an interpreter or Python notebook from the GUI. You just need to navigate to the right directory afterwards.

PR #438, opened 2017-12-21
Added setup instructions for Windows with no Unix shell

tiagopereira:
To address https://github.com/swcarpentry/python-novice-inflammation/issues/436, added a couple of lines to the setup instructions so that Windows users with no Unix shell installed can instead run Python or Jupyter from `cmd`.


valentina-s:
Thanks for the suggestion! This can be helpful. I believe  the `cd` commands from above (navigating to the data directory) will not work in the Windows prompt. Could you add a version of the command compatible with windows? Do you think there is any benefit to use the anaconda prompt vs command prompt?

tiagopereira:
Hi @valentina-s thank you for noting that the `cd` commands were not working. I've made a slight change to enable access to the Desktop folder in Windows with `cd` from `cmd`. It should work, but please let me know if this works (I don't have a Windows machine to test).

valentina-s:
@tiagopereira Thank you for the addition. I am pretty sure you do not need the /d (that seems to be needed if you need to switch between drives, which most probably they will not).

tiagopereira:
You're right, `/d` is not necessary, but it will make it work in any circumstance.

Issue #445, opened 2018-01-16
Make numpy matrix indexing convention clear early on


nspies:
Minor fix to make it clear that `data[30,20]` accesses row 30, column 20 -- otherwise, this key piece of information is buried several paragraphs later in the "In the Corner" section.

maxim-belkin:
Closing this issue as corresponding PR has been merged.

PR #446, opened 2018-01-16
Make row,column indexing convention clear early on

nspies:
See #445


valentina-s:
Good point! Thank you for the contribution!

PR #448, opened 2018-01-19
Swap Newton with Curie

marwahaha:
We taught this lesson today. Someone commented on the "diverse professions" of the scientists... but it'd be nice to have diverse role models in our examples. Small but well-meaning change.

### DIFF
--- a/_episodes/03-lists.md
+++ b/_episodes/03-lists.md
@@ -73,7 +73,7 @@ but we cannot change individual characters in a string.
 For example:
 
 ~~~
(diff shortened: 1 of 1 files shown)


valentina-s:
Good point! Thanks!

slayoo:
BTW, if Curie refers to Maria Skłodowska-Curie, why not use both of her surnames (the Polish and the French one). See, e.g., the "The Marie Curie Actions will be renamed Marie Skłodowska-Curie Actions" entry in the European Commission's FAQ: https://ec.europa.eu/commission/presscorner/detail/en/MEMO_12_830

PR #457, opened 2018-02-08
Call winpty to access Python interpreter via Windows Git Bash

petebachant:
Could be confusing since there is no error output.


ldko:
I have seen this with trying to launch a Python interpreter in Git Bash by just typing `python`. Nothing seems to happen; interpreter does not launch, nor do you get back to the prompt. Using `winpty python` as suggested in this PR did work for me. It might also be worth adding that Windows users can launch a Python interpreter from Anaconda Prompt.

vahtras:
Thank you! A problem that has annoyed me for some time has now been solved.

maxim-belkin:
Thanks @petebachant! Could you please change `we need to call` to `you need to call` to make the entire section consistent?

setup.md
```diff
@@ -36,12 +36,20 @@ Then create a new notebook using the drop-down menu on the right to select 'Pyth
...
-To start the Python interpreter without the notebook, open a terminal or git bash and type the command:
+To start the Python interpreter without the notebook, open a terminal
+or command prompt and type the command:
 
 ~~~
 $ python
 ~~~
 {: .source}
 
+Note: If using Git Bash on Windows, we need to call Python via `winpty`:
```
maxim-belkin: Please change `we need to call` to `you have to call`.

setup.md
```diff
@@ -36,12 +36,20 @@ Then create a new notebook using the drop-down menu on the right to select 'Pyth
 
 ![](../fig/new-notebook.png)
 
-To start the Python interpreter without the notebook, open a terminal or git bash and type the command:
+To start the Python interpreter without the notebook, open a terminal
+or command prompt and type the command:
```
maxim-belkin: Actually, could you please also change `type the command` to `execute the command`?

maxim-belkin:
Merged! Thanks for your contribution, @petebachant ! 👍 
Please indicate if this is part of your instructor checkout process so that I can label it accordingly.

PR #467, opened 2018-02-09
Improve clarity and robustness in 'tidying up'

### LINKED ISSUE #466: Ep6 'tidying up' -- filenames are unsorted
Re:
 ```for f in filenames[:3]:
    print(f)
    analyze(f)
    detect_problems(f)
```
Earlier in the lesson we use 
```
filenames = glob.glob('inflammation*.csv')
filenames = filenames[0:3]
```
though this will be updated to the following if #462 gets merged
```
filenames = sorted(glob.glob('inflammation*.csv'))
filenames = filenames[0:3]
```
So the files are not currently sorted. For robustness and as a reminder, I think this code to list the filenames should be repeated here.

Equally, if we're using a subset of the filenames to only treat the first three files, then `filenames[:3]` is redundant.

gcapes:
Fix #466

### DIFF
--- a/_episodes/06-func.md
+++ b/_episodes/06-func.md
@@ -283,6 +283,8 @@ we can now read and reuse both ideas separately.
 We can reproduce the previous analysis with a much simpler `for` loop:
 
 ~~~
+filenames = sorted(glob.glob('inflammation*.csv'))
+
 for f in filenames[:3]:
     print(f)
     analyze(f)


annefou:
As discussed, sorting all the files is not necessary for the exercise. So the discussion is whether it adds clarity and robustness to the exercise.  I am new as a maintainer so I don't have enough experience in this area. I tend to stick to the rule: do not introduce new notion if unecessary.
I  label it as "help-wanted" so other can give their point of view!
Thanks,
Anne.

ldko:
Whether or not `filenames` gets `sorted` before just taking three of them here, I think it is a good idea to re-set the `filenames` variable in this code block. It is not set in this episode (episode 6), and has not been set in the main content of the lesson since episode 4. If learners have closed their Jupyter notebook or interpreter, they may not have the `filenames` variable set in memory anymore. For that reason and because it doesn't introduce a new concept, I think it is worth adding this line of code here.

gcapes:
@ldko 
Yep, that was the 'robustness' idea.
Given we previously sorted the filenames, it's not introducing any new concepts here to keep the sort., but reassigning `filenames` is the main point.

annefou:
Perfect! Thank you for your help and contribution.

Issue #464, opened 2018-02-09
Ep 05 'Sorting a list into Buckets' - solution should use clues given


gcapes:
> Add code to the template below to do this. Note that the string method startswith returns True if and only if the string it is called on starts with the string passed as an argument.

Yet the solution uses the `if x in y` notation instead, so there seems no reason to have mentioned the `startswith` method. Given the `x in y` notation hasn't been introduced, I propose changing the solution to use the `startswtih` method.

cleebp:
I think this change is needed - I just created a [pull request](https://github.com/swcarpentry/python-novice-inflammation/pull/468) to change the solution code to use the startswith command over the 'if x in y' style previously being used.

gcapes:
See also #465, of which your PR is a duplicate ;)

gcapes:
No worries! It's just waiting for a maintainer to look at it.

PR #468, opened 2018-02-13
modified solution to 'Sorting a list into Buckets'

cleebp:
In reference to [issue 464](https://github.com/swcarpentry/python-novice-inflammation/issues/464) - modified the solution to 'Sorting a list into Buckets' to use clues as given (startswith method) rather than 'if x in y'.


gcapes:
Duplicate of #465

PR #476, opened 2018-02-21
Update 01-numpy.md – update heat map color explanation

cleary3:
Regarding the heat map, the colors vary from blues to yellows. The text indicated that red shows high values, but that's incorrect. It's yellow that indicates high values.

---


ldko:
This is something that needed to be changed. It looks like a similar fix was also presented as part of [another PR](https://github.com/swcarpentry/python-novice-inflammation/pull/482/files#diff-d5ac5e019d91f8991ce21e58ffdd24caR698) that has now been merged.

annefou:
The fix was already included so I close this PR.

PR #480, opened 2018-02-21
Update 08-defensive.md — en dash for numeric range

cleary3:
Us an en dash instead of a figure dash in order to represent a numeric range.


annefou:
Thanks for your contribution.

PR #481, opened 2018-02-21
edited explanation of numpy.diff

pandrews:
Array is not a built-in function and hasn't been imported specifically. Changed to call by numpy.array to prevent NameError


annefou:
You are right. Thank you for your contribution.
Anne.

PR #486, opened 2018-02-27
Lesson 7 - Illustrate the structure of a Python traceback schematically

dwillcox:
Lesson 7, "Errors and Exceptions" discusses examples of real tracebacks and what different kinds of errors a user can encounter. However, the structure of a Python traceback could be a bit intimidating for new users.

This pull request adds an SVG schematic showing the main pieces of a traceback and the connections between them. I also added supporting text to briefly describe the overall structure of a traceback and give tips on what to look for when debugging a traceback.

### DIFF
--- a/_episodes/07-errors.md
+++ b/_episodes/07-errors.md
@@ -29,7 +29,26 @@ they become much easier to fix.
 
 Errors in Python have a very specific form,
 called a [traceback]({{ page.root }}/reference/#traceback).
-Let's examine one:
+A traceback can contain a lot of information, so the following
+schematic highlights the main sections of a traceback.
+
+![Schematic of a Python Traceback](../fig/traceback_schematic.svg)
+
+In this diagram the green box in the top left indicates the type of
+error that occurred. The blue shaded regions below are optionally
+printed depending on how many function calls there are between the
+Python program or interpreter and the location where the error
+occurred. For each function described, the line number and
+corresponding line of code affected by the error is shown.
+
+The most recently executed section of code containing the error is
+shown highlighted in green at the bottom together with a final line
+indicating the nature of the error in this section. This is often the
+best place to start when debugging a traceback because many times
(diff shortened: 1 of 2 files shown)


rgaiacs:
Thanks for the issue. I have mixed feeling if the SVG is the best way to illustrate the structure. Maybe a table where the first column is the Python traceback and the second column is the structure information? Using tables should make easy to have our lesson accessible.

annefou:
Yes I fully agree. A table would be easier to maintain and offer more possibility to easily contribute to the lesson. @dwillcox what is your opinion? Do you think you could build a table?

Thanks for your contribution.

maxim-belkin:
1. SVG rocks.
2. I know the basic syntax of SVG but not enough to make complex changes (if they are ever required) manually. We can use some software to work with SVGs (like @dwillcox did here) but we have to discuss this first.

Meanwhile, @dwillcox could you please clean up SVG to remove unnecessary code and make it as simple as possible so that we could estimate the difficulty of maintaining it in the long run?

rgaiacs:
> so that we could estimate the level of effort that will be required to maintain it in the long run?

From my experience, maintain SVG files is hard both in terms of browser support (see one issue that we had a few months ago) and edit software (some editors add additional information).

_episodes/07-errors.md
```diff
@@ -29,7 +29,26 @@ they become much easier to fix.
 
 Errors in Python have a very specific form,
 called a [traceback]({{ page.root }}/reference/#traceback).
-Let's examine one:
+A traceback can contain a lot of information, so the following
+schematic highlights the main sections of a traceback.
```
gvwilson: I like the idea of a diagram to explain tracebacks, but I have to admit that I found this diagram more confusing than enlightening: I didn't understand what the red arrows were pointing from or to.

Issue #459, opened 2018-02-09
Explain or remove "center a dataset" in 06-func


mikej888:
Provide a 1-2 line description of what this concept is in http://swcarpentry.github.io/python-novice-inflammation/06-func/m, and why it's done, in terms that a non-maths/stats student (or instructor!) can comprehend. Alternatively, preferably, rewrite that page to introduce the concepts of functions without such stats/maths concepts also.

mikej888:
Thanks :-)

PR #487, opened 2018-02-28
06-func.md: Replace "center" function name

neurosutras:
This issue https://github.com/swcarpentry/python-novice-inflammation/issues/459 recommended either replacing the function name "center", or providing a better explanation for what it did. Here I replace "center" with a potentially more intuitive "offset_mean". Since the function name and its description are repeated many times throughout the lesson, the relationship between its name and its purpose have many opportunities to be made clear and that understanding reinforced.

### DIFF
--- a/_episodes/06-func.md
+++ b/_episodes/06-func.md
@@ -302,10 +302,11 @@ we can do so in a single line.
 Once we start putting things in functions so that we can re-use them,
 we need to start testing that those functions are working correctly.
 To see how to do this,
-let's write a function to center a dataset around a particular value:
+let's write a function to offset a dataset so that it's mean value 
+shifts to a user-defined value:
 
 ~~~
-def center(data, desired):
+def offset_mean(data, desired):
     return (data - numpy.mean(data)) + desired
 ~~~
 {: .language-python}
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Hi, Aaron! Thanks for the PR!

I like the idea and agree that `offset_mean` is better than `centered`. Personally, I would prefer `offset_mean_value` and also change `desired` to `new_mean_value`.

@MikeTrizna, what are your thoughts on this?

Aaron, is this, incidentally, part of instructor checkout?

MikeTrizna:
@maxim-belkin, I agree to `new_mean_value` (or `target_mean_value`, etc.) is a good replacement for desired, but I think that `offset_mean` is fine for the function name. `offset_mean_value` seems like it would be a good variable name to store the results of the `offset_mean` function (although I don't think we do that in this lesson at all).

_episodes/06-func.md
```diff
@@ -315,11 +316,11 @@ but since we don't know what the values ought to be,
 it will be hard to tell if the result was correct.
 Instead,
 let's use NumPy to create a matrix of 0's
-and then center that around 3:
+and then offset it to have a mean value of 3:
```
maxim-belkin: I'd suggest changing `offset` to `offset its values`

gvwilson:
This is a good change - there are a few minor comments, but I think it can be merged as-is.

maxim-belkin:
@neurosutras just checking in to see if you'll have time to make the proposed changes.
It looks like you've enabled "Allow changes from maintainers", so we can also push to your branch if you don't have time at the moment. Just let us know. 

Thank you!

neurosutras:
Hi @maxim-belkin , you are more than welcome to push to this branch and oversee it's merging to master. Thank you.

maxim-belkin:
OK, thanks @neurosutras! Will do!

PR #492, opened 2018-03-28
Merging updates

maxim-belkin:
Merging updates from [swcarpentry/styles](https://github.com/swcarpentry/styles)


maxim-belkin:
I will redo this PR after merging in #493 and removing merge commit

PR #503, opened 2018-04-20
setup.md: add section headers, rewrite, and simplify

maxim-belkin:
* Add section headers

This should help workshop instructors if attendees have any troubles
during preparation. They can easily refer to a specific section.

* Reorganize

The first option to launch Python interpreter is now `python` command.
Remove duplicate references to Unix shell.

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -5,36 +5,56 @@ permalink: /setup/
 root: ..
 ---
 
-In preparation for this lesson, you will need to download two zipped files and place them in the specified folder:
+## Get the data
 
-1. Make a new folder in your Desktop called `python-novice-inflammation`.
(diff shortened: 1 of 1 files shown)


annefou:
I would not have Optional for jupyter notebook. It does not really correspond to what is taught. I mean jupyter notebook are vastly used for teaching python.

maxim-belkin:
> I would not have Optional for jupyter notebook. It does not really correspond to what is taught. I mean jupyter notebook are vastly used for teaching python.

OK. Thank you!

maxim-belkin:
I'd like to get a few more comments on this PR before merging it... so let's keep it open for some time.

maxim-belkin:
I just realized that the first two titles describe some actions (**Get** the data, **Launch** the interpreter). It would be better if the third title follows the same rules.

Issue #504, opened 2018-04-20
Add note to copy-paste TabError example (ep. 7)


yeemey:
In [episode 7](http://swcarpentry.github.io/python-novice-inflammation/07-errors/) **Tabs and Spaces** callout, the example code has to be copied and pasted directly to reproduce the ```TabError``` (see #221 ).

PR #505, opened 2018-04-20
Added instruction to copy-paste TabError example

yeemey:
#504 Edited to specifically note copying and pasting the example.

### DIFF
--- a/_episodes/07-errors.md
+++ b/_episodes/07-errors.md
@@ -186,7 +186,7 @@ it *always* means that there is a problem with how your code is indented.
 > because they are both [whitespace]({{ page.root }}/reference/#whitespace).
 > The Jupyter notebook actually gives us a bit of a hint,
 > but not all Python editors will do that.
-> In the following example,
+> Copy and paste the following example, where
 > the first two lines are using a tab for indentation,
 > while the third line uses eight spaces:
 >


_episodes/07-errors.md
```diff
@@ -186,7 +186,7 @@ it *always* means that there is a problem with how your code is indented.
 > because they are both [whitespace]({{ page.root }}/reference/#whitespace).
 > The Jupyter notebook actually gives us a bit of a hint,
 > but not all Python editors will do that.
-> In the following example,
+> Copy and paste the following example, where
```
maxim-belkin: I'd prefer to leave the beginning of the sentence as it was before (`In the following example,`) because without specific direction as to *where* to paste it in, instructions seem unclear. My suggestion would be to add something like `(be sure to copy and paste the example into your interpreter instead of typing it in to preserve Tabs)`

maxim-belkin:
Thanks for the contribution, @yeemey. I added my comments on this.

_episodes/07-errors.md
yeemey: Makes sense, thanks!

maxim-belkin:
Thank you @yeemey!
Thank you @annefou!

PR #510, opened 2018-04-30
Use bullet points in debuggin lesson

timkittel:
Hi all,

I created this Pull Request in order to suggest to use bullet points in this particular session. While it might be a matter of style whether to use paragraph style text or bullet points for the rest of the lessons, I see bullet points as the first choice here. I demonstrated this here for the "Make it fail" and "Change" sections and would be happy to do the rest of whole lesson in this style (while not actually changing the contents).

Reasons for bullet points:

This lesson is different from the others, as it's not a life-coding lesson but basically just the speaker talking for a while. When I gave that lesson, I presented the bullet points, similar to the demo I gave here, on the beamer so the participants could follow easily and have a summary while I was speaking. Further, we made all lesson notebooks afterwards available to the participants, hence it would be good to have something that people can grasp quickly as scientist are often not so eager to spend much time on things which are not exactly their research (:

Best,
Tim

PS: I know this is a personal opinion and I am happy to discuss the pros and cons (:
PS2: When we found an idea that everybody is happy with, I'd adjust the whole lesson accordingly.

### DIFF
--- a/_episodes/09-debugging.md
+++ b/_episodes/09-debugging.md
@@ -137,7 +137,7 @@ and so that they don't waste time repeating the same experiments.
 ## Be Humble
 
 And speaking of help:
-if we can't find a bug in 10 minutes,
+if we can't find a bug in 30 minutes,
 we should *be humble* and ask for help.
 Just explaining the problem aloud is often useful,
 since hearing what we're thinking helps us spot inconsistencies and hidden assumptions.
--- a/_episodes/09-debugging.md
+++ b/_episodes/09-debugging.md
@@ -17,7 +17,7 @@ keypoints:
 ---
 
 Once testing has uncovered problems,
-the next step is to fix them.
+the next step is to fix them. This is called *debugging*.
 Many novices do this by making more-or-less random changes to their code
 until it seems to produce the right answer,
 but that's very inefficient
@@ -84,7 +84,7 @@ scientists tend to do the following:
     If "new" patients start appearing out of nowhere as we move through our pipeline,
     it's probably a sign that something is wrong.
 
-5.  *Visualize.*
+5.  *Visualize your data.*
     Data analysts frequently use simple visualizations to check both
     the science they're doing
     and the correctness of their code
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Hi @timkittel and thank you for the PR.

I agree that we should better structure this part of the episode because, in the end, it will help both instructors and learners. I don't agree, however, with dropping all the "prose" here.

A few other points:

1. It should be, in fact, just `Visualize` because we're talking about visualizing the results produced by our program ("analysis"), and not about visualizing data (that is, unprocessed input).
2. `Make it fail` and `Change` are too ambiguous (even though I see how you're trying to use them).

I would argue that the current titles `Make It Fail Every Time` and `Make It Fail Fast` are ambiguous too. What we really want to say here is that we want to concentrate on a failure and not so much on the parts that are "fine". So, we don't want to "make it fail every time"... we want to know how to consistently reproduce failure. And we don't want to "make it fail fast", we want to terminate execution of the program and report the error immediately upon failure.

For this PR, I'd suggest keeping the first change only and updating it to something like:

> If tests report errors, then there are mistakes in our program. These mistakes are called "bugs". The natural next step is to fix these mistakes or, in other words, "debug" our program.

or 

> Every test failure corresponds to a mistake in our program. These mistakes are called "bugs". The natural next step is to fix these mistakes or, in other words, "debug" our program.

timkittel:
Hi @maxim-belkin and thanks for you feedback.

if I summarize it, you would basically like me to remove the last commit and then add the explanation of a bug. Do I understand you correctly? (:

maxim-belkin:
Um, I'm sorry for the confusion... when I said `the first change` I meant the [this one](https://github.com/swcarpentry/python-novice-inflammation/pull/510/files#diff-7ecf85c2ce71ed14f4b56be7d6995c65R20) 😬

timkittel:
Assuming there is not much going to happen here anymore and it seem my suggestion with the bullet points would be a bigger discussion anyway, I am closing this PR. That way I have it out of my todo list (:

Issue #514, opened 2018-05-01
Python lessons don't mention the ':' character and its importance


gkiar:
I've noticed in both the loops and conditionals lessons that no explanation of the syntax for these operators is explicit about including the `:` character at the end of the declaration - this is an important distinction from other languages, and perhaps not obvious.

maxim-belkin:
I agree, mandatory nature of colon warrants at least a brief word about it. Would you like to submit a PR?

gkiar:
Yep! See #516

PR #516, opened 2018-05-02
added description of the ':'

gkiar:
Addresses #514

### DIFF
--- a/_episodes/02-loop.md
+++ b/_episodes/02-loop.md
@@ -141,14 +141,6 @@ for variable in collection:
 ~~~
 {: .language-python}
 
-> ## What is the ":" for?
->
-> One thing you may have noticed is that our `for` statement ends with a `:`. You actually will see the `:` regularly in Python, and
-> it actually can serve two functions. The first is indexing, which you became familiar with in the previous episode. The second, as is
-> the case in `for` loops, is the opening of a "block". A block is one or more lines of code that are indented from their parent and
-> will be run together. To leave a block, you just need to leave the indent!
-{: .callout}
-
 Using the oxygen example above, the loop might look like this:
 
 ![loop_image](../fig/loops_image.png)
@@ -157,8 +149,10 @@ where each character (`char`) in the variable `word` is looped through and print
 The numbers in the diagram denote which loop cycle the character was printed in (1 being the first loop, and 6 being the final loop).
 
 We can call the [loop variable]({{ page.root }}/reference/#loop-variable) anything we like,
-but to open the block we can't forget to put a colon at the end of the line starting the loop,
-then indent the code we want to run inside the loop. Unlike many other languages, there is no
+but to open the block we can't forget to put a colon (`:`) at the end of the line starting the loop,
+then indent the code we want to run inside the loop. The colon identifies the beginning of a list of instructions, and the
+instruction before it tells us how to handle the list, in this case meaning it is a set of instructions we will loop over.
+Unlike many other languages, there is no
 command to signify the end of the loop body (e.g. `end for`); the code block (i.e. indented section) 
 after the `for` statement belongs to the loop -- you can also nest these!
 
--- a/_episodes/02-loop.md
+++ b/_episodes/02-loop.md
@@ -144,7 +144,7 @@ for variable in collection:
 > ## What is the ":" for?
 >
 > One thing you may have noticed is that our `for` statement ends with a `:`. You actually will see the `:` regularly in Python, and
-> it actually can serve two functions. The first is indexing, which you became familiar with in the last lesson. The second, as is
+> it actually can serve two functions. The first is indexing, which you became familiar with in the previous episode. The second, as is
 > the case in `for` loops, is the opening of a "block". A block is one or more lines of code that are indented from their parent and
 > will be run together. To leave a block, you just need to leave the indent!
 {: .callout}
@@ -157,10 +157,10 @@ where each character (`char`) in the variable `word` is looped through and print
 The numbers in the diagram denote which loop cycle the character was printed in (1 being the first loop, and 6 being the final loop).
 
 We can call the [loop variable]({{ page.root }}/reference/#loop-variable) anything we like,
-but there must be a colon at the end of the line starting the loop,
-and we must indent anything we want to run inside the loop. Unlike many other languages, there is no
-command to signify the end of the loop body (e.g. `end for`); what is indented after the `for` statement belongs to the loop.
-
+but to open the block we can't forget to put a colon at the end of the line starting the loop,
+then indent the code we want to run inside the loop. Unlike many other languages, there is no
+command to signify the end of the loop body (e.g. `end for`); the code block (i.e. indented section) 
+after the `for` statement belongs to the loop -- you can also nest these!
 
 > ## What's in a name?
 >
--- a/_episodes/02-loop.md
+++ b/_episodes/02-loop.md
@@ -141,6 +141,14 @@ for variable in collection:
 ~~~
 {: .language-python}
 
+> ## What is the ":" for?
+>
+> One thing you may have noticed is that our `for` statement ends with a `:`. You actually will see the `:` regularly in Python, and
+> it actually can serve two functions. The first is indexing, which you became familiar with in the last lesson. The second, as is
+> the case in `for` loops, is the opening of a "block". A block is one or more lines of code that are indented from their parent and
+> will be run together. To leave a block, you just need to leave the indent!
+{: .callout}
+
 Using the oxygen example above, the loop might look like this:
 
 ![loop_image](../fig/loops_image.png)


maxim-belkin:
Let's leave line 18 for now.

I see that you're using `gh-pages` branch for this PR and you shouldn't. to get out of this situation do:

```
git fetch origin
git branch mypr
git reset --hard origin gh-pages
git checkout mypr
git push https://github.com/gkiar/python-novice-inflammation
git checkout gh-pages
```
and submit a new PR using that `mypr` branch I then will be able to commit to that branch directly to make corrections.

gkiar:
Why, @maxim-belkin? If the destination is set to be the appropriate branch (`gh-pages`, as it is), then it doesn't matter the name of the source branch. I get using feature-specific branch names in general, and if it is a policy I'm happy to adopt moving forward, but this was such a tiny modification that it hardly seems necessary at this stage... If the concern is the descriptiveness of the commit-log and what the merge contributed, it can be merged with a squash-commit and given a descriptive name there.

gkiar:
I disagree with your statement that your proposal is the only correct solution... To answer your questions:
> How can you submit another pull request?
1. Exactly how I submitted this one. :)
> How can you update your fork?
2. By merging in the base prior to opening the PR, which can be done any number of ways, perhaps most simply by adding a remote and fetching it locally.
>How can you rebase your PR on top of the latest changes in the main repository?
3. Rebasing is not necessary if you manage merge commits locally, which I mentioned in the answer to 2. Rebasing can be detrimental and cause intermediate commits which may have been made with functioning code to not be functioning after a rebase if conflicts need to be resolved at the end of the tree, compromising the integrity of the history and contributions... 1 merge commit (as is the case here) is not a mess, but being unable to go through history on a branch to leverage intermediate commits in their original form is potentially harmful. ([this medium post, and in particular the 'top highlighted' quote emphasizes this nicely](https://medium.com/@fredrikmorken/why-you-should-stop-using-git-rebase-5552bee4fed1))

Regardless, it appears we could argue about it for longer than it took to make the changes (and we're not the only ones, as it is indeed a style and priority debate, as indicated in the post I shared above). If that is a style which this repository requires, I suggest updating [Line 188 of the contributing guide](https://github.com/swcarpentry/python-novice-inflammation/blob/gh-pages/CONTRIBUTING.md), from "...Create a new branch in your desktop copy of this repository for each significant change." to "...Create a new branch in your desktop copy of this repository (no matter how small the change)." Otherwise, I see no reason not to merge this PR.

maxim-belkin:
> Exactly how I submitted this one. :)

Please submit any other PR to prove me wrong.

maxim-belkin:
> Otherwise, I see no reason not to merge this PR.

I will add comments to your text so that it is clear why this PR is not ready yet.

_episodes/02-loop.md
```diff
@@ -141,6 +141,14 @@ for variable in collection:
 ~~~
 {: .language-python}
 
+> ## What is the ":" for?
+>
+> One thing you may have noticed is that our `for` statement ends with a `:`. You actually will see the `:` regularly in Python, and
```
maxim-belkin: 1.
> You might have noticed that ...

seems easier to say than

>  One thing you may have noticed is that

2. 
> our `for` statement

it is not "our". This are just "`for` statements"

_episodes/02-loop.md
```diff
@@ -141,6 +141,14 @@ for variable in collection:
...
+> it actually can serve two functions. The first is indexing, which you became familiar with in the previous episode. The second, as is
```
maxim-belkin: 1. word "actually" appears twice in this sentence.
2. We should not draw student's attention from the current topic. Reminding is good but only once you finished explaining current topic. We should explain the purpose of colons in this particular case.

_episodes/02-loop.md
```diff
@@ -141,6 +141,14 @@ for variable in collection:
...
+> the case in `for` loops, is the opening of a "block". A block is one or more lines of code that are indented from their parent and
```
maxim-belkin: `"block"` is jargon and, in this case, ambiguous. I know what you mean but those who are just starting with Python - don't.

`of a "block"` -> `of the indented block of code that follows`.

maxim-belkin: `parent`  jargon as well.

There is simply too much cognitive load in this paragraph.

_episodes/02-loop.md
```diff
@@ -141,6 +141,14 @@ for variable in collection:
...
+> will be run together. To leave a block, you just need to leave the indent!
```
maxim-belkin: > will be run together

meaning is not clear

> To leave a block

jargon

> leave the indent

jargon

_episodes/02-loop.md
```diff
@@ -149,10 +157,10 @@ where each character (`char`) in the variable `word` is looped through and print
 The numbers in the diagram denote which loop cycle the character was printed in (1 being the first loop, and 6 being the final loop).
 
 We can call the [loop variable]({{ page.root }}/reference/#loop-variable) anything we like,
-but there must be a colon at the end of the line starting the loop,
-and we must indent anything we want to run inside the loop. Unlike many other languages, there is no
-command to signify the end of the loop body (e.g. `end for`); what is indented after the `for` statement belongs to the loop.
-
+but to open the block we can't forget to put a colon at the end of the line starting the loop,
```
maxim-belkin: same as above: `block` is jargon. 
This section's initial content was fine as is. I agree that some context on the purpose colon would be helpful but not by increasing the cognitive load of the section.

_episodes/02-loop.md
```diff
@@ -149,10 +157,10 @@ where each character (`char`) in the variable `word` is looped through and print
...
+then indent the code we want to run inside the loop. Unlike many other languages, there is no
+command to signify the end of the loop body (e.g. `end for`); the code block (i.e. indented section) 
+after the `for` statement belongs to the loop -- you can also nest these!
```
maxim-belkin: same as before: why mention loop nesting if this is not covered in the episode?

gkiar:
I'd be eager to hear another maintainer's thoughts on all of the above.

annefou:
I tend to agree with @maxim-belkin in the sense I would not complexify this part unnecessarily. We do not aim our target audience to become professional software developer in python but mostly be able to write clean and reproducible codes for their research.

tbekolay:
Another thought I had after posting my review... I think part of the reason students aren't confused by the colon and indenting is because there is plenty of precedent for this in the real world -- hence why, I assume, Python chose this convention. One way of explaining it without using jargon would be to make an analogy to one of those real world examples... something like

```
Consider a recipe. It's common to write out the steps of the recipe like this:
  1. instruction one
  2. instruction two
Note that we used a colon before starting the list of steps, and we denote what is part of the list by indenting the text, or using a number of bullet.

Similarly, we tell Python that we want to repeat a set of instructions by putting a colon after the for statement, and indenting those instructions. 
```

This is too long for the lesson as is, I believe, but hopefully you get the idea.

gkiar:
@maxim-belkin @annefou @tbekolay thanks for the feedback. I particularly like @tbekolay's analogy relating this to lists, and have simplified my contribution. Regarding the discussion earlier in this thread, if it is indeed expected practice that all pull-requests come from separate branches, I will open a separate PR which updates the contributing guide to clarify this point.

maxim-belkin:
@gkiar, regarding branches: the wording (perhaps, not super clear) is already there -- see [CONTRIBUTING.md#using-github](https://github.com/swcarpentry/python-novice-inflammation/blob/gh-pages/CONTRIBUTING.md#using-github)

And, for our own reference: https://markmail.org/message/ve7mwqxhci4pm6lw

_episodes/02-loop.md
```diff
@@ -149,10 +149,12 @@ where each character (`char`) in the variable `word` is looped through and print
...
-but there must be a colon at the end of the line starting the loop,
-and we must indent anything we want to run inside the loop. Unlike many other languages, there is no
-command to signify the end of the loop body (e.g. `end for`); what is indented after the `for` statement belongs to the loop.
-
+but to open the block we can't forget to put a colon (`:`) at the end of the line starting the loop,
+then indent the code we want to run inside the loop. The colon identifies the beginning of a list of instructions, and the
+instruction before it tells us how to handle the list, in this case meaning it is a set of instructions we will loop over.
+Unlike many other languages, there is no
+command to signify the end of the loop body (e.g. `end for`); the code block (i.e. indented section) 
+after the `for` statement belongs to the loop -- you can also nest these!
```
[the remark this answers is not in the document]
gkiar: The reason I mention a code block, is that this is no-doubt a phrase they will hear in the real world of programming, and it is not a complicated on to grasp (as you see I defined it simply as the "indented section"). It's also not actually jargon - a "block of text" is a universal concept in English, that means the exact same thing in programming.

Also, the mention of nesting is an important concept so that they grasp, that they can string together complex logic in this way. Introducing these phrases do not considerably increase cognitive load, but plant seeds that will make taking to other lessons and self-guided instruction later more flexible and robust.

>"instructions before" don't tell us "how to handle the list" but tell Python how many times to repeat the loop and update loop variable, ...)

Colons are used for `if` statements as well, and the instructions before (i.e. `if ___` or `for ____`) tell us how to handle the instructions, whether to loop or conditionally run them... I think changing the phrase "list of instructions" to "set of instructions" would perhaps make it more clear, since your lessons don't seem to introduce Python `sets`, reducing the risk of overloading a word.

gkiar:
@maxim-belkin I referenced this contributing guide to you already - and according to it, my pull request is fine. If you require a new branch *always*, then the guide should be updated to remove the words `"for each significant change"`.

maxim-belkin:
I guess we read the following sentence differently:

> Create a new branch in your desktop copy of this repository for each significant change.

This does not mean you should not create branches for "small" changes (it would've been weird if the word `significant` nullified the entire preceding paragraph.). What this sentence means is that if you want to fix some issue A **and** issue B, you should:
- create a branch for issue A and a separate branch for B when both are "significant"
- create only one branch when one of the issues is not significant (typo)

maxim-belkin:
Closing for now due to no activity. Thank you, @gkiar, for your work anyway.

Issue #506, opened 2018-04-24
Simple exercise to differentiate print vs return


maxim-belkin:
We should probably add the following simple exercise somewhere:
```
print(print(1))
```

tlvdberg:
Hi Maxim!
I made an exercise for learning the difference between return and print. I will submit a pull request. You can see if you like it, and merge it or not. 
Let me know if you would rather like any changes. Let me know if something's not as it should be, I'm new to this. 
I will make it part of my check-out exercise(s). I already made some other small contributions, but am not sure that'll be enough.
Best regards,
Tineke

maxim-belkin:
Hi Tineke! Thank you for the note! As for the instructor checkout - size (of the PR) does not matter. Even a one-word change counts.

maxim-belkin:
Closed in #524

Issue #489, opened 2018-03-20
Add exercise to stress how functions are different from math functions


iglpdc:
Mark Guzdial had [an interesting post a few days ago](https://computinged.wordpress.com/2018/02/26/confusion-over-the-forms-of-programming-problems-mathematics-physics-versus-cs/) on how learners with a math background may assume mistakenly that a function in programming is the same as a function in math. The exercise just asks the learner what's the output of:

```python
def computeSalary():
  rate = 10.00
  hours = 40
  salary = hours * rate
  hours = 30
  print salary
  increase = 1.00
  salary = hours * (rate + increase)
  print salary
```
I tried it and got it wrong and realized that, in fact, I've always thought in functions as mathematical functions.

As we have many learners from science and math camps, I think it would be a good addition to the episode on functions.

annefou:
a will be None because the function one does not return a value. Same for print. 
If there is no return statement in the function code, the function ends, when the control flow reaches the end of the function body and the value "None" will be returned.

maxim-belkin:
:) I [know](https://stackoverflow.com/a/43029507/7410631)

iglpdc:
I think the point of Mark's exercise is not so much distinguishing `return` vs `print`, but the fact that people with a mathematical background tend to assume that functions in programming are equivalent to mathematical functions. A mathematical function will map an input to an output or some input argument and return a value (`None` counts as a value here).

In the exercise, people may follow this trap and say that there is *only one* number being printed. I think not using `return` and instead using `print` twice stresses the issue. A function returns thru one `return` line every time is executed, so it looks like a mathematical function, but in cases like the above we see that is actually a set of statements. 

I think @maxim-belkin 's exercise is too "easy" :)

```
def one():
    print(1)
    print(5)

a = one() # what is the value of `a`?
```
Also the value of `a` is not the point, in the sense that `a` can have only one value. The point is that two numbers are printed (I guess most people would say that here and that's why I'm saying it's easy). 

But in Mark's example, you start doing the math and think that the important thing is that the value of `salary` changes, you only keep track of the updated value and because it really looks like a mathematical function, you may miss the first `print(salary)` if you have the (wrong) preconception that functions are mathematical functions.

@annefou's response might reinforce the wrong preconception: 

> a will be None because the function one does not return a value. Same for print.
If there is no return statement in the function code, the function ends, when the control flow reaches the end of the function body and the value "None" will be returned.

[...]

PR #524, opened 2018-05-04
Return vs print exercise

tlvdberg:
I added an exercise making clear the difference between return and print. As asked for in issue:
https://github.com/swcarpentry/python-novice-inflammation/issues/506

Lines are approximately 807-850.

This is part of an check-out exercise (if accepted).

### DIFF
--- a/_episodes/06-func.md
+++ b/_episodes/06-func.md
@@ -804,6 +804,41 @@ readable code!
 > {: .solution}
 {: .challenge}
 
+> ## Return versus print
+>
+> Note that `return` and `print` are not interchangeable.
+> `print` is a Python function that *prints* data to the screen.
+> It enables us, *users*, see the data.
+> `return` statement, on the other hand, makes data visible to the program.
+> Let's have a look at the following function:
+>
+> ~~~
+> def add(a, b):
+>     print(a + b)
+> ~~~
+> {: .language-python}
+>
+> **Question**: What will we see if we execute the following commands?
+> ~~~
+> A = add(7, 3)
+> print(A)
+> ~~~
+> {: .language-python}
+>
+> > ## Solution
+> > Python will first execute the function `add` with `a = 7` and `b = 3`,
+> > and, therefore, print `10`. However, because function `add` does not have a
+> > line that starts with `return` (no `return` "statement"), it will, by default, return
+> > nothing which, in Python world, is called `None`. Therefore, `A` will be assigned to `None`
+> > and the last line (`print(A)`) will print `None`. As a result, we will see:
+> > ~~~
+> > 10
+> > None
+> > ~~~
+> > {: .output}
+> {: .solution}
+{: .challenge}
+
 > ## Selecting Characters From Strings
 >
 > If the variable `s` refers to a string,


maxim-belkin:
Thanks, @tlvdberg! A couple of thoughts:

1. We can discuss `print vs. return` after introducing functions, which we do in episode 6. So, I think this should be moved to episode 6.
2. It might make sense to have a simpler example, something like `print(1)` vs `return 1`  or simply discuss the outputs of `print(print(1))`  or  `print(print(1) + 2)`

I should also note that we should probably revisit the use of the word `"return"` in episodes prior to 6.

jttkim:
I very much like the idea of this exercise. Realising that that which we see and that which the computer "sees", or rather processes are two very different things is a key step in learning to program, and to work with computers in general. It may further be useful to point out parallels to shell scripting, in particular the joint good practice of dealing with errors by writing a meaningful error via `echo ... >& 2` to the standard error and exiting with a non-zero code.

I think @maxim-belkin is right in suggesting to move the exercise to a stage where functions are introduced.

I'd prefer minimising any discussion and dwelling on the `None` value implicitly generated by Python when nothing is returned. The conflation of returning `None` with returning nothing is quite an infelicity of Python's design. Guiding learners to using `type` for exploring returned values is a generally useful practice, but I'm not sure whether they'd get much use or benefit from the `print(print(1))` construct.

maxim-belkin:
I'd avoid drawing parallelism with shell scripting at this point for a few reasons:
1. in shells values are passed by "printing" them and capturing output, rather than by returning values
2. in shells `if success then something fi` is equivalent to `if 0 then something fi` whereas in Python it is equivalent to `if not 0: something`

I had a paragraph on `None` in my original response to this PR but then scrapped it to give myself time to think about it. We use it in lesson 06-func but we don't "explain" it. And I think that any example demonstrating the difference between print and return will inevitably mention `None`

maxim-belkin:
@tlvdberg have a look at swcarpentry/python-novice-inflammation#489

tlvdberg:
Hi All,
I actually also think it is better to discuss this right after introducing function, but as you asked for an exercise, I put it in the exercises. Now, didn't I change this to episode 6 (functions)??? Under 'files changed' it says '_episodes/06-func.md'... What are you referring to?

The problem I would rise is that the lesson is already quite long... I actually think that we could take out the 'Integer Division' section, that is in episode 6 right after introducing functions. Because it's not a problem in Python 3 and discussing things of Python 2 might only confuse learners. It is information they don't need for working successfully with Python 3. Maybe we could replace this by some comments about return vs. print. I would make the lesson much longer than it already is...

About #489: interesting. Maybe good indeed comparing functions with and without return statements.

maxim-belkin:
> Now, didn't I change this to episode 6 (functions)??? Under 'files changed' it says '_episodes/06-func.md'... What are you referring to?

My apologies. Not sure what went wrong here.

> I actually think that we could take out the 'Integer Division' section, that is in episode 6 right after introducing functions.

Let's do that. In fact, we can move it to the "additional exercises" page (`_extras/extra_exercises.md`).

_episodes/06-func.md
```diff
@@ -804,6 +804,51 @@ readable code!
 > {: .solution}
 {: .challenge}
 
+> ## Return versus print
+>
+> Note that `return` and `print` are not interchangeable.
+> `print` is a build-in function, it sends output to the screen.
```
maxim-belkin: Say:

> `print` is a function that _prints_ data to the screen.

_episodes/06-func.md
```diff
@@ -804,6 +804,51 @@ readable code!
...
+> This allows the *user* to see the data.
+> `return` controls the flow of data between different functions.
```
maxim-belkin: > `return` allows data to be seen by the program.

_episodes/06-func.md
```diff
@@ -804,6 +804,51 @@ readable code!
...
+> It allows the *program* to move data around.
```
maxim-belkin: I'd remove this line

_episodes/06-func.md
```diff
@@ -804,6 +804,51 @@ readable code!
...
+> Think about what the following code will give you?
+>
+> ~~~
+> def addto(a,b):
+>     print(a+b)
+> 
+> addto(2,addto(3,4))
+> ~~~
```
maxim-belkin: Let's, instead, do something like `A = add(1, 2)`

maxim-belkin:
@tlvdberg please [rebase this PR](https://github.com/swcarpentry/python-novice-inflammation/wiki/Rebasing-a-PR)

tlvdberg:
Hi Maxim,
Thanks. I applied the requested changes, and then rebased as you asked. This is the first time I did this, I hope I did it correctly.
I did not more the 'Integer Division' section to the extra exercises section, I guess that could be done in another PR. 
 Let me know if anything's wrong, and I'll correct it...

maxim-belkin:
I've made some changes and merged your exercise. Please have a closer look at it when you have a chance and feel free to submit PRs with further improvements!
Thank you, @tlvdberg!

_episodes/06-func.md
kvndrs: Next, would you indicate that 'return' is not like print in that it does not send output to the screen?

PR #526, opened 2018-05-14
Correct pepper-lists analogy

### LINKED ISSUE #458: Pepper indexing analogy is incorrect in ep3 (lists)
The last image should show the first grain of pepper for the analogy to hold. That's what `x[0][0]` would refer to. The current image is more an unordered representation of 
```
for i, ival in enumerate(x):
    for j, jval in enumerate(ival):
        print(x[i][j])
```
Which gives
```
pepper
zucchini
onion
cabbage
lettuce
garlic
apple
pear
banana
```

For the analogy to be useful, it shouldn't break down at the end.

gcapes:
Fix #458.

For the pepper analogy to hold, where x is a list of lists
represented by a pepper pot full of pepper sachets,
x[0] is a sachet of pepper outside of the pot,
x[0][0] represents the first grain of pepper.

The analogy doesn't quite hold because the pepper grains
aren't ordered, but at least it correctly shows one grain
rather than all of the grains in a heap, which is clearly
incorrect.


maxim-belkin:
agreed. Thanks @gcapes

PR #530, opened 2018-05-15
Remove var swapping exercise

### LINKED ISSUE #460: Why use a list to swap contents of variables?
Episode 3 has this challenge
```
left = 'L'
right = 'R'
left, right = [right, left]
```
but there is no explanation as to why a list is used instead of 
``` 
left, right = right, left
```
which gives the same result.

gcapes:
Fix #460.


maxim-belkin:
Thank you @gcapes !

PR #533, opened 2018-05-19
Changing exercise with enumerate() to a simpler example.

lwrubel:
Addresses issue #378: "02-loop example should not suggest naive evaluation of polynomials".

### DIFF
--- a/_episodes/02-loop.md
+++ b/_episodes/02-loop.md
@@ -337,45 +337,51 @@ so we should always use it when we can.
 > {: .solution}
 {: .challenge}
 
-> ## Computing the Value of a Polynomial
+> ## Making a List of Chemical Elements
 >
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Thanks, Laura. I like your changes (with a minor note that I'm not sure about using `start=1` the first time `enumerate` is used... that is probably fine...).

Your exercise also highlights the importance of talking about lists first. There is a `lists-loops` branch that does that but I haven't had a chance to look into that switch more closely (need to verify that the logic of episodes does not break somewhere).

lwrubel:
Thanks, yes, I think this would require that lists come before loops in the curriculum. FWIW,  I totally support having lists before loops, especially for making it possible to have relevant and realists exercises and challenges with loops.

maxim-belkin:
I've added `status:wait` label to indicate that this issue depends on having episode on `lists` before the episode on `for` loops

Issue #494, opened 2018-04-10
Clean up SVG figures


maxim-belkin:
We have quite a few figures in our repository and they have to be easily maintainable. Therefore, we need to clean them up so that they:

1. are minimal in size (don't have unnecessary information)
2. are well formatted (using 2- or 4-space indentation) 
```
<svg xmlns="http://www.w3.org/2000/svg" width="200" height="30">
  <text x="10" y="20" stroke="none" fill="red">
    print("SWC SVG FTW")
  </text>
</svg>
```
3. don't use extra fonts
4. don't use jargon

Please feel free to submit PRs and reference this issue.
Each such PR should modify one figure only (not all at once).

akeshavan:
hi @maxim-belkin : I've cleaned up 1 file in this PR #537 -- let me know if you like the new format, and I can do the remaining images.

maxim-belkin:
Example SVG:
```
<svg xmlns="http://www.w3.org/2000/svg" width="2500" height="1500">
 <defs>
    <style type="text/css"><![CDATA[
      .gray { fill: rgb(220,220,220); stroke-width:0; }
      .white { fill: none; stroke-width:2; stroke: #000; }
      .text { fill: #000; font-size: 60px; text-anchor: middle;  }
    ]]></style>
 </defs>
 <rect width="300" height="100" class="gray" />
 <rect width="300" height="100" y="150" class="gray" />
 <rect width="300" height="100" x="400" class="white" />
 <rect width="300" height="100" x="400" y="150" class="white" />
 <text x="150" y="70" class="text">final</text>
 <text x="150" y="220" class="text">original</text>
 <text x="550" y="70" class="text">0.0</text>
 <text x="550" y="220" class="text">32</text>
</svg>
```

maxim-belkin:
I wrote a Python program that can be used as the first step in cleaning up / optimizing SVG files:

[fig/optimize_svg.py](https://github.com/swcarpentry/python-novice-inflammation/blob/gh-pages/fig/optimize_svg.py).

It can use [svgcleaner](https://github.com/RazrFalcon/svgcleaner), [scour](https://github.com/scour-project/scour), and [svgo](https://github.com/svg/svgo). It also performs steps to clean up SVG figures generated with Matplotlib.

vinisalazar:
Hi @maxim-belkin,

I tried running your script and got the following results:

**Original repo size:** 1665560 bytes (1.7 MB)
[...]

PR #537, opened 2018-05-23
addressing issue #494: Clean up SVG figures

akeshavan:
For issue #494 : I cleaned up one SVG so far, `fig/python-call-stack-07.svg`, so that it is smaller and  formatted with 2 spaces. If this looks good, I will do this for the remaining SVG figures.


maxim-belkin:
Hi Anisha! Thank you for working on this. You chose a difficult issue for your first contribution to SWC :) Anyways, here are my comments.

1. Text has almost disappeared from the figure.
2. I was looking for a bit more stripped-down SVG, something like this (incomplete):
```
<svg xmlns="http://www.w3.org/2000/svg" width="2500" height="1500">
 <defs>
    <style type="text/css"><![CDATA[
      .gray { fill: rgb(220,220,220); stroke-width:0; }
      .white { fill: none; stroke-width:2; stroke: #000; }
      .text { fill: #000; font-size: 60px; text-anchor: middle;  }
    ]]></style>
 </defs>
 <rect width="300" height="100" class="gray" />
 <rect width="300" height="100" y="150" class="gray" />
 <rect width="300" height="100" x="400" class="white" />
 <rect width="300" height="100" x="400" y="150" class="white" />
 <text x="150" y="70" class="text">final</text>
 <text x="150" y="220" class="text">original</text>
 <text x="550" y="70" class="text">0.0</text>
 <text x="550" y="220" class="text">32</text>
</svg>

```
Dashed lines: 
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-dasharray

maxim-belkin:
@akeshavan, as an alternative, you could look at the images used in the `01-numpy` lesson.
Since the change in `weight` from 55 to 60, the images used in this lesson are not correct...
You could use it to practice the art of svg

maxim-belkin:
No response. Closing for now.

PR #538, opened 2018-05-23
01-numpy: add a few section headers. Improve text

maxim-belkin:


_episodes/01-numpy.md
```diff
@@ -80,10 +125,18 @@ weight in pounds: 132.0
...
+~~~
+{: .language-python}
+
+~~~
+60
+~~~
+{: .output}
 
-We can also change a variable's value by assigning it a new one:
+To change variable's value, we have to assign it a new one:
```
ldko: Depending on the type of variable (immutable vs. mutable), it is not true that "we have to assign it a new one" to change the value. I find the old wording more accurate. If you want to use the new wording, it should say "a variable's value".

_episodes/01-numpy.md
```diff
@@ -29,34 +29,68 @@ keypoints:
...
+ - `weight` and `Weight` are different variables
+
+## Types of data
+Python knows various types of data. The most common ones are:
+
+* integer numbers
+* floating point numbers, and
+* strings.
+
+In the example above, variabe `weight_kg` has an integer value of `60`.
```
ldko: Typo: "variabe"

_episodes/01-numpy.md
```diff
@@ -29,34 +29,68 @@ keypoints:
...
-[case sensitive]({{ page.root }}/reference/#case-sensitive).
-We can create a new variable by assigning a value to it using `=`.
-When we are finished typing and press <kbd>Shift</kbd>+<kbd>Return</kbd>,
-the notebook runs our command.
+From now on, whenever we use `weight_kg`, Python will substitute the value we assigned to
+it. In essence, **a variable is just a name for a value**.
+
+In Python, variable names:
+
+ - must begin with a letter, and
```
ldko: I realize that the statement "must begin with a letter" is not new to the episode here, but variable names are not actually required to begin with a letter. They could begin with an underscore. Would it make sense to specify that the variable names can include upper and lowercase letters, numbers, and underscores, but cannot start with a number?

_episodes/01-numpy.md
```diff
@@ -29,34 +29,68 @@ keypoints:
...
+ - are [case sensitive]({{ page.root }}/reference/#case-sensitive).
 
-Once a variable has a value, we can print it to the screen:
+This means that, for example:
+ - `weight0` is a valid variable name, whereas `0weight` is not
+ - `weight` and `Weight` are different variables
+
+## Types of data
+Python knows various types of data. The most common ones are:
```
ldko: Are these "the most common"? It could say "Three common ones".

_episodes/01-numpy.md
```diff
@@ -29,34 +29,68 @@ keypoints:
...
+
+And to create a string we simply have to add single or double quotes around some text, for example:
+
+~~~
+weight_kg_text = 'weight in kilograms:'
+~~~
+{: .language-python}
+
+## Using Variables in Python
+To display the value of a variable to the screen in Python, we can use `print` function:
```
ldko: Should this say "the `print` function"?

_episodes/01-numpy.md
```diff
@@ -96,19 +149,18 @@ weight in kilograms is now: 65.0
 ~~~
 {: .output}
 
-If we imagine the variable as a sticky note with a name written on it,
-assignment is like putting the sticky note on a particular value:
+A variable is analoguous to a sticky note with a name written on it:
+assigning value to a variable is like putting that sticky note on a particular value.
```
ldko: Typo: analoguous
Should be "assigning a value"

alee:
:+1: @ldko

It'd be great to communally establish across all @swcarpentry/python-maintainers some consistent text that introduces core programming language concepts - we have similar content here:

http://swcarpentry.github.io/python-novice-gapminder/02-variables/

_episodes/01-numpy.md
maxim-belkin: Hm.... interesting point. How about
```
To change value of `weight_kg` variable, we have to assign it a new one:
```

>  If you want to use the new wording, it should say "a variable's value".

I believe this contradicts the first statement:

> Depending on the type of variable (immutable vs. mutable), it is not true that "we have to assign it a new one

_episodes/01-numpy.md
maxim-belkin: sounds good

_episodes/01-numpy.md
maxim-belkin: thank you

_episodes/01-numpy.md
maxim-belkin: Sounds good

_episodes/01-numpy.md
maxim-belkin: > variable names can include upper and lowercase letters, numbers, and underscores, but cannot start with a number

sounds reasonable. Shall we distinguish numbers and digits?

_episodes/01-numpy.md
maxim-belkin: thank you

maxim-belkin:
> I am wondering if there is any sort of policy or common practice on a maintainer having a second maintainer review a PR of his/hers before merging it.

I raised this question during maintainer onboarding.
Basically, there is no policy. But we do practice it when time permits and changes are more noticeable: see, for example, https://github.com/swcarpentry/python-novice-inflammation/pull/536 where I requested a review from @annefou. The main change in this PR is, as you noticed, is that I mention strings without  deviating away from the main story-line + headers.

maxim-belkin:
> It is subjective on what changes are considered noticeable though.

100% agree.

> I tend to think of PRs that could skip review (because of noticeability and time sensitivity) things like fixing a broken link, whereas PRs that change multiple explanations would benefit from a review, if not just for the sake of catching typos, because what seems clear and absolute to one person may not be interpreted the same by others.

1. This PR does not change multiple explanations.
2. Typos are typos.

This PR formally mentions strings, introduces a few "headers", and moves Notebook-specific text outside. At present, it is more important to pay attention to the overall state of the episode:
a) there is no way it can be taught in 30 minutes. It requires more like an hour.
b) it is difficult for learners because it goes through a lot of material (including slicing numpy arrays and grouping plots in matplotlib!!!!!)
c) all exercises are at the very end!

> It'd be great to communally establish across all @swcarpentry/python-maintainers some consistent text that introduces these types of core programming language concepts - we have similar content here:

There should be an agreement on what topics/concepts have to be covered but "consistent text" would not make much sense, I think...

PR #539, opened 2018-05-24
01-numpy.md: fix typos, reword, improve

maxim-belkin:
See discussion here: https://github.com/swcarpentry/python-novice-inflammation/pull/538

### DIFF
--- a/_episodes/01-numpy.md
+++ b/_episodes/01-numpy.md
@@ -60,7 +60,8 @@ it. In essence, **a variable is just a name for a value**.
 
 In Python, variable names:
 
- - must begin with a letter, and
+ - variable names can include upper and lowercase letters, digits, and underscores, but cannot
+   start with a digit
  - are [case sensitive]({{ page.root }}/reference/#case-sensitive).
 
 This means that, for example:
@@ -68,13 +69,13 @@ This means that, for example:
  - `weight` and `Weight` are different variables
 
 ## Types of data
-Python knows various types of data. The most common ones are:
+Python knows various types of data. Three common ones are:
 
 * integer numbers
 * floating point numbers, and
 * strings.
 
-In the example above, variabe `weight_kg` has an integer value of `60`.
+In the example above, variable `weight_kg` has an integer value of `60`.
 To create a variable with a floating point value, we can execute:
 
 ~~~
@@ -90,7 +91,7 @@ weight_kg_text = 'weight in kilograms:'
 {: .language-python}
 
 ## Using Variables in Python
-To display the value of a variable to the screen in Python, we can use `print` function:
+To display the value of a variable to the screen in Python, we can use the `print` function:
 
 ~~~
 print(weight_kg)
@@ -136,7 +137,7 @@ print(weight_kg)
 ~~~
 {: .output}
 
-To change variable's value, we have to assign it a new one:
+To change value of the `weight_kg` variable, we have to assign it a new one:
 
 ~~~
 weight_kg = 65.0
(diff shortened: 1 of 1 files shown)


maxim-belkin:
@ldko would you be willing to comment on this PR?

_episodes/01-numpy.md
```diff
@@ -60,21 +60,22 @@ it. In essence, **a variable is just a name for a value**.
 
 In Python, variable names:
 
- - must begin with a letter, and
+ - variable names can include letters, digits, and underscores, but cannot
```
ldko: The "variable names" doesn't need to be here.

_episodes/01-numpy.md
```diff
@@ -136,7 +137,7 @@ print(weight_kg)
 ~~~
 {: .output}
 
-To change variable's value, we have to assign it a new one:
+To change value of the `weight_kg` variable, we have to assign it a new one:
```
ldko: This could be "To change the value"...

maxim-belkin: I'll disagree here. If we say `To change the value`, the complete sentence would read:
> To change the value, we have to assign it a new one

and `it` in the second half of the sentence becomes ambiguous.

_episodes/01-numpy.md
```diff
@@ -60,21 +60,22 @@ it. In essence, **a variable is just a name for a value**.
 
 In Python, variable names:
 
- - must begin with a letter, and
+ - can include letters, digits, and underscores, but cannot
+   start with a digit
```
maxim-belkin: shall we make an independent bullet point out of`cannot start with a digit`?

ldko: Sounds good to me.

_episodes/01-numpy.md
ldko: Ok, leaving it as you have it works for me.

maxim-belkin: how about something like this:
> To change the value of the `weight_kg` variable, we have to **assign** `weight_kg` to a new value using the equals `=` sign.

_episodes/01-numpy.md
```diff
@@ -60,21 +60,22 @@ it. In essence, **a variable is just a name for a value**.
 
 In Python, variable names:
 
- - must begin with a letter, and
+ - can include letters, digits, and underscores
```
maxim-belkin: include -> contain?

_episodes/01-numpy.md
```diff
@@ -60,21 +60,22 @@ it. In essence, **a variable is just a name for a value**.
...
+ - cannot start with a digit
  - are [case sensitive]({{ page.root }}/reference/#case-sensitive).
 
 This means that, for example:
  - `weight0` is a valid variable name, whereas `0weight` is not
  - `weight` and `Weight` are different variables
 
 ## Types of data
-Python knows various types of data. The most common ones are:
+Python knows various types of data. Three common ones are:
```
maxim-belkin: common -> basic?

_episodes/01-numpy.md
```diff
@@ -90,7 +91,7 @@ weight_kg_text = 'weight in kilograms:'
 {: .language-python}
 
 ## Using Variables in Python
-To display the value of a variable to the screen in Python, we can use `print` function:
+To display the value of a variable to the screen in Python, we can use the `print` function:
```
maxim-belkin: display -> see ? (because we're talking about our selves)

_episodes/01-numpy.md
ldko: I am not sure that "basic" is better, but I think either way works.

_episodes/01-numpy.md
ldko: I think of assigning a value to a variable, not the other way around, so I suggest a slight change of removing "to":
"To change the value of the weight_kg variable, we have to assign weight_kg a new value using the equals = sign."

maxim-belkin:
Thanks @ldko

PR #540, opened 2018-05-24
01-numpy: add dotted notation header

maxim-belkin:
Dotted (Dot) notation


maxim-belkin:
Closing for now as it looks like we need to revisit the first episode more carefully.

Issue #541, opened 2018-05-24
Heterogeneous data


maxim-belkin:
As discussed in swcarpentry/python-novice-inflammation#278 it might be worth adding a callout box about heterogeneous data.

> I think a simple example like sample_age = [2, 5.5, "Unknown"] gets the point across, and is something that a scientist could see being useful.

annefou:
Thanks for this example.

PR #547, opened 2018-06-06
Added a callout box about heterogeneous lists data as suggested in #541

Ahmed42:
An example about heterogeneous lists as many users could encounter this issue.


annefou:
Thanks @Ahmed42.

PR #550, opened 2018-06-14
Changed first example function in Errors and Exceptions to execute a …

aahowel3:
…practical command rather than a list of ice creams. In student's experience functions are designed to do something. New example function also incorporates if/else statements learned in previous lesson.

### DIFF
--- a/_episodes/07-errors.md
+++ b/_episodes/07-errors.md
@@ -34,32 +34,28 @@ Let's examine one:
 ~~~
 # This code has an intentional error. You can type it directly or
 # use it for reference to understand the error message below.
-def favorite_ice_cream():
-    ice_creams = [
-        "chocolate",
-        "vanilla",
-        "strawberry"
-    ]
-    print(ice_creams[3])
-
-favorite_ice_cream()
+def lucky_dice(number):
+    outcomes=['win','lose']
+    if number >= 6:
+        print(outcomes[1])
+    if number < 6:
+        print(outcomes[2])
+lucky_dice(3)
 ~~~
 {: .language-python}
 
 ~~~
 ---------------------------------------------------------------------------
 IndexError                                Traceback (most recent call last)
-<ipython-input-1-70bd89baa4df> in <module>()
-      6     print(ice_creams[3])
-      7
-----> 8 favorite_ice_cream()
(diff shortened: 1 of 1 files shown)


_episodes/07-errors.md
```diff
@@ -70,16 +66,16 @@ You can determine the number of levels by looking for the number of arrows on th
 In this case:
 
 1.  The first shows code from the cell above,
-    with an arrow pointing to Line 8 (which is `favorite_ice_cream()`).
+    with an arrow pointing to Line 9 (which is `lucky_dice(3)`).
```
maxim-belkin: Line 1 ?

aahowel3: 1 as in the first level of the traceback.

maxim-belkin: my bad! `Line 9` is correct

maxim-belkin:
Thanks for the PR, Abigail. I'm not yet sure what I think about it, but the content-wise it looks good to me. If you could provide a better justification as to why this part should be changed - that would be very much appreciated.

_episodes/07-errors.md
```diff
@@ -34,32 +34,28 @@ Let's examine one:
...
-        "strawberry"
-    ]
-    print(ice_creams[3])
-
-favorite_ice_cream()
+def lucky_dice(number):
+    outcomes=['win','lose']
+    if number >= 6:
+        print(outcomes[1])
+    if number < 6:
```
maxim-belkin: else?

annefou: Yes definitely "else". I am not sure this new example is clearer. I agree the initial example can be found a bit unrealistic but the main advantage is that it is fairly straightforward. I wonder if we would divert attendees by adding complexity while our main objective is to demonstrate how python tackles errors.

_episodes/07-errors.md
```diff
@@ -34,32 +34,28 @@ Let's examine one:
...
+        print(outcomes[2])
+lucky_dice(3)
```
maxim-belkin: what if we change our example to:

```python
def select_favorite(items):
    print(items[3])

ice_cream = ["chocolate",  "vanilla", "strawberry"]

select_favorite(ice_cream)
```

Would that be any better?

PR #552, opened 2018-06-18
Update python-sticky-note-variables* figures

maxim-belkin:
Updating SVG figures


annefou:
I like the new figures! 
Stupid question but do we keep both .svg and .odg files?

maxim-belkin:
I'm removing corresponding .odg files in this PR and will add a README.md to that folder later on how to work with SVG files.

annefou:
Yes that would be great to give good recommendation on how to work with SVG. (I don't know myself...).

maxim-belkin:
I've added a `README.md` file to the `fig` directory with basic guidelines on how to use third-party editors.

PR #554, opened 2018-06-21
List index figure and example in 03_lists.md on new branch

pinpea:
added fig/03-lists_oddsindex.svg
updated _episodes/03-lists.md to add figure
extended example in _episodes/03-lists.md to match figure


maxim-belkin:
Gently ping, @pinpea !

PR #555, opened 2018-06-21
fig/README.md: add notes about cleaning up SVG files

maxim-belkin:
Adding notes about cleaning up SVG files.

### DIFF
--- a/fig/README.md
+++ b/fig/README.md
@@ -54,10 +54,37 @@ editor and follow these steps:
 Ubuntu font can be downloaded for free from here:
   <https://design.ubuntu.com/font/>
 
+
 ## Not an SVG expert
 
 Please don't be discouraged by the guidelines above. If you can't follow
 them, please submit your contribution "as is" and mention that you could not
-follow the guidelines for SVG. Maintainers will be happy to help you. Be sure,
-however, that you allow maintainers to commit to the branch you used to submit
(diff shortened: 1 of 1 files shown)


annefou:
Maybe we should mention that these instructions are valid for Linux and Mac-OSX? Do we need to add some information on which tools need to be installed to manipulate SVG files?

maxim-belkin:
Good points. I will mention that these are Linux/Mac instructions. 

I'm still not sure if we want to have these instructions here or somewhere in a wiki for maintainers... What do you think? I'm afraid contributors might feel overwhelmed with the amount of instructions we have :). Perhaps, we should just say:

> use any tool of your choice to create SVG images. When submitting PNG images, please provide exact instructions how these images were obtained.

annefou:
I like the idea to give some information on how the SVG files were created; it helps to show that we welcome contributions with new or updated SVG files. And of course, anyone is welcome to use any other tools.
If we don't add it in the README file, should we add it in CONTRIBUTING.md?

maxim-belkin:
Thanks for approving this, @annefou! I will make minor updates to the instructions for the tools a bit later.

maxim-belkin:
Consider using only one tool. See instructions here: https://github.com/swcarpentry/python-novice-inflammation/pull/573#issuecomment-418911655

slel:
Thanks for these tips! I propose some slight rephrasing in #623.

Issue #444, opened 2018-01-16
Making Choices - Checking our Data - Simplify example


jamkoch:
The following code example seems too confusing for a Python beginner:

```python
if numpy.max(data, axis=0)[0] == 0 and numpy.max(data, axis=0)[20] == 20:
    print('Suspicious looking maxima!')
```

I suggest using an extended code example to allow beginners to focus on the conditional statement:

```python
max1 = numpy.max(data, axis=0)[0]
max2 = numpy.max(data, axis=0)[20]
if max1 == 0 and max2 == 20:
    print('Suspicious looking maxima!')
```

---

_edited by @maxim-belkin to remove template text and add code blocks (2018-04-13)_

PR #560, opened 2018-07-05
Clarify 'checking our data' example by naming some values

jasonmotha:
… conditional statement

https://github.com/swcarpentry/python-novice-inflammation/issues/444

### DIFF
--- a/_episodes/05-cond.md
+++ b/_episodes/05-cond.md
@@ -147,7 +147,9 @@ seemed to rise like a straight line, one unit per day.
 We can check for this inside the `for` loop we wrote with the following conditional:
 
 ~~~
-if numpy.max(data, axis=0)[0] == 0 and numpy.max(data, axis=0)[20] == 20:
+max1 = numpy.max(data, axis=0)[0]
+max2 = numpy.max(data, axis=0)[20]
+if max1 == 0 and max2 == 20:
     print('Suspicious looking maxima!')
 ~~~
 {: .language-python}
@@ -174,7 +176,9 @@ Let's test that out:
 
 ~~~
 data = numpy.loadtxt(fname='inflammation-01.csv', delimiter=',')
-if numpy.max(data, axis=0)[0] == 0 and numpy.max(data, axis=0)[20] == 20:
+max1 = numpy.max(data, axis=0)[0]
+max2 = numpy.max(data, axis=0)[20]
+if max1 == 0 and max2 == 20:
     print('Suspicious looking maxima!')
 elif numpy.sum(numpy.min(data, axis=0)) == 0:
     print('Minima add up to zero!')
@@ -190,7 +194,9 @@ Suspicious looking maxima!
 
 ~~~
 data = numpy.loadtxt(fname='inflammation-03.csv', delimiter=',')
-if numpy.max(data, axis=0)[0] == 0 and numpy.max(data, axis=0)[20] == 20:
+max1 = numpy.max(data, axis=0)[0]
+max2 = numpy.max(data, axis=0)[20]
+if max1 == 0 and max2 == 20:
     print('Suspicious looking maxima!')
 elif numpy.sum(numpy.min(data, axis=0)) == 0:
     print('Minima add up to zero!')
--- a/_episodes/05-cond.md
+++ b/_episodes/05-cond.md
@@ -142,16 +142,16 @@ import numpy
 ~~~
 {: .language-python}
 
-In the first couple of plots, the maximum inflammation per day
-seemed to rise like a straight line,
- one unit per day which is suspicious.
-We can check for this inside the `for` loop we wrote in the previous episode by doing a quick check.
-The below is checking if the maximum on day 0 and day 20 is equal to the day number, 0 and 20 respectively.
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Thanks, @jasonmotha. 

I'd like to invite @annefou and @ldko to discuss this PR.

Do you think this change can be merged "as-is" or shall we update the narrative that precedes the changes? In particular, I'm talking about:

https://github.com/swcarpentry/python-novice-inflammation/blob/6a3ad177fe950ba4d2e5b4ff97d6a091fe7c0f57/_episodes/05-cond.md#L147

_episodes/05-cond.md
```diff
@@ -147,7 +147,9 @@ seemed to rise like a straight line, one unit per day.
 We can check for this inside the `for` loop we wrote with the following conditional:
 
 ~~~
-if numpy.max(data, axis=0)[0] == 0 and numpy.max(data, axis=0)[20] == 20:
+max1 = numpy.max(data, axis=0)[0]
+max2 = numpy.max(data, axis=0)[20]
```
ldko: I agree with @annefou that `max1` and `max2` are not descriptive enough names to make things less confusing. We could use more descriptive variable names identifying the significance of the values such as `max_day_0` and `max_day_20`.

_episodes/05-cond.md
```diff
@@ -147,7 +147,9 @@ seemed to rise like a straight line, one unit per day.
 We can check for this inside the `for` loop we wrote with the following conditional:
```
ldko: As @annefou says, it could be confusing to refer to the `for` loop from the previous episode. Also, if the changes presented in this PR are merged, the conditional would be preceded by two variable definitions that could be confusing when the narrative says "the following conditional."

Also, I remember in our last workshop, one of the learners asked for clarification on what checking `if numpy.max(data, axis=0)[0] == 0 and numpy.max(data, axis=0)[20] == 20` was doing. We could remove what is in line 147 and make the narrative more descriptive of what we are checking with text more along the lines of:

```
Since it's unlikely that for each day of the study the maximum level of inflammation would change by exactly one unit, let's check for this condition using a conditional statement. First we will get the maximum inflammation values for the first day and after twenty days. If in both cases the value is equal to the number of days that have passed since the study began, we will print an alert to indicate we should later look more closely at our data for errors.
```

ldko: You are right @maxim-belkin , we are only checking that the maximum value of inflammation is equal to the number of days that have passed on the study on two specific days not actually that the max changes one unit on all days. I should have used words more like "let's do a quick check for the possibility of this condition." I certainly don't expect my exact wording to be used; I more meant to indicate that I have observed the lack of explanation of what the conditional is testing to also cause confusion in the past and if the narrative is being changed, perhaps more detail could be included. :)

jasonmotha:
@maxim-belkin @annefou 
I have made your changes as requested. Let me know if there are more improvements I can make.

_episodes/05-cond.md
```diff
@@ -143,11 +143,16 @@ import numpy
 {: .language-python}
 
 In the first couple of plots, the maximum inflammation per day
-seemed to rise like a straight line, one unit per day.
-We can check for this inside the `for` loop we wrote with the following conditional:
+seemed to rise like a straight line,
+ one unit per day which is suspicious.
```
maxim-belkin: > From the first couple of plots we saw that maximum daily inflammation exhibits a suspicious behavior and raises one unit a day.

_episodes/05-cond.md
```diff
@@ -143,11 +143,16 @@ import numpy
...
+We can check for this inside the `for` loop we wrote in the previous episode by doing a quick check.
```
maxim-belkin: > Let's check for such behavior and report it as 'suspicious'.

_episodes/05-cond.md
```diff
@@ -143,11 +143,16 @@ import numpy
...
+The below is checking if the maximum on day 0 and day 20 is equal to the day number, 0 and 20 respectively.
```
maxim-belkin: > However, let's not check the entire study but instead do a simplified test and check if maximum inflammation in the beginning and in the middle of the study were equal to the day number.

_episodes/05-cond.md
```diff
@@ -143,11 +143,16 @@ import numpy
...
+This is a simplified check, only checking day 0 and day 20 but will alert us to suspicious data.
```
maxim-belkin: This line can be removed once the previous line is rephrased.

maxim-belkin:
@ldko please review this PR once @jasonmotha addresses my comments.

jasonmotha:
@maxim-belkin Is this able to be merged into the repository?

slel:
Could you change the title of the PR to the following?

> Clarify 'checking our data' example by naming some values

Currently it has a typo ("simplifing").

maxim-belkin:
Thanks, @jasonmotha and @jamkoch!

maxim-belkin:
@jasonmotha 
You now need to execute `git reset --hard origin/gh-pages` because you used your `gh-pages` branch to submit the PR.

PR #561, opened 2018-07-10
05-cond.md: explain the need to import numpy library

maxim-belkin:
Numpy has been already imported, so there is no need to do it again.

CC @ldko @annefou

### DIFF
--- a/_episodes/05-cond.md
+++ b/_episodes/05-cond.md
@@ -135,6 +135,14 @@ at least one test is true
 
 Now that we've seen how conditionals work,
 we can use them to check for the suspicious features we saw in our inflammation data.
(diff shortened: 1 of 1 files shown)


maxim-belkin:
OK, absence of strong opinions indicates that there's got to be a better way.

I guess I don't like this part:

> Since we are going to be working with data

I can rephrase it as a reminder to:

```markdown
We will be using functions provided by the `numpy` module. 
Therefore, if you're using a new Python session, make sure to load the module again with:

~~~
import numpy
~~~
{: .language-python}
```

maxim-belkin:
Thank you both for your comments, @ldko and @annefou!

PR #569, opened 2018-07-18
Change variable name from 'char' to 'letter'

donnaaboise:
I propose changing the variable name used to introduce Python loops from 'char' to 'letter'.   The concern is possible confusion with the use of `char` in compiled languages.

### DIFF
--- a/_episodes/02-loop.md
+++ b/_episodes/02-loop.md
@@ -97,8 +97,8 @@ Here's a better approach:
 
 ~~~
 word = 'lead'
-for char in word:
-    print(char)
+for letter in word:
+    print(letter)
 
 ~~~
 {: .language-python}
@@ -116,8 +116,8 @@ hundred-letter string --- and more robust as well:
 
 ~~~
 word = 'oxygen'
-for char in word:
-    print(char)
+for letter in word:
+    print(letter)
 ~~~
 {: .language-python}
 
@@ -145,7 +145,7 @@ Using the oxygen example above, the loop might look like this:
 
 ![loop_image](../fig/loops_image.png)
 
-where each character (`char`) in the variable `word` is looped through and printed one character
+where each character (`letter`) in the variable `word` is looped through and printed one character
 after another. The numbers in the diagram denote which loop cycle the character was printed in (1
 being the first loop, and 6 being the final loop).
 
@@ -158,8 +158,7 @@ of the loop body (e.g. `end for`); what is indented after the `for` statement be
 > ## What's in a name?
 >
 >
-> In the example above, the loop variable was given the name `char` as a mnemonic;
-> it is short for 'character'.
+> In the example above, the loop variable was given the name `letter` as a mnemonic;
 > We can choose any name we want for variables. We might just as easily have chosen the name
 > `banana` for the loop variable, as long as we use the same name when we invoke the variable inside
 > the loop:


_episodes/02-loop.md
```diff
@@ -158,8 +158,7 @@ of the loop body (e.g. `end for`); what is indented after the `for` statement be
 > ## What's in a name?
 >
 >
-> In the example above, the loop variable was given the name `char` as a mnemonic;
-> it is short for 'character'.
+> In the example above, the loop variable was given the name `letter` as a mnemonic;
```
maxim-belkin: this is no longer a mnenomic, right?

maxim-belkin:
Thanks @donnaaboise. This change would also require updating the related figure.

_episodes/02-loop.md
donnaaboise: Hm... strictly speaking, I don't think `char` was a mnemonic either.  Perhaps it would be better to write "descriptive variable" in stead of mnemonic?

donnaaboise:
I have looked around in the files that generate this lesson, and don't see how to change the figure.  What am I missing?

annefou:
You are right. The figure loops_image.png is located in the fig directory but I can't find its corresponding svg file.

donnaaboise:
Good - so at least I was looking in the right place.

_episodes/02-loop.md
donnaaboise: I rephrased this line

maxim-belkin:
Nice! Thanks for finding this, @ldko.

@donnaaboise, this is how to get this figure back to your local repo:

```
git checkout af0e017^1 fig/loops_image.svg
```

donnaaboise:
I checked out the SVG file, but really don't have any idea how to edit it.  Just viewing it as a text file (after tidying it up), there is no place where I can obviously change 'char' to 'letter'.   I didn't make much progress in Inkscape either.  I pushed the SVG file to the repo (the untidy  version - I can also push a tidy version, if that helps).   Can someone let me know how to edit this file?  Or maybe make the few minor changes?  Thanks.

maxim-belkin:
Prepared two versions (don't mind the lack of spaces in front of `print(letter)` in one of them - I will fix this later).

The first one is just an updated version of the "old" figure.

In the new one I tried to improve the explanation: variable `letter` is a messenger/mediator between `word` and `print` function.

# 1. Updated figure
loops_image-new-plain

# 2. New version
loops_image-new-version

Please share your thoughts and ideas how to improve the figures.

maxim-belkin:
Forgot to mention: please comment on all aspects of figures: font, font size, colors, line widths, (missing) arrowheads, rounded corners of rectangles, etc.

maxim-belkin:
What we can change in the 'new version' of the figure: we can move the line from `letter` to `print(letter)` to connect the two `letter`s, rather than the middle points of two rectangles. Then, we can move the line that comes out of `print(letter)` to start under `print` to emphasize the point that it is `print` that prints things to the screen.

donnaaboise:
I'll admit that I am not fan of either of these figures, and wonder if they are even necessary.  Of course, they are very nicely done, but I think they may be more confusing than helpful.  For example, it might not be obvious to a beginner what the numbers mean, since no numbers are printed by the actual code, and there is no reference to an index in the code.  

I wonder if a second code fragment would be helpful : 

    for i in range(len(word)):
        letter = word[i]
        print([i,letter])

This could precede the example without the indexing and might help to make the figures more clearly understood.  Then the second (original) code could be presented as  the "python" way of writing the loop. 

I also appreciate just presenting the original code. It is certainly cleaner.  If the lesson emphasizes that the letters will be printed in order (this might not be obvious from the more Pythonic code), the figures might be more easily understood.

maxim-belkin:
This PR changes lines 100 -- 160. 
`len` is first mentioned on line 240. Range is mentioned on lines ~260.
Lists (in the current version of the lesson) are mentioned in the next episode.

I'm OK with dropping the figure altogether as I'm not a big fan of it either but visuals usually help understand concepts. We just have to make a good visual if we're to keep it.

donnaaboise:
Okay - then there does need to be a way to indicate that the letters are printed in order, and a figure could certainly help.  I think about what can be done to tweak the current figure.  For example, do we need the numbers (indices)?

donnaaboise:
This looks great!  

The thing that a figure can't quite get is the dynamic nature of the loop and the idea that the code is called several times, one iteration at at time.    So I wonder if adding something like : 

 'first pass through loop', 'second pass through loop' and so on

above each of your 6 figures would help with this.  

In some sense, the very first code fragment, which prints `word[0]`, word[1]`, `word[2]` and so on illustrates this idea best.  So it is tempting to just use this fragment : 

    for i in [0,1,2,3,4,5]:
        letter = word[i]
        print(letter)

would illustrate what is going on.  But if they haven't seen lists, this could be confusing.

maxim-belkin:
I like your suggestion to add explanatory text. I like the `for` loop too (it would be easy to explain it) but, as you understand, we can't do that until we swap the order of episodes.

abostroem:
Apologies for being late to the conversation. I have found that people transitioning from an index based language to an object oriented language often struggle with the idea of looping over objects (and not indices). I want to raise the concern that this example could enhance that confusion and is no demonstrating the loop idea of for object in list...  .

maxim-belkin:
@abostroem, do you mean this example or this figure/image?

maxim-belkin:
or what I proposed [here](https://github.com/swcarpentry/python-novice-inflammation/pull/569#issuecomment-410302947)

abostroem:
The figure looks great!
My concern is about the code fragment and changing it from:
`for letter in word:`
to 
`for i in [0, 1, ...]:` or for `i in range(len(word)):`

maxim-belkin:
What if we delete the current figure (line 146 in the episode and the PNG file) in this PR and save its SVG version for the future... It looks like we again arrive at the necessity to talk about Python lists before Python `for` loops and this change (even though I already started it) might take some time.

maxim-belkin:
Also, for our own reference this figure was added in 9c8b5f4568cc99d43fa63c9e7a16562b485fc0dc

donnaaboise:
One big difference between index-based languages (C, Fortran) and object languages (Python, Matlab) is that in both Python and Matlab, the loop variable takes iterates through items in a list or array.  Whether the code is 

    % Matlab
    for i = [0,1,2,3,4,5]
        disp(i)
    end

or the equivalent Python code, or the `letter` code, the loop iterates over values in a list.    I would think that a simple introduction to this concept could be useful, even if a more complete discussion of lists were saved for a following episode.

knarrff:
The code in question does not assign the value of the string slice to a variable anymore, removing the motivation of the original request. Would it be ok to close this PR?

Note: I have just opened a similar PR for a similar case at another place in the material. However, there the focus is not on slicing (but ```for```-loops), so there, a rename is all that is needed. (https://github.com/swcarpentry/python-novice-inflammation/pull/1028).

PR #570, opened 2018-08-02
01-numpy.md: minor code result / content fixes

tylerjereddy:
In preparing for an instructor training test I noticed a few minor issues in the first episode; a few typos and a few cases where `weight_kg` is printed as an integer when it was actually reassigned as a floating point value before the print. That's probably a drawback of hard-coding the results of the prints in as regular text.

### DIFF
--- a/_episodes/01-numpy.md
+++ b/_episodes/01-numpy.md
@@ -100,7 +100,7 @@ print(weight_kg)
 {: .language-python}
 
 ~~~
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Thank you, @tylerjereddy!

All of the changes are obvious typo fixes, so I'm going to merge this without asking for additional reviews.

PR #573, opened 2018-09-05
Redraw and cleanup python function layout figure

leouieda:
Doing some work on #494 

Redrew the `python-function.svg` in Inkscape to replace the automatically generated one. Seemed like overkill to me and used png embeds and SVG that wouldn't render in Inkscape. I added a marker for the function body which is mentioned in the text but isn't shown in the figure. I also matched the colors used on the website.

This is what the new figure looks like:

python-function

Compared to what the original looked like on my machine:

2018-09-04-160653_3200x1800_scrot

I did my best to clean up the SVG a bit but I'm not entirely sure what I can delete and messing with the indentation was messing up the figure so I gave up.


maxim-belkin:
Fantastic work, @leouieda!
Let's split this PR into two: 
(1) for changes related to the SVG figure and 
(2) for the changes in `06-func.md`

I cleaned up your SVG figure even further (from 9K to 1K) and pushed it to your branch.


For your reference, here is the command that I used to get a cleaner version:
```
$ svgcleaner --indent 2 --ungroup-defs no --coordinates-precision 1 --properties-precision 1 --paths-coordinates-precision 1 python-function.svg v2.svg
```
then I did some manual cleanup (removed duplicate markers, unnecessary tspans, empty <text> block, wrong css (font-size 16)  ... a few other things) and then repeated the cleaning procedure (just in case):

```
$ svgcleaner --indent 2 --ungroup-defs no --multipass --coordinates-precision 1 --properties-precision 1 --paths-coordinates-precision 1 v2.svg new-svg.svg
```

leouieda:
@maxim-belkin thanks for sharing the cleaning commands! I'll keep those in my toolbox for next time.

The changes to `06-func.md` were mostly my editor stripping whitespace. Sorry about that, should have cleaned that up before committing. The only real change is to remove the comment regarding how the figure is generated. I thought that would be good to have here because it's no longer accurate with the new figure in place.

maxim-belkin:
>  I thought that would be good to have here because it's no longer accurate with the new figure in place.

OK, agreed!

BTW, most of the time I also type `svn` and then correct it to `svg` :)

maxim-belkin:
Thanks again, @leouieda!

PR #576, opened 2018-09-24
Remove quotes from output

saraedum:
as these are not actually printed


maxim-belkin:
Excellent catch! Thanks, @saraedum!

PR #577, opened 2018-10-10
Update 02-loop.md

ajtag:
the language in the psudocode  includes the python keyword "with"
i have updated the psudocode to change with  to using so as to be one less thing for a learner to trip over

---

---

### DIFF
--- a/_episodes/02-loop.md
+++ b/_episodes/02-loop.md
@@ -137,7 +137,7 @@ The general form of a loop is:
 
 ~~~
 for variable in collection:
-    do things with variable
+    do things using variable
 ~~~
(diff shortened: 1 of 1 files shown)


maxim-belkin:
I like the change but I don't like the explanation. 😄  We do not talk about `with` keyword in this lesson, so it's not really a tripping point. But we also do not do things _with_ the variable (changing variable inside of a loop is a bad practice) but _using_ the variable... So, again, I like the change and I approve it, but the fact that there is `with` keyword is not really important here.

maxim-belkin:
Thanks @ajtag!

Issue #498, opened 2018-04-13
Long lines in lessons


maxim-belkin:
We need to shorten the following long lines in lessons (based on `make lesson-check-all`):

 - [ ] ./CONDUCT.md: Internally-defined links may be missing definitions: "Carpentry Code of Conduct"=>"coc", "reporting guidelines"=>"coc-reporting"
 - [ ] ./README.md: Line(s) are too long: 1, 2, 8
 - [ ] ./_episodes/01-numpy.md: Line(s) are too long: 32, 39, 40, 49, 140, 160, 175, 177, 178, 199, 227, 311, 322, 323, 345, 399, 560, 766, 820
 - [ ] ./_episodes/02-loop.md: Line(s) are too long: 115, 148, 149, 154, 160, 161, 180
 - [ ] ./_episodes/03-lists.md: Line(s) are too long: 114, 115, 118, 119, 120, 123, 155, 156, 175, 250, 269, 270, 287, 351, 451, 468, 525, 553
 - [ ] ./_episodes/04-files.md: Line(s) are too long: 36, 46
 - [ ] ./_episodes/05-cond.md: Line(s) are too long: 96, 314, 368
 - [ ] ./_episodes/06-func.md: Line(s) are too long: 46, 69, 229, 358, 371, 406, 409, 426, 502, 519, 542, 621, 710, 927, 999
 - [ ] ./_episodes/07-errors.md: Line(s) are too long: 58, 80, 86, 110, 193, 240, 284, 386, 480, 537
 - [ ] ./_episodes/08-defensive.md: Line(s) are too long: 52, 95, 97, 99, 102, 114, 151, 182, 459, 469
 - [ ] ./_episodes/10-cmdline.md: Line(s) are too long: 66, 67, 145, 312, 596, 760, 898, 926, 941
 - [ ] ./_extras/discuss.md: Line(s) are too long: 57, 62, 148
 - [ ] ./index.md: Line(s) are too long: 26, 58
 - [ ] ./reference.md: Line(s) are too long: 23, 45, 64, 161, 180, 247
 - [ ] ./setup.md: Line(s) are too long: 8, 24

doctornerdis:
Just for my own learning, why is it important to limit the lines to 100 characters? Is it about readability of the code? Are there other formatting conventions I should keep in mind?

doctornerdis:
Also, please ignore #519. I forgot to delete my old fork before I started work on the new file.

[...]

PR #579, opened 2018-10-12
fix long lines _episodes/06-func.md #498

katkoler:
Hi, one of 10 PRs for #498

### DIFF
--- a/_episodes/06-func.md
+++ b/_episodes/06-func.md
@@ -27,7 +27,7 @@ keypoints:
 - "Parameters can be passed by matching based on name, by position, 
    or by omitting them (in which case the default value is used)."
 - "Put code whose parameters change frequently in a function, 
-then call it with different parameter values to customize its behavior."
+   then call it with different parameter values to customize its behavior."
 ---
 
 At this point,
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Thanks, @katkoler!

What would you say about lining up the text that you moved to the new lines with the above lines?
I mean, change this:
```md
- "If a variable is not defined within the function it is used, 
Python looks for a definition before the function call"
```
to this:
```md
- "If a variable is not defined within the function it is used, 
   Python looks for a definition before the function call"
```

katkoler:
That's a good point! I did it to some, but I'll double check the other ones for this as well.

_episodes/06-func.md
```diff
@@ -18,12 +18,16 @@ keypoints:
...
 - "Put docstrings in functions to provide help for that function."
-- "Specify default values for parameters when defining a function using `name=value` in the parameter list."
-- "Parameters can be passed by matching based on name, by position, or by omitting them (in which case the default value is used)."
-- "Put code whose parameters change frequently in a function, then call it with different parameter values to customize its behavior."
+- "Specify default values for parameters when defining a function using `name=value` 
+   in the parameter list."
+- "Parameters can be passed by matching based on name, by position, 
+   or by omitting them (in which case the default value is used)."
+- "Put code whose parameters change frequently in a function, 
+then call it with different parameter values to customize its behavior."
```
maxim-belkin: one more line to fix

katkoler: Good spot! Thanks :)

PR #580, opened 2018-10-12
fix long lines _episodes/08-defensive.md

katkoler:
Hi, 2/10 PRs for #498

### DIFF
--- a/_episodes/08-defensive.md
+++ b/_episodes/08-defensive.md
@@ -9,13 +9,17 @@ objectives:
 - "Add assertions that check the program's state is correct."
 - "Correctly add precondition and postcondition assertions to functions."
 - "Explain what test-driven development is, and use it when creating new functions."
-- "Explain why variables should be initialized using actual data values rather than arbitrary constants."
+- "Explain why variables should be initialized using actual data values
+rather than arbitrary constants."
 keypoints:
-- "Program defensively, i.e., assume that errors are going to arise, and write code to detect them when they do."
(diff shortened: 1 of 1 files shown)


_episodes/08-defensive.md
```diff
@@ -148,7 +158,8 @@ AssertionError                            Traceback (most recent call last)
 
 <ipython-input-20-408dc39f3915> in normalize_rectangle(rect)
       1 def normalize_rectangle(rect):
-      2     '''Normalizes a rectangle so that it is at the origin and 1.0 units long on its longest axis.'''
+      2     '''Normalizes a rectangle so that it is at the origin and 1.0 units long 
+      on its longest axis.'''
```
maxim-belkin: Could you please re-run the above code cells (I think you need only two) to re-generate this error and replace it entirely?

maxim-belkin: For simplicity, you can size your terminal to be 100-characters wide and copy here what you get.

katkoler: Looks like the script for the function has updated, so the line numbers in the error message were also updated and the long line issue avoided.

katkoler:
I've spotted that there another 2 error messages with the line numbers not matching the current function. I know this is not in the scope of this issue, shall I open another one/submit another PR/leave it?

maxim-belkin:
No-no, don't leave it! We don't like long lines. Please fix it in another PR.

PR #585, opened 2018-10-12
fix long lines reference.md #498

katkoler:
7/10 PRs for #498

### DIFF
--- a/reference.md
+++ b/reference.md
@@ -20,7 +20,8 @@ assertion
     Programmers typically put assertions in their code to check for errors;
     if the assertion fails (i.e., if the expression evaluates as false),
     the program halts and produces an error message.
-    See also: [invariant](#invariant), [precondition](#precondition), [postcondition](#postcondition).
+    See also: [invariant](#invariant), [precondition](#precondition), 
+    [postcondition](#postcondition).
 
 assign
 :   To give a value a name by associating a variable with it.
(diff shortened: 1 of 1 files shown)


reference.md
```diff
@@ -175,18 +179,20 @@ postcondition
...
 return statement
-:   A statement that causes a function to stop executing and return a value to its caller immediately.
+:   A statement that causes a function to stop executing and return a value 
+    to its caller immediately.
 
 RGB
 :   An [additive model](#additive-color-model)
     that represents colors as combinations of red, green, and blue.
-    Each color's value is typically in the range 0..255
+    Each color's value is typically in the range 0-255
```
maxim-belkin: What was the motivation for this change? :)

katkoler: I'm sorry, that's out of scope. I thought it was a typo as I have never seen ranges written with ".." and after a quick google it didn't seem the most common way to indicate a range. Does it have special meaning? 
Happy to change it back.

maxim-belkin: I don't think single dash `-` indicates a range either, though. En-dash (does)[https://www.thepunctuationguide.com/en-dash.html]. To get en-dash we have to use two dashes `--` in text.
 
Regarding ranges: `git` uses two dots to indicate a range (sort of) but I'm sure that was not the reason for using two dots here.

I'd suggest reverting this change here (as the main changes have to be related to wrapping long lines) and addressing it separately.

Instead of dashes or dots, we could just use words...

katkoler: Ah yes, was going for the en-dash. But, yes you are right, words will probably make it clearer and improve readability. 
I will revert the change here.

maxim-belkin: Thank you!

maxim-belkin:
Thank you, @katkoler, a lot for all the pull requests and your willingness to cooperate! 🎉

katkoler:
@maxim-belkin Thank you for all your help! You made my first contribution an enjoyable experience!

PR #587, opened 2018-10-12
fix long lines _episodes/03-lists.md #498

katkoler:
9/10 PRs for #498


katkoler:
Indentation here doesn't look right, but it displayed right when rendered locally. 
Let me know if there is a better way of indenting it, happy to change it.

maxim-belkin:
Thanks, @katkoler! If you could please replace tabs with spaces - it will look great everywhere!

PR #591, opened 2018-10-16
refactor python-overlapping-ranges.svg source, remove odg file (towards #494)

edbennett:
ran this SVG file through a minifier, then went and hand-tuned the result. in principle it could be simplified slightly further by removing the view box, but that then basically needs redrawing everything in the new co-ordinate system so it still lines up. Still around 97% smaller than the old SVG. Removed the ODG file because it doesn't make sense to edit and export any more.


maxim-belkin:
Amazing! Thanks, Ed!
I'll see if I can quickly change the viewbox issue (as having font size of 388 is a bit weird).

edbennett:
As a PS, I also changed the hyphens to minus signs, as I'm pedantic about that kind of thing.

edbennett:
Thanks for fixing that Maxim; looks good to me.

maxim-belkin:
Thank you, Ed, very much for your contribution to the lesson! 🎉

PR #592, opened 2018-10-22
01-numpy.md: simplify wording

maxim-belkin:
1. `automatically` seems to be unnecessary as the main point is that `weight_lb` is not updated.
2. `languages like` seems to be not necessary either as at this point learners know that we're talking about Python and no need to generalize.
3. Removed two extra blank lines...


maxim-belkin:
Thanks, Anne! Fun fact: I was, actually, more certain about removing `automatically` than about the other change...

PR #600, opened 2018-11-07
use floats instead of integers when introducing functions

bast:
Apart from changing the example, I have also changed examples further
below in the episode to floats instead of integers.

Motivation:

1) The previous code produced **wrong results** in Python 2.
In Python 2 these two functions do not produce the same result:

```python
def fahr_to_celsius(temp):
    return ((temp - 32) * (5/9))

def fahr_to_celsius_better(temp):
    return ((temp - 32.0) * (5.0/9.0))
```

I know we use Python 3 but I think it is not a good idea to introduce
code which produces a different result with Python 2 which still many
people and codes use.

2) Using decimals annotates to the reader that the function is expected
to work with floats, not only with integers.  If I see an example using
integers, I assume that the function expects integers. Of course this
may be in the category "personal taste".

3) The example is more consistent with the rest of the episode where in
"Composing Functions" we suddenly switch to floats.

### DIFF
--- a/_episodes/06-func.md
+++ b/_episodes/06-func.md
@@ -50,7 +50,7 @@ from Fahrenheit to Celsius:
 
 ~~~
 def fahr_to_celsius(temp):
-    return ((temp - 32) * (5/9))
+    return ((temp - 32.0) * (5.0/9.0))
 ~~~
 {: .language-python}
 
@@ -73,16 +73,16 @@ back to whoever asked for it.
 Let's try running our function.
 
 ~~~
-fahr_to_celsius(32)
+fahr_to_celsius(32.0)
 ~~~
 {: .language-python}
 
-This command should call our function, using "32" as the input and return the function value.
+This command should call our function, using "32.0" as the input and return the function value.
 
 In fact, calling our own function is no different from calling any other function:
 ~~~
-print('freezing point of water:', fahr_to_celsius(32), 'C')
-print('boiling point of water:', fahr_to_celsius(212), 'C')
+print('freezing point of water:', fahr_to_celsius(32.0), 'C')
+print('boiling point of water:', fahr_to_celsius(212.0), 'C')
 ~~~
 {: .language-python}
 
@@ -791,12 +791,12 @@ readable code!
 > k = 0
 >
 > def f2k(f):
->     k = ((f-32)*(5.0/9.0)) + 273.15
+>     k = ((f - 32.0)*(5.0/9.0)) + 273.15
 >     return k
 >
-> f2k(8)
-> f2k(41)
-> f2k(32)
+> f2k(8.0)
+> f2k(41.0)
+> f2k(32.0)
 >
 > print(k)
 > ~~~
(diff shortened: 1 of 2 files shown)


maxim-belkin:
Thank you for the pull request, Radovan. Unfortunately I have to decline it because this lesson targets Python 3 and this is clearly stated on the first page of the lesson https://swcarpentry.github.io/python-novice-inflammation. Even in the current state this lesson is overloaded with information so we have to consider changes we make very carefully and adding Python 2 compatibility is not a priority for a number of reasons:

1. Those who're just starting to learn Python should not bother with facts about its previous version -- current code will work fine in Python 3.
2. Chances are that by the time they're confident in their Python skills, Python 2 will be EOL https://pythonclock.org
3. If you feel strongly about making this lesson Python 2 compatible, I would suggest you add a section either to `setup.md` or `index.md` to explain how to make it work for Python 2 users. Use
```py
from __future__ import print_function
from __future__ import division
```
and similar. Again, thank you for the pull request - you did a very diligent job here (you also changed the SVG figure) and I really appreciate it and feel bad for having to decline it.

bast:
OK I respect the decision but I will explain why I respectfully disagree with it:

The lesson is overloaded but my change is zero sum. Also I know about Python 2 EOL but the real life is that there is a lot of Python 2 code out there. Also I organize a number of workshops (not SWC but similar) and although we clearly advertise that this is for Python 3, there are always 1-2 people with a Python 2 environment. This example will throw them off for no reason.

I will teach this lesson this Friday and I will there manually adjust to decimals on the screen.

The `from __future__` imports would add unnecessary confusion.

I understand that we should not compromise clarity to have more compatibility but I don't think my example falls into this category.

annefou:
I disagree with you and definitely agree with Maxim. The course is for novices and we should make sure they start with python 3.

maxim-belkin:
Even though technically the change is minor, it might result in an additional and, in case of Python 3, unnecessary discussion and even confusion:

* if learners are active, they might wonder:

> why do we have to type in `32.0` and not simply `32` if we're interested in 32 F, not 32.0 F? Why measures temperature with such a precision?

* if learners are passive and no one asks this question, they might assume that `.0` has to be added to the end of all (integer) numbers, and this is not so.

I would suggest starting the lesson by demoing calculating capabilities of Python interactive shell and asking about what they get once they execute `2/3`. If someone reports `0` - proceed as you think it is most appropriate.

bast:
I accept the decision.

If somebody would ask about the extra decimal point, they will now ask one page later. My change would move this question to one page earlier and make the material more consistent.

About "course is for novices": Of course it is. But not everybody who arrives at the course arrives with a pristine Python 3 environment that they installed one day before. Again, yes, ideally everybody will have the same environment. But in real life, those who have started writing Python one year earlier and still consider them as novices and who are convinced that "they already have the right environment" may arrive with Python 2. And we are there to help them. But again, in real life it may be difficult to fix that on the spot.

I would also reject changes which complicate the Python 3 learning experience just to be compatible but again, this change is not. Python 3 people will not notice anything strange about this example, Python 2 people will not get a different result.

What I definitely want to avoid in this point is to spend 10 minutes discussing the different meanings of integer divisions which are IMO irrelevant in practice for many research codes.

The current form is not a problem for me, I will adjust and I rest my case.

annefou:
My advice (this is usually what we do) is to check installation at the beginning of the course and spend time to install python 3 if necessary. We usually have USB keys with anaconda installer to speed up the installation (WiFi is often) slow.

PR #601, opened 2018-11-22
Change x.max() to numpy.max(x)

aflaxman:
I think we intended to make all calls `numpy.max(x)` instead of `x.max()`, to reduce cognitive load for learners.


maxim-belkin:
Thanks, @aflaxman!

@annefou, I'm going to merge this one as using `numpy.method(object)` rather than `object.method` is indeed what we agreed upon

maxim-belkin:
Thank you, @aflaxman!

aflaxman:
You are welcome, and thanks to you, too! :)

Issue #602, opened 2018-12-13
Clarification of "Lesson" vs "Episode" in the text


dvalters:
In the text of some of the episodes, the words _lesson_ and _episode_ are used interchangeably in some places, which lead to a bit of confusion for in preparing for the instructor training. I think it is clear now (Lesson - a whole set of episodes on a core topic i.e. Python, Unix Shell etc.; Episode - each subsection/separate page with a number in the Lesson)

I was going to open a PR with some minor changes to fix this, if that seems reasonable?

maxim-belkin:
Absolutely! Please do so, changing one episode per PR.

maxim-belkin:
Thanks for your pull requests! Have you found any other improper uses of the word `lesson`?

dvalters:
Yep - couple more PR's to follow - just a bit caught up with other work today

PR #603, opened 2018-12-14
Clarify lesson-episode term (Episode 2)

dvalters:
Following on from #602, this clears up the term in episode 2. (Using 1 PR per episode as requested.)


maxim-belkin:
Thank you, @dvalters! 🎉

PR #605, opened 2019-01-10
Added a few sentences to help clarify the meaning of the data set.

wirawan0:
I want to help clarify the meaning of the numbers in the dataset. For a long time, whenever I read this lesson, I got hung up with not understanding what the numbers mean. Although the sentence prior to the illustrative dataset says it: "data sets of their daily inflammation", I did not get what those numbers mean. I added these two sentences to help clarify.

Wirawan

### DIFF
--- a/index.md
+++ b/index.md
@@ -22,11 +22,9 @@ The first three rows of our first file look like this:
 0,1,1,3,3,2,6,2,5,9,5,7,4,5,4,15,5,11,9,10,19,14,12,17,7,12,11,7,4,2,10,5,4,2,2,3,2,2,1,1
 ~~~
 {: .source}
-Each number represents the number of inflammation bouts that a particular patient experienced
-on a given day.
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Seems reasonable. Thank you, @wirawan0. Could you please wrap lines to 100-characters (you can copy the text in the quote below)?

> Each number represents the number of inflammation bouts that a particular patient experienced on a 
given day. For example, value "6" at row 3 column 7 of the data set above means that the third 
patient was experiencing inflammation six times on the seventh day of the clinical study.

annefou:
It would be nice to be able to close this PR.  I think I added what was requested to accept it. @maxim-belkin  let us know if it is ok. Thanks.

maxim-belkin:
Look good to me! Thank you, Anne!

wirawan0:
Thanks @annefou for doing what I ought to be doing. Thanks @maxim-belkin for accepting the PR.

PR #606, opened 2019-01-14
01-numpy.md: add a solution for an exercise

Lisanna:
Added solution for "Check Your Understanding" exercise

### DIFF
--- a/_episodes/01-numpy.md
+++ b/_episodes/01-numpy.md
@@ -856,8 +856,16 @@ the graphs will actually be squeezed together more closely.)
 > age = 122
 > mass = mass * 2.0
 > age = age - 20
+> print(mass, age)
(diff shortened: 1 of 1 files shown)


annefou:
Yes you can add a message too. We have a mixture in the lesson (sometimes with and sometimes without message). If you add a message it is nice to add a : to separate the text from the value:

print("the mass is:", mass, " the age is:", age)

versus

print(mass, age)

Maybe it is enough without a message (not too many information).

maxim-belkin:
Thanks, @Lisanna for the contribution! 
@annefou, I'm going to merge this "as is". I think this extra text would be nice but not super mega necessary. ❤️

annefou:
perfect! Thanks!

PR #607, opened 2019-01-14
Add solution to exercises, lesson 5

Lisanna:
Add solution to "What Is Truth?" and "That’s Not Not What I Meant"

### DIFF
--- a/_episodes/05-cond.md
+++ b/_episodes/05-cond.md
@@ -280,6 +280,10 @@ freeing us from having to manually examine every plot for features we've seen be
 >     print('one is true')
 > ~~~
 > {: .language-python}
+> > ## Solution
+> > A non-empty string, non-empty list and 1 are considered true.
+> > An empty string, an empty list and 0 are considered false.
+> {: .solution}
 {: .challenge}
 
 > ## That's Not Not What I Meant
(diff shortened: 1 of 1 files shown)


_episodes/05-cond.md
```diff
@@ -299,6 +303,10 @@ freeing us from having to manually examine every plot for features we've seen be
 >     print('not not True is true')
 > ~~~
 > {: .language-python}
+> > ## Solution
+> > Again, empty string is not true, and non-empty string is true.
+> > Not (not true, i.e. false) is true. 
+> {: .solution}
```
maxim-belkin: Um, I'm confused. The exercise says:

> After reading and running the code below,
> write some `if` statements that use `not` to test the rule
> that you formulated in the previous challenge.

annefou: Should we add a solution such as:

if not not '':
    print('empty string is true')
if not not 'word':
    print('word is true')
if not not []:
    print('empty list is true')
if not not [1, 2, 3]:
    print('non-empty list is true')
if not not 0:
    print('zero is true')
if not not 1:
    print('one is true')

maxim-belkin: The task is to
> write some `if` statements that use `not` to test the rule
> that you formulated in the previous challenge.

So, we have to formulate the rule first.
If the rule is
> A string is "true" as long as it contains at least one symbol.
> A list is "true" as long as it contains at least one element.
> Any non-zero number is "true".

Then we could write:
> if not 0:
>    print('not 0 is true')
> if not '0':
>     print('this will not print because our string is not empty')
> if not []:
>     print('not empty list is true')
> if not [[]]:
>     print('this will not print because our list has one element - another empty list')

annefou: Yes I agree. And it would be much clearer!

PR #611, opened 2019-01-23
episode 05 - another way to get start of string

tj-sun:
Another way to compare the start of the string using list slicing.

This doesn't introduce any new conceptual burden as learners have already seen list slicing earlier.
I also think it's a more commonly used way to compare substrings, and also helps reinforces the concept of list slicing, which is very useful.

### DIFF
--- a/_episodes/05-cond.md
+++ b/_episodes/05-cond.md
@@ -399,6 +399,16 @@ freeing us from having to manually examine every plot for features we've seen be
 > True
 > ~~~
 > {: .output}
+> Alternatively, we can use the [`list slicing`](http://swcarpentry.github.io/python-novice-inflammation/03-lists/index.html)
+> syntax that we saw earlier to check the start of a string:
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Thanks, @tj-sun. While this does not introduce any "new" material, this does break the flow of the explanation there (the sentence that starts with "But..." no longer refers to the statement that precedes it but to the one before it...).

tj-sun:
Got it - that's a good point, we certainly shouldn't digress from the logic flow. Probably better that the instructor just brings this up when he/she sees fit.

Feel free to close this.

maxim-belkin:
How about, we start with this comparison and then mention `startswith` ?
CC @annefou

annefou:
@maxim-belkin I definitely like your last suggestion; and make it clear to the instructor that this can be raised if it fits the attendance (depending on your audience and questions raised).

PR #614, opened 2019-01-28
Resolve iPython vs. Jupyter Notebook ambiguity

martinosorb:
I propose to remove two places where there's ambiguity between IPython and Jupyter notebook.

In line 605: tab completion works both in IPython (command line) and in the IPython/Jupyter notebook.

Most importantly, line 734 caused a lot of confusion in a recent workshop I was helping at, where 3-4 participants thought using "%matplotlib inline" was needed for all IPython uses (it's appropriate only when using the Notebook). And using it in the command line IPython meant the interpreter could not find the backend, and plotted nothing.

It is unlikely new users today will find anything called "IPython notebook", since it's now called "Jupyter notebook", so I removed the ambiguity.

### DIFF
--- a/_episodes/01-numpy.md
+++ b/_episodes/01-numpy.md
@@ -602,7 +602,7 @@ standard deviation: 4.61383319712
 > ## Mystery Functions in IPython
 >
 > How did we know what functions NumPy has and how to use them?
-> If you are working in the IPython/Jupyter Notebook, there is an easy way to find out.
+> If you are working in IPython or in a Jupyter Notebook, there is an easy way to find out.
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Thank you, @martinosorb!

maxim-belkin:
Quick note: because you submitted your PR from your `gh-pages` branch, you will likely have to do the following to continue contributing to our repo:
```sh
git checkout -f gh-pages
git fetch https://github.com/swcarpentry/python-novice-inflammation.git gh-pages
git reset --hard FETCH_HEAD
git push --force your-fork-nickname-that-you-use-locally
```

PR #615, opened 2019-01-28
Rename variables in `range_overlap` function

nsoranzo:
Change `low` and `high` variables to `left` and `right` respectively since that's what is normally used to call the interval endpoints.

Also I think `lowest` and `highest` are quite confusing since they are actually calculated as `max` and `min` respectively. I have therefore changed these to `max_left` and `min_right`.

I've also added an additional `assert` to justify the:
```
if not ranges:
    # ranges is None or an empty list
    return None
```
check in the final exercise solution.

Finally, I've removed an unused `import numpy` in the final exercise.

### DIFF
--- a/_episodes/08-defensive.md
+++ b/_episodes/08-defensive.md
@@ -402,18 +402,19 @@ but we're now ready to do so:
 
 ~~~
 def range_overlap(ranges):
-    '''Return common overlap among a set of [low, high] ranges.'''
-    lowest = 0.0
-    highest = 1.0
-    for (low, high) in ranges:
-        lowest = max(lowest, low)
-        highest = min(highest, high)
-    return (lowest, highest)
+    '''Return common overlap among a set of [left, right] ranges.'''
+    max_left = 0.0
+    min_right = 1.0
+    for (left, right) in ranges:
+        max_left = max(max_left, left)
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Hey, @nsoranzo, thank you for the PR and apologies for the delay. 
Could you please split this PR into separate ones so that one PR changes one file only (or as few as possible if the changes are related)? Have a look at our wiki if you need help with this.

Looking at the changes: they are pretty much all good. However, changes like `[low, high]` to `[left, right]` require at least an explanation.
Thanks again!

nsoranzo:
> Hey, @nsoranzo, thank you for the PR and apologies for the delay.
Could you please split this PR into separate ones so that one PR changes one file only (or as few as possible if the changes are related)? Have a look at our wiki if you need help with this.

Sure, I can do that, I actually spend most of my time on GitHub. I'll move the changes to episodes 7 and 9 to separate pull requests.

> Looking at the changes: they are pretty much all good. However, changes like [low, high] to [left, right] require at least an explanation.

@maxim-belkin As mentioned in the pull request description, the changes were explained in each commit message.

nsoranzo:
I've rebased this PR leaving only the changes to episode 8, updating title and description.

PR #617, opened 2019-02-03
Fix Python capitalization

nsoranzo:


maxim-belkin:
Thank you, @nsoranzo!

PR #618, opened 2019-02-03
Mention also `bisect` command

nsoranzo:
as a way to find the commit that introduced an issue in "Version Control Revisited".


maxim-belkin:
Thanks again, Nicola!

PR #620, opened 2019-02-09
README.md: Episode table: add 'Time' column

maxim-belkin:
Add a time column to the episode table. The format to be used in this column is up for a discussion. Our options (Teaching time = L, Exercise time = E, Total time = TT)
1. TT
2. L/E
3. L+E
4. L\<br>E
5. L min / E min
Other?

This PR implements option 2 above but skips `/0`.
Please cast your vote with an emoji:
1 - 👍 
2 - 👀 
3 - 😄 
4 - 🎉 
5 - ❤️ 

React with 👎 if you think this column is not necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ This lesson is also available in [R][R] and [MATLAB][MATLAB].
 
 | # |  Episode | Time | Question(s) |
 |--:|:---------|:----:|:------------|
-| 1 | [Analyzing Patient Data][episode01] | 60/30 | How can I process tabular data files in Python? |
+| 1 | [Analyzing Patient Data][episode01] | 90 | How can I process tabular data files in Python? |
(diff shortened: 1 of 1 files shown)


annefou:
👍  my only concern is that the values may not be very accurate and very often under-estimated.

maxim-belkin:
@annefou, thank you for your feedback. Did you mean that you vote for "total time" or that you agree with the PR in general?

annefou:
Yes, sorry I wasn't very clear. I definitely agree with the PR in general and I think it would be useful to have TT (for all the Carpentries lessons!) in the README file. I just wonder how to make sure the times fit what is written in the lesson. Should we automatically extract these times?

maxim-belkin:
Automatic extraction is possible but, I think, not in a Markdown file.
I'll change the PR to use total time (TT) instead of L/E.

annefou:
Ok! Thats is already great with TT!

PR #623, opened 2019-02-17
Rephrase fig/README.md

slel:
Follow-up to #555 for clarity, concision, fluidity. Open to discussion!


maxim-belkin:
Thank you very much, Samuel! This is definitely a big improvement and I like it as is! 🎉

PR #625, opened 2019-02-27
Add explanation of negative indexing before first usage

jemrobinson:
At present, negative indexing is used in lesson 3 of `Programming with Python` before it is defined. This PR adds a short explanation of negative indexing before it is first used. The later explanation of negative indexing (in the section on slicing) now uses reminder of the earlier explanation as a hint.

### DIFF
--- a/_episodes/03-lists.md
+++ b/_episodes/03-lists.md
@@ -55,7 +55,8 @@ last element: 7
 ~~~
 {: .output}
 
-Yes, we can use negative numbers as indices in Python. When we do so, the index `-1` gives us the last element in the list, `-2` the second to last, and so on.
+Yes, we can use negative numbers as indices in Python. When we do so, the index `-1` gives us the
+last element in the list, `-2` the second to last, and so on.
 Because of this, `odds[3]` and `odds[-1]` point to the same element here.
 
 If we loop over a list, the loop variable is assigned elements one at a time:
--- a/_episodes/03-lists.md
+++ b/_episodes/03-lists.md
@@ -38,8 +38,8 @@ odds are: [1, 3, 5, 7]
 ~~~
 {: .output}
 
-We can select individual elements from lists by indexing them.
-Indexing allows us to access individual elements inside a container (such as list or string).
+We can access elements of a list using indices -- numbered positions of elements in the list.
+These positions are numbered starting at 0, so the first element has an index of 0.
 
 ~~~
 print('first and last:', odds[0], odds[3], odds[-1])
@@ -51,7 +51,7 @@ first and last: 1 7 7
 ~~~
 {: .output}
 
-We already saw how to use indexing to count elements starting from the beginning of the container, but we can also use negative indices to count backwards from the end.
+We already saw how to use indexing to count elements starting from the beginning of the list, but we can also use negative indices to count backwards from the end.
 Because of this, `odds[3]` and `odds[-1]` point to the same element here.
 
 If we loop over a list, the loop variable is assigned elements one at a time:
--- a/_episodes/03-lists.md
+++ b/_episodes/03-lists.md
@@ -38,23 +38,23 @@ odds are: [1, 3, 5, 7]
 ~~~
 {: .output}
 
-We select individual elements from lists by indexing them.
+We can select individual elements from lists by indexing them.
 Indexing allows us to access individual elements inside a container (such as list or string).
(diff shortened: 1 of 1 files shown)


_episodes/03-lists.md
```diff
@@ -38,7 +38,10 @@ odds are: [1, 3, 5, 7]
 ~~~
 {: .output}
 
-We select individual elements from lists by indexing them:
+We select individual elements from lists by indexing them.
+Indexing allows us to access individual elements inside a container (such as list or string).
+Previously, we used indexing to count elements starting from the beginning, but we can also count from the end of the container using negative indices.
+We will discuss this further when we come to the topic of "slicing":
```
maxim-belkin: I suggest moving (and slightly modifying) this discussion to right after the code block. I don't think it is worth to explain negative indexing before they see it for the first time. I think this is one of those moments... "Look what I can do in Python".... "And here is why". Perhaps, it's worth even splitting the code block into `odd[0]`, `odd[3]` and `odd[-1]`. But

maxim-belkin:
Thank you, James. I agree that we have to explain negative indexing (perhaps even with a figure). But let's make it a "look what I can do" moment rather then "you're about to experience this" moment.

_episodes/03-lists.md
jemrobinson: Hi @maxim-belkin. I agree with that - I've rewritten the Markdown along those lines. Let me know if you want further changes (I think that the current phrasing might be a bit awkward).

_episodes/03-lists.md
```diff
@@ -38,20 +38,23 @@ odds are: [1, 3, 5, 7]
 ~~~
 {: .output}
 
-We select individual elements from lists by indexing them:
+We can select individual elements from lists by indexing them.
```
maxim-belkin: While we're at it, let's improve this sentence.

> We can select individual elements from lists by indexing them.

There are two problems with it: we're not selecting elements (as you correctly wrote below, we're **accessing** them), and we're not indexing them (indexing == the process of assigning indices).

I suggest saying something like this:

> We can access elements of a list using indices -- numbered positions of elements in the list. These positions are numbered starting at 0, so the first element has an index of 0.

_episodes/03-lists.md
```diff
@@ -38,20 +38,23 @@ odds are: [1, 3, 5, 7]
...
+Indexing allows us to access individual elements inside a container (such as list or string).
```
maxim-belkin: Let's remove this line -- one introductory sentence is enough, plus let's not mention "containers"

_episodes/03-lists.md
```diff
@@ -38,20 +38,23 @@ odds are: [1, 3, 5, 7]
...
 
 ~~~
-print('first and last:', odds[0], odds[-1])
+print('first and last:', odds[0], odds[3], odds[-1])
```
maxim-belkin: We can no longer say `first and last` as we have three values. What if we split it?
```py
print('first element:', odd[0])
print('last element:', odd[3])
print('"-1" element:', odd[-1])
```

_episodes/03-lists.md
jemrobinson: Done

_episodes/03-lists.md
jemrobinson: Done in 97af8a1

_episodes/03-lists.md
jemrobinson: Done in a5ba448

_episodes/03-lists.md
```diff
@@ -38,20 +38,27 @@ odds are: [1, 3, 5, 7]
...
-first and last: 1 7
+first element: 1
+last element: 7
+"-1" element: 7
 ~~~
 {: .output}
 
-and if we loop over a list,
-the loop variable is assigned elements one at a time:
+We already saw how to use indexing to count elements starting from the beginning of the list, but we can also use negative indices to count backwards from the end.
```
maxim-belkin: We've just demonstrated the `-1` index, so I would suggest saying something like this (feel free to modify/adjust):

> Yes, we can use negative numbers as indices in Python. When we do so, index of `-1` gives us the last element in the list, `-2` -- second to last, and so forth...

jemrobinson: Done

maxim-belkin: Thank you! Could you please also break this line so we don't exceed a 100 characters per line limit?

maxim-belkin:
Thank you very much for your contribution, James! 🎉

PR #630, opened 2019-03-22
Fix for Python 3

bmorris3:
In Python 3, you can't have non-default arguments followed by default arguments, as was presently implemented in the example. The old example returns a syntax error: `SyntaxError: non-default argument follows default argument`. The fix turns the offending default argument into a non-default argument, allowing the example to work.

---

---

### DIFF
--- a/_episodes/06-func.md
+++ b/_episodes/06-func.md
@@ -821,7 +821,7 @@ readable code!
 > Given the following code:
 >
 > ~~~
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Brett, please have a look at the answer # 4 and the next question -- there is a clear connection between these two questions. Because of that connection we shouldn't be fixing this question... 
Moreover, there is a solution that explains all of that...

Thanks anyway!

Issue #491, opened 2018-03-21
Episode 2 - range doesn't return an array


biavillas:
In episode 2 (_Repeating Actions with Loops_) the first exercise is called "_From 1 to N_". This is a good exercise, but I believe its description might give rise to some confusion for novices.

The exercise introduces the concept of looping through the elements of a collection using Python's built-in function "_range_". On the first bullet point explaining what range does, we have:

- _If one parameter is given, range creates an array of that length_ ...

However, the function _range_ in Python2 returns a list and in Python3 returns an object of _range_  type. Since learners had just been introduced to lists and arrays and how to find out type(some_thing), it would be important to be consistent and specific about types. 

Maybe to avoid going into the details of the type _range_ this early in the lesson, we could simply use numpy.arange (which would return an array) throughout this exercise without compromising its overall goal.

labrax:
To avoid these problems we can write this point similar to the next ones:

Instead of:
> If one parameter is given, range creates an array of that length, starting at zero and incrementing by 1. For example, range(3) produces the numbers 0, 1, 2.

We can write as:
> If one parameter is given, range starts at 0 and ends just before the parameter given, incrementing by one. For example, range(3) produces the numbers 0, 1, 2.

Using this approach we omit the explanation of the function return and explain what is the goal of the lesson.

biavillas:
Thanks for the suggestion. I agree this would be the simplest solution to the problem.

However, I'm still a little concerned about the fact that the built-in _range_  function in python3 returns an object of type _range_. I imagine that if a learner tries to understand what is going on the exercise and type 

`print(range(3))`
  
the learner will not get 0, 1, 2, which is what you would expect from  the explanation in the exercise:

> If one parameter is given, range starts at 0 and ends just before the parameter given, incrementing by one. For example, range(3) produces the numbers 0, 1, 2.

I believe this could be rather confusing.

labrax:
Indeed, my solution will not help in the case of Python 3. Thank you.

How about we create a case just like the ones before, something in the lines of:
```
    Python has a built-in function called range that assists creating a sequence of numbers. range can accept 1, 2, or 3 parameters.

    If one parameter is given, _range_ will return to _for_ an array of that length, starting at zero and incrementing by 1. For example:

    for number in range(3):
        print(number)

    0
    1
    2

    If two parameters are given, our _for_ starts at the first and ends just before the second, incrementing by one. For example:

    for number in range(2,5):
        print(number)

    2
[...]

PR #632, opened 2019-03-27
Lesson 02 - Rephrase "range" object return value - See Issue #491

appeltel:
Use the phrase "generates a sequence" to use consistent language for lazy-evaluators as iterables in a for loop and avoid suggesting that the sequence is stored as an array in memory.

See issue #491.


maxim-belkin:
Thank you, Eric and Anne! 🎉

Issue #496, opened 2018-04-12
Swapping the order of episodes on lists and loops


maxim-belkin:
It has been decided (see #453) that **lists** should be introduced _before_ **loops**.

Because this change might take some time to implement, it was decided to first pursue this effort in a separate branch and later merge this branch into the main `gh-pages` branch.

This special branch has been created and is called `lists-loops`. We, therefore, ask anyone who is interested in contributing to this effort to submit Pull Requests against this `lists-loops` branch.

### How to contribute

Start by fetching and checking out the `lists-loops` branch into your current repository

```bash
git fetch https://github.com/swcarpentry/python-novice-inflammation.git lists-loops
git checkout lists-loops
```

Now, you can start working on it as usual:
```bash
git checkout -b my-changes lists-loops
# ... make changes...
git add -u
git commit
git push <your-fork>
```
... and submit a PR! When you submit a pull request related to swapping the order of episodes on loops and lists, make sure to select `lists-loops` branch of this repository as the base branch.

That's it!

_We hope to see **your** Pull Request soon_!

bhickson:
@maxim-belkin 
For a carpentries lesson this last February I did some reorg and partial rewrite of the python lesson into Jupyter notebooks that follows this idea. Basically introducing variables, data types, structures,  statements, conditionals, loops, etc in that order. Would this order be worth considering for this issue/branch?

https://github.com/UA-Carpentries-Workshops/2018-02-10-Tucson/blob/master/python_lessons/01%20-%20Introduction%20to%20Programming%20with%20Python.ipynb

If so, I'd be happy bring them back out of the notebook to the branch.

maxim-belkin:
@bhickson, thanks for offering your help! The notebook you reference has changes that might be appropriate for both branches (`gh-pages` and `lists-loops`) so please feel free to suggest your changes _via_ pull requests.

kekoziar:
Is there anything else that needs to be worked on for this branch?

biokcb:
[...]

PR #634, opened 2019-04-07
[lists-loops branch] replace string loops with lists

biokcb:
Removing all looping over strings in the loops episode of the lists-loops episode swap branch to avoid adding in an extra concept to discuss that might not add enough value for learners. Going to read over this again to check for typos and clarity before marking it ready for review (set as a draft for now). Let me know if there's additional changes that could be made. Thanks!

In reference to #496

### DIFF
--- a/_episodes/03-loop.md
+++ b/_episodes/03-loop.md
@@ -35,7 +35,7 @@ odds = [1, 3, 5, 7]
 
 We can access a number in the list using its index. For example,
 we can get the first
-number in the list `'odds'`,
+number in the list `odds`,
 by using `odds[0]`. One way to print each number is to use
 four `print` statements:
 
@@ -192,8 +192,8 @@ of the loop body (e.g. `end for`); what is indented after the `for` statement be
 Here's another loop that repeatedly updates a variable:
 
 ~~~
-names = ['Curie', 'Darwin', 'Turing']
 length = 0
+names = ['Curie', 'Darwin', 'Turing']
 for value in names:
     length = length + 1
 print('There are', length, 'names in the list.')
@@ -207,7 +207,7 @@ There are 3 names in the list.
 
 It's worth tracing the execution of this little program step by step.
 Since there are three names in `names`,
-the statement on line 3 will be executed three times.
+the statement on line 4 will be executed three times.
 The first time around,
 `length` is zero (the value assigned to it on line 1)
 and `value` is `Curie`.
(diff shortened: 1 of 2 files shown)


_episodes/03-loop.md
```diff
@@ -192,33 +192,34 @@ of the loop body (e.g. `end for`); what is indented after the `for` statement be
...
+for value in names:
     length = length + 1
-print('There are', length, 'vowels')
+print('There are', length, 'names in the list.')
 ~~~
 {: .language-python}
 
 ~~~
-There are 5 vowels
+There are 3 elements in the list.
```
maxim-belkin: `elements` or `names` or ...?

biokcb: Yep, thanks for catching that, it's supposed to be names :)

biokcb:
Ok, I've read through it again and fixed anything else that popped up as weird/as a typo. Would appreciate feedback on whether or not the exercise I used is okay or if there is a better suggestion ("Summing a list"). Others I thought of would need to introduce the concept of creating an empty list (which, at this point, I'm wondering why it isn't mentioned at all) or having them essentially do what `enumurate` does but explicitly. Thanks!

annefou:
I wonder if removing the last example "Reverse a String" is needed. I find it quite useful. Any reason why you removed it?

biokcb:
@annefou Hi, we discussed removing iteration over strings as a concept overall in #496. Since lists are introduced before looping now, we can loop over them instead of strings. To keep the string iteration, we have to also introduce this as a concept and it does not add much utility in terms of understanding the concept of looping or usefulness in most projects. This I believe will help reduce overall cognitive load as well. I had a tough time coming up with a replacement problem for this one, so if you have suggestions on a better one that would be just as useful, that would be awesome! 

I also just realized there is another string based looping problem in a later lesson that should probably also be replaced. Any thoughts?

maxim-belkin:
@biokcb, let's move the exercise on reversing a string to `_extras/extra_exercises.md`.

annefou:
Very good suggestion @maxim-belkin 
Thanks.

maxim-belkin:
@biokcb @annefou, I added that exercise to the "Extra" page.

annefou:
Thanks @maxim-belkin

maxim-belkin:
Thanks, Anne! And thank you, Kristen, for you contribution!

PR #638, opened 2019-04-10
Update 01-numpy.md

davidbenncsiro:
Although it's true that data[3:3, 4:4] and data[3:3, :] return the empty array [], is that what we see on the output? It's array([], shape=(0, 0), dtype=float64) and array([], shape=(0, 40), dtype=float64) respectively. Should that be shown instead of [] and []? I could go either way on this but just wanted to raise it for your consideration.

---

---

### DIFF
--- a/_episodes/01-numpy.md
+++ b/_episodes/01-numpy.md
@@ -948,8 +948,8 @@ the graphs will actually be squeezed together more closely.)
 >
 > > ## Solution
 > > ~~~
-> > []
-> > []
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Hi David! Thank you for the PR. I think it makes sense because we're showing the output that people should expect to see and `data` is a numpy array, not a list.

davidbenncsiro:
Thanks @maxim-belkin

PR #639, opened 2019-04-16
Typo correction

bmorris3:
The variable name in the demo is `odds` not `odd`. Correcting this small typo.


maxim-belkin:
Thank you, Brett!

PR #641, opened 2019-04-18
02-loop.md: fix exercise

maxim-belkin:
Fix for swcarpentry/python-novice-inflammation#421

CC @annefou


maxim-belkin:
Thanks, Anne!

PR #643, opened 2019-04-24
02-loop.md: explain string (see #287)

sebranchett:


maxim-belkin:
And again, thank you, Susan!
I'd like to invite @annefou to this PR.

Here are my thoughts.
I think it might be good to blend this new sentence with the next one. I suggest using something like this:

> In Python, a string is just an ordered collection of characters, so every character has a unique number associated with it -- its index. This means that we can access characters in a string using their indices. For example, ...

or this:

> It might seem unusual at first, but a good way of thinking of strings in Python is as of ordered collections of characters: every character in a string has a unique number associated with it -- its index. This means that one can access any character in a string using its index. For example,

maxim-belkin:
ping, @annefou!

annefou:
Yes @maxim-belkin linking the sentence to the rest of the text is a good idea.  I like your first proposal:

> In Python, a string is just an ordered collection of characters, so every character has a unique number associated with it -- its index. This means that we can access characters in a string using their indices. For example, ...

Thank you!

maxim-belkin:
Susan, do you agree with the proposed changes? If you do, would you like to implement changes yourself or would you like us to make them? If you don't, please share your thoughts and ideas on this.

Thank you!
Maxim

sebranchett:
Hi @maxim-belkin, yes I agree with @annefou and I have made the change.
Thanks, Susan

annefou:
Fantastic. Thank you @sebranchett

PR #646, opened 2019-05-01
return vs print statements in python functions?

Ovo-Akpome1:
Learners should understand that print can also be used in place of return. The only down side is that it always returns its data as a string value.

---

---


maxim-belkin:
Hi Ovo-Akpome1.

The point you're trying to make is, unfortunately, incorrect: `return` statement is very different from the `print` function -- one returns a value and prints nothing, and the other prints its arguments to the screen and returns nothing (`None`).

Moreover, we already have a discussion about `return` and `print` in the lesson: https://swcarpentry.github.io/python-novice-inflammation/06-func/index.html#return-versus-print

Thank you for trying to improve our lesson! 👍 Please continue to look for the missing pieces.

Ovo-Akpome1:
Thanks for the correction🙂

Issue #622, opened 2019-02-14
Replace `assertions` by raising Errors


skuschel:
I was reading through python exercise and It seems to me that the [Defensive Programing](http://swcarpentry.github.io/python-novice-inflammation/08-defensive/index.html) session is introducing assertions, where I believe Type- or ValueErrors should be raised instead:

`Assertions` are used in algorithm development, ensuring a state, which might not be the obvious outcome or may be violated due to numerical errors (like ensuring, that the eigenvalues of a hermitean matrix are real after the matrix has been diagonlized).
Throwing an assertion indicates, that something went wrong really really deep inside an algorithm and it is usually no way to fix this as a user. That is why compiler optimizations are typically removing assertions (which happens if a program is started with `python -O`). A function precheck should raise a `TypeError` or `ValueErrror` instead.
This would have some advantages:
* This would be consistent with most python libraries.
* When students advance, they will -- at some point -- start to catch errors as well and they will try to catch `TypeError` and `ValueError` most often. An `AssertionError` should almost never be caught.
* The participants know various kinds of errors from the section before already. I find it more logical to understand that functions raise Exceptions and in the section after, I learn how to raise them myself.

For the post-checks it probably doesnt hurt to raise `ValueErrors` as well.

Regarding the test-driven development: I have done this in my own code and this is at least a few hours on its own. May it be more useful to introduce the `try... except` statement in that space? That would feel like a more coherent program to me. I dont really know how to approach this. Do you want me to open another issue for that discussion?

annefou:
This is definitely an interesting discussion and I fully agree with you that introducing `try... except` would be useful. However, I am not sure it is important for novices. We usually do not have time to go through this section (Defensive programming). I think it would be very interesting to learn once you really start to code woth Python and are familiar enough to move to the next step.

skuschel:
What do you think about the assertions? The `try... except` vs. test-driven development was just a side note. Maybe worth another issue here.

maxim-belkin:
Some time ago, I'd be strongly against adding `try...except` block to this lesson. However, I recently heard TDD (Test-Driven Development) doesn't provide any significant advantage over the standard approach in the long-run. The `try...except` block, on the other hand, is there to stay and will definitely be a useful tool for anyone. On yet another hand, we have to carefully consider the breadth of topics we cover. One of the responses that I got at SciPy 2017 was that we should really talk about `argparse`. And it's hard to disagree with that too...

[...]

PR #647, opened 2019-05-03
add a callout that shows asserts are not used for user input

varioustoxins:
add a callout that shows asserts are not used for user input
---

---

### DIFF
--- a/_episodes/08-defensive.md
+++ b/_episodes/08-defensive.md
@@ -560,5 +560,8 @@ This violates another important rule of programming:
 > > {: .language-python}
 > {: .solution}
 {: .challenge}
-
+An afternote: asserts are for debugging only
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Thanks, @varioustoxins. I don't think we can include your addition in the current lesson because (1) we don't talk about optimized mode and (2) we don't mention the fact that users can raise exceptions...

However, there is an idea about changing the episode on assertions with an episode on exceptions --  swcarpentry/python-novice-inflammation#622. So, if you're interested, it'd be great if you join this effort.

varioustoxins:
This was a trading checkout submission.I didn’t necessarily expect it to succeed as these lessons are quite carefully balanced. I will comment further on #622

PR #648, opened 2019-05-07
setup.md: fix header and add instructions to install Anaconda

maxim-belkin:
Two changes:

1. add `index.html` to permalink.

Currently, the link to the setup page works correctly online, when the site is served (locally), but not when it is built with `jekyll build`. Addint `index.html` should solve the last problem.

2. Explicitly mention that one has to install Anaconda as the first step of the setup.

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -1,9 +1,18 @@
 ---
 layout: page
 title: Setup
-permalink: /setup/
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Thank you, Anne!

PR #650, opened 2019-05-20
07-errors.md: typo

maxim-belkin:
Obvious typo: occurr -> occur

My editor also stripped an empty whitespace at the end of another line which is not ideal (I prefer to make clean commits) but I'm going to leave it the way it is because I can never justify submitting a separate PR to remove unnecessary whitespace.


maxim-belkin:
@annefou, I'm going to merge this now as this is a very simple typo fix.

PR #657, opened 2019-06-12
01-numpy.md: Loading data: update introductory text

maxim-belkin:
Update introductory text in "Loading data into Python".

Reasons:
- justify the need to load data
- remove `access a library` as it seems rather vague
- add a comma before `which stands for`
- change `you want to do fancy things with numbers` to `you want to do fancy things with lots of numbers`


maxim-belkin:
ping, @annefou

maxim-belkin:
Rebased. Thank you for reviewing this, Lauren.

PR #658, opened 2019-06-13
Cleaned up figure python-call-stack-07.svg

ssvassiliev:
Addressing issue #494 I have cleaned up the file "python-call-stack-07.svg" to make it easily maintainable. The new file has 36 lines of code including comments. If you approve changes I can work on the other images.


maxim-belkin:
Wow! It looks fantastic! The only comment I have is that you set width and height to very high values (2500 and 1500) and as a result the image is not being displayed properly on the rendered HTML page. It would be great if you could set width and height to the actual width and height. I believe you can use 705 for width and 375 for height.

ssvassiliev:
Right, no problem I'll fix it

PR #659, opened 2019-06-17
Add list type under variables

sunray1:
Arrays are used later in the lesson, but the idea that brackets equate to a list isn't explicitly described, it is implied.

### DIFF
--- a/_episodes/01-numpy.md
+++ b/_episodes/01-numpy.md
@@ -212,9 +212,9 @@ that can be called upon when needed.
 
 ## Loading data into Python
 In order to load our inflammation data, we need to access
-([import]({{ page.root }}/reference/#import) in Python terminology) a library called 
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Hi, @sunray1! Thank you for the pull request. Something went wrong with GitHub and it shows as if all lines have been changed. Could you please follow [these instructions](https://github.com/swcarpentry/python-novice-inflammation/wiki/Fixing-a-PR-created-from-the-gh-pages-branch) to see if submitting a pull request from a different branch helps or not.

maxim-belkin:
Hey, Chandra!

I looked at the PR locally and the reason GitHub shows all lines as changed is because you used a text editor that converted all Linux-style line breaks to Windows-style line breaks (that is, converted LF (Line Feed) to CRLF (Carriage Return + Line Feed) characters). I fixed that in 05d2d05 but you should instruct you Git client to convert CRLF to LF upon committing changes with:

```sh
git config --global core.autocrlf true
```

I'll review changes suggested in this PR in a separate comment

PR #660, opened 2019-07-02
Stressed that functions do not need to return a value. Issue #489.

IgnasiLucas:
This is meant to help issue #489.

### DIFF
--- a/_episodes/06-func.md
+++ b/_episodes/06-func.md
@@ -59,7 +59,7 @@ def fahr_to_celsius(temp):
 
 The function definition opens with the keyword `def` followed by the
 name of the function (`fahr_to_celsius`) and a parenthesized list of parameter names (`temp`). The
-[body]({{ page.root }}/reference/#function-body) of the function --- the
+[body]({{ page.root }}/reference/#body) of the function --- the
 statements that are executed when it runs --- is indented below the
 definition line.  The body concludes with a `return` keyword followed by the return value.
 
@@ -189,6 +189,9 @@ def detect_problems(filename):
         print('Seems OK!')
 ~~~
 {: .language-python}
+
+Did we forget the [return statement]({{ page.root }}/reference/#return-statement)?
+We didn't. It's just that not all functions need to send a result back to the caller.
 
 Notice that rather than jumbling this code together in one giant `for` loop,
 we can now read and reuse both ideas separately.
--- a/_episodes/06-func.md
+++ b/_episodes/06-func.md
@@ -191,9 +191,9 @@ def detect_problems(filename):
 {: .language-python}
 
 Wait! Didn't we forget to specify what both of these functions should return? Well, we didn't.
-In Python, functions are not required to return a value. In cases like ours, they are used for
+In Python, functions are not required to include a `return` statement and can be used for
 the sole purpose of grouping together pieces of code that conceptually do one thing. In such cases,
-function names usually describe what they do.
+function names usually describe what they do, _e.g._ `analyze`, `detect_problems`.
 
 Notice that rather than jumbling this code together in one giant `for` loop,
 we can now read and reuse both ideas separately.
--- a/_episodes/06-func.md
+++ b/_episodes/06-func.md
@@ -190,8 +190,10 @@ def detect_problems(filename):
 ~~~
 {: .language-python}
 
-Did we forget the [return statement]({{ page.root }}/reference/#return-statement)?
-We didn't. It's just that not all functions need to send a result back to the caller.
+Wait! Didn't we forget to specify what both of these functions should return? Well, we didn't.
+In Python, functions are not required to return a value. In cases like ours, they are used for
+the sole purpose of grouping together pieces of code that conceptually do one thing. In such cases,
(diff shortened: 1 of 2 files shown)


_episodes/06-func.md
```diff
@@ -190,6 +190,9 @@ def detect_problems(filename):
 ~~~
 {: .language-python}
 
+Did we forget the [return statement]({{ page.root }}/reference/#return-statement)?
+We didn't. It's just that not all functions need to send a result back to the caller.
```
maxim-belkin: My suggestion:

> Wait! Didn't we forget to specify what both of these functions should return? Well, we didn't. In Python, functions are not required to return a value. In cases like ours, they are used for the sole purpose of grouping together pieces of code that conceptually do one thing. In such cases, function names usually describe what they do.

reference.md
```diff
@@ -102,8 +102,9 @@ for loop
     See also: [while loop](#while-loop).
 
 function
-:   A group of instructions (i.e., lines of code) that transform
-    some input [arguments](#argument) to some output.
+:   A group of instructions (i.e., lines of code) that may be executed repeatedly,
+    every time it is [called](#function call). Functions can accept some input
+    [arguments](#argument), and they can [return](#return-statement) some output.
```
maxim-belkin: How about:

> A named group of instructions that is executed when function's name is used in the code. Occurrence of a function name in the code is called "function call". Functions may process input arguments and return some value(s), or may be used to group together logical pieces of code without returning any value.

I'm not super happy about "without returning any value" in my suggestion above (function will return `None`), but it might be OK in our case.

IgnasiLucas:
Being an important concept, I would not mind the definition of 'function' in the glossary to be longer and richer than novices may need. That's why I suggest mentioning there the 'None' value. Let me know if the phrasing is not OK. I already got the handle of updating the pull request :-)

reference.md
```diff
@@ -102,8 +102,12 @@ for loop
     See also: [while loop](#while-loop).
 
 function
-:   A group of instructions (i.e., lines of code) that transform
-    some input [arguments](#argument) to some output.
+:   A named group of instructions that is executed when the function's name is used in
+    the code. Occurrence of a function name in the code is a [function call](#function-call).
+    Functions may process input [arguments](#argument) and return some value(s). If functions
+    are used just to group logical pieces of code together, without a [return statement](#return-statement)
```
maxim-belkin: Suggestion: `return some value(s)` -> `return the result back`.

What do you think about changing "If functions are used..." sentence to:

> Functions may also be used for grouping together logical pieces of code. In such cases, they don't need to return any meaningful value and can be written without the return statement completely. Such functions return a special value `None`, which is a way of saying "nothing" in Python.

IgnasiLucas: Yes, I think your suggestions make the language less technical and more understandable.

maxim-belkin:
Thank you, @IgnasiLucas. The PR looks good to me.

@annefou and @ldko, could you please have a look at this PR and approve/comment if you have any suggestions?

reference.md
```diff
@@ -102,8 +102,12 @@ for loop
     See also: [while loop](#while-loop).
 
 function
-:   A group of instructions (i.e., lines of code) that transform
-    some input [arguments](#argument) to some output.
+:   A named group of instructions that is executed when the function's name is used in
+    the code. Occurrence of a function name in the code is a [function call](#function-call).
+    Functions may process input [arguments](#argument) and return the result back. Functions
+    may also be used for grouping together logical pieces of code. In such cases, they don't
```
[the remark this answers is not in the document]
maxim-belkin: Good catch! This is my bad. `logically grouping together pieces of code` makes a lot more sense than `grouping together logical pieces of code`. I'm not even sure what `logical pieces of code` are.

reference.md
```diff
@@ -102,8 +102,12 @@ for loop
...
+    need to return any meaningful value and can be written without the return statement completely.
```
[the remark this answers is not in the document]
maxim-belkin: Agreed. Using either one or both suggestions is fine with me. Your call, @IgnasiLucas.

_episodes/06-func.md
```diff
@@ -190,6 +190,11 @@ def detect_problems(filename):
 ~~~
 {: .language-python}
 
+Wait! Didn't we forget to specify what both of these functions should return? Well, we didn't.
+In Python, functions are not required to return a value. In cases like ours, they are used for
+the sole purpose of grouping together pieces of code that conceptually do one thing. In such cases,
+function names usually describe what they do.
+
```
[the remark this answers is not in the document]
maxim-belkin: Yes, I like your wording better. Thank you!
@IgnasiLucas, could you please use Lauren's suggestion?

maxim-belkin:
@ldko, I implemented your suggestions. I kept "for the sole purpose" as it makes the purpose of using such functions very specific. Let me know if you object to keeping this wording.

maxim-belkin:
> Looks good to me 👍

Thank you!

maxim-belkin:
Thank you both, @IgnasiLucas and @ldko, for working on this pull request! 🎉

Issue #670, opened 2019-07-18
Lesson 5 - typo


HeatherAn:
In Lesson 5 Conditionals section, it says 'If the test that follows the if statement is true, the body of the if (i.e., the lines indented underneath it) are executed'. It should be 'is' since it is referring to 'the body'. In the parenthesis it could say 'the set of lines'.

maxim-belkin:
Thanks again, @HeatherAn. Yes, this is a typo. Would you be willing to submit a PR? Would you need any help if so?

sarah-peter:
I took the liberty to submit a PR to fix this.

maxim-belkin:
Thank you, @sarah-peter for the PR 🥇  and
thank you, @HeatherAn, for reporting the issue 🎉 !

PR #679, opened 2019-07-23
Fix typo in Conditionals section

sarah-peter:
This is a fix for #670.


maxim-belkin:
Thank you, @sarah-peter!

PR #682, opened 2019-07-25
Update 03-lists.md

HeatherAn:
Rephrased text as stated in discussion of issue 668


maxim-belkin:
Thank you for improving our lesson, @HeatherAn! 🎖

Issue #672, opened 2019-07-18
Lesson 6 - triple quotes


HeatherAn:
In Lesson 3, in the 'Testing and Documenting' section, it refers to docstrings. It says 'we don't need to use triple quotes when we write one, but if we do, we can break the string across multiple lines'. It would be good to add here that if we use triple quotes within a function, then that text shows up when calling help (the text does not show up when using help() if we use '#' to write a comment). This is written below as code, but it is not explicitly mentioned.

maxim-belkin:
Good suggestion, @HeatherAn! Would you be willing to submit a pull request?

maxim-belkin:
Hi Heather. The lesson currently has the following wording:

> If the first thing in a function is a string that isn’t assigned to a variable, that string is attached to the function as its documentation:...
... This is better because we can now ask Python’s built-in help system to show us the documentation for the function:

I think it covers what you added in #672. Nonetheless, thank you for taking the time to submit this issue and create #672! 🥇

PR #683, opened 2019-07-25
Update 06-func.md

HeatherAn:
Triple quotes - issue 672

### DIFF
--- a/_episodes/06-func.md
+++ b/_episodes/06-func.md
@@ -363,7 +363,9 @@ offset_mean(data, target_mean_value)
 A string like this is called a [docstring]({{ page.root }}/reference/#docstring).
 We don't need to use triple quotes when we write one,
 but if we do,
-we can break the string across multiple lines:
+we can break the string across multiple lines. Also, if we use 
+triple quotes within a function, then that 
+text shows up when calling help() (this does not happen if we use # to write the text). 
 
 ~~~
 def offset_mean(data, target_mean_value):


maxim-belkin:
Thank you for the pull request, @HeatherAn!
Any string (single, double, or triple-quoted) in the beginning of a function body will produce a help message. The advantage of triple-quoted strings is that they allows us to write documentation the way we normally write and not use any special characters such as `\n` to break the line and `\` to continue working on the text on the next line:

```python
def a():
 "function a\n\
does nothing and returns 1"
 return 1

def b():
 """function b
   does nothing and returns 2"""
 return 1

help(a)
help(b)
```
will produce
```
>>> help(a)
Help on function a in module __main__:

a()
    function a
    does nothing and returns 1

Help on function b in module __main__:

b()
    function b
    does nothing and returns 2
```

Could you please update the PR to reflect this?

maxim-belkin:
@HeatherAn, I noticed that your `gh-pages` branch diverged from the `gh-pages` branch in this repository. If you're using command-line tools, you might want to execute
```sh
git fetch origin
git checkout gh-pages
git reset --hard origin/gh-pages
```
and then you can
```sh
git push --force https://github.com/HeatherAn/python-novice-inflammation.git  gh-pages
```
to update the branch on GitHub.

maxim-belkin:
@annefou, I'm closing this PR because our lesson already has the wording suggested here (see my note in #672).

@HeatherAn, if you think current wording is not sufficient / not ideal -- please feel free to submit another pull request or open an issue so we can discuss how to improve the lesson. Again, thank you for your contribution!!! I truly appreciate it! 🌷 🥇

PR #686, opened 2019-07-26
03-lists.md: Use `pop()` for consistency of method over statement

### LINKED ISSUE #669: Lesson 3 - del function
In lesson 3, when referring to 'Heterogeneous Lists' the function del is used to delete the first element of the 'odds' list. In the lesson, it appears as 'del odds[0]', which is ok. However during the lessons we have used the functions with the parenthesis, so it might be a bit confusing for the student to see that del is used with a space, while almost all the other functions are used with parenthesis. Thus I would suggest, to either use 'del(odds[0])' or explain why 'del odds[0]' also works.

ldko:
This is a suggestion to resolve #669 . Since `del` is a statement, not a method/function, the unexplained difference in syntax may be confusing to learners. Using `pop` shows that you can remove a value from a list and make use of the value that was removed.

### DIFF
--- a/_episodes/03-lists.md
+++ b/_episodes/03-lists.md
@@ -243,13 +243,15 @@ odds after adding a value: [1, 3, 5, 7, 11]
 {: .output}
 
 ~~~
-del odds[0]
+removed_element = odds.pop(0)
 print('odds after removing the first element:', odds)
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Thank you very much, @ldko! 🎖

PR #688, opened 2019-07-30
Moved note on importing numpy as np

tretherington:
The note explaining that libraries can be imported with a shorter name seems better placed at the point in the lesson where library imports are introduced.

When reviewing the lesson for the first time in preparation for teaching, when describing how to import I was surprised this wasn't introduced as pretty much every code example I see usual has import numpy as np, so I immediately thought this was a gap in the lesson.  But then I did find this note further down in the lesson which says everything that is needed, but it strikes me as rather tangential in its current position, but I think would be much more relevant in the position I've suggested as at this point import has just been used for the first time.

### DIFF
--- a/_episodes/01-numpy.md
+++ b/_episodes/01-numpy.md
@@ -237,6 +237,18 @@ array([[ 0.,  0.,  1., ...,  3.,  0.,  0.],
 ~~~
 {: .output}
 
+> ## Scientists Dislike Typing
+>
+> We will always use the syntax `import numpy` to import NumPy.
+> However, in order to save typing, it is
+> [often suggested](http://www.scipy.org/getting-started.html#an-example-script)
+> to make a shortcut like so: `import numpy as np`.
+> If you ever see Python code online using a NumPy function with `np`
+> (for example, `np.loadtxt(...)`), it's because they've used this shortcut.
+> When working with other people, it is important to agree on a convention of how common libraries
+> are imported.
+{: .callout}
+
 The expression `numpy.loadtxt(...)` is a [function call]({{ page.root }}/reference/#function-call)
 that asks Python to run the [function]({{ page.root }}/reference/#function) `loadtxt` which
 belongs to the `numpy` library. This [dotted notation]({{ page.root }}/reference/#dotted-notation)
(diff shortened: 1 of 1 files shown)


maxim-belkin:
I agree that positioning of this callout box is not the best. As suggested, however, it would disturb explanation of `numpy.loadtxt`.
I think it would be better to move it to right after "so we only import what we need for each program. " or before "Since we haven’t told it to do anything else".

maxim-belkin:
Thank you, Thomas!

tretherington:
Hmm, I think I agree with you, good point.  I'd suggest right after "so we only import what we need for each program." as at that point the lesson is still talking about importing libraries, whereas after that the lesson has moved on to how to use the library.

maxim-belkin:
> I'd suggest right after "so we only import what we need for each program."

That was my preference too. Moreover, I think it breaks the paragraph in the "right" place.

maxim-belkin:
Thank you for your contribution, Thomas! 👍

maxim-belkin:
A quick follow up: looks like  we have to tweak wording a little bit: `.loadtxt` hasn't been introduced yet, so it might be a bit awkward to talk about NumPy functions here. Perhaps, we should talk about situation when "people refer to NumPy using `np`".

Also, what do you think about moving the word `syntax` in the first sentence to after `import numpy`:

> We will always use the `import numpy` syntax to import NumPy.

tretherington:
Yeah, I think that is partly why I initially put the callout after loadtxt had been introduced, but I think you are right, that this is the best place for it, so better to change the wording.  I can happily have a go at that.

While doing that, I was also wondering: (1) if the [often suggested](https://www.scipy.org/getting-started.html#an-example-script) link is actually very useful, as it just shows an arbitrary scipy script with no further explanation associated with it, so I might be inclined to remove that as I don't see that it adds value and may just confuse, and (2) is it worth extending the callout here to also introduce `from x import y` as that is an increasingly common import form that people will immediately come across when looking for help - or will that just confuse?

maxim-belkin:
+1 for removing the link (had the same concern).
-1 for adding from x import y as z. Introducing more than one piece of
information at once is not good (lesson is already overloaded).

tretherington:
I'd already deleted my branch when you merged, so new pull request [here](https://github.com/swcarpentry/python-novice-inflammation/pull/689). Apologies for any confusion.

ajuavinett:
Hello! Returning to this conversation because I no longer see a note on the lesson about nicknaming packages. Given how common it is to `import numpy as np`, it feels very important to at least mention this in 02-AnalyzingPatientData, or better yet (IMHO), introduce it as the way to import this package.

PR #690, opened 2019-08-14
Fleshed out the schedule a little more to avoid confusion.

nishen:
There was a bit of confusion about the lengths of the coffee breaks and the goind straight from Coffee to Lunch. I've just added in the sections between to increase clarity and give a default schedule for coffee and lunch breaks. 4 x 80 minute segments broken up by coffee and lunch breaks. Obviously just a guideline...

### DIFF
--- a/_includes/sc/schedule.html
+++ b/_includes/sc/schedule.html
@@ -3,10 +3,12 @@
     <h3>Day 1</h3>
     <table class="table table-striped">
       <tr> <td>09:00</td>  <td>Automating tasks with the Unix shell</td> </tr>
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Hi Nishen and thank you for submitting this pull request!

The files that you suggest to change here are part of the workshop template and are present in this repository only because workshop and lesson templates are all based of [carpentries/styles](https://github.com/carpentries/styles) repository. Please submit your ideas against that repository. I would also suggest that you first open an issue to explain the confusion that happened during the workshop so that maintainers have a chance to discuss it and, perhaps, change schedules for other lessons as well.

Thanks anyway!!! 🥇

PR #691, opened 2019-08-14
Remove the word "simple" from 06-func.md

rrasins:
I feel that using the word "simple" can be a bit dismissive. I don't think we can assume that novices would agree that the tests outlined would be simple for them to come up with when they're first starting out with Python. I think this can be reworded to convey a similar meaning without being as dismissive.

### DIFF
--- a/_episodes/06-func.md
+++ b/_episodes/06-func.md
@@ -268,7 +268,7 @@ print(offset_mean(data, 0))
 {: .output}
 
 It's hard to tell from the default output whether the result is correct,
-but there are a few simple tests that will reassure us:
+but there are a few tests that we can run to reassure us:
 
 ~~~
 print('original min, mean, and max are:', numpy.min(data), numpy.mean(data), numpy.max(data))


maxim-belkin:
Thank you, @rrasins, for submitting this pull request! Dismissive language has no place in our lesson so let's find out if we indeed have it here.

> I don't think we can assume that novices would agree that the tests outlined would be simple for them to come up with when they're first starting out with Python

To help us all decide on what to do here, could you please answer the following questions:

- Please describe what these "tests" do.
- Why do you think they're not "simple"?
- How programming (or Python-related) part of this lesson contributes to the complexity?
 
Another question: are these, in fact, "tests"? Or shall we call them "checks"?

Again, thank you for submitting this PR.

ldko:
@maxim-belkin I agree with @rrasins that the use of the word "simple" here can be seen as dismissive and should be removed. I think the `offset_mean` example and the checks are more complicated than necessary to introduce the idea of "testing" functions, and that for many, the tests/checks and explanation are not "simple". I don't teach this section of the episode because it is not made clear enough why one would want to offset the mean (something that may confuse learners) and the explanations of why these checks tell us the functions might be correct can also leave learners confused. The explanation may seem  unconvincing  to some with such language as "seems almost right" and "It’s still possible that our function is wrong".

Regarding your question of whether these are "tests" or "checks", since the section is called "Testing and Documenting" I think one could argue that they are "tests" in that context. I think either "tests" or "checks" could be used though. Whichever of those terms we use, I think "simple" should be removed.

maxim-belkin:
Thanks, Lauren. Since @rrasins hasn't responded to my questions, would you mind doing that?
I'm asking these questions because I'd like to establish a procedure that we would be able to follow in the future when dealing with such issues. I'd like to understand whether there is a misunderstanding of the paragraph(s) in question, how exactly the paragraph/wording is dismissive, and whether it is because we assume a lot of programming knowledge or some other knowledge (and which knowledge, in that case):

1. The first questions inquires the details about the section in question. It will be helpful in situations where there is a misunderstanding of the contents and we need to fix some other text too.
2. The second questions asks about why the submitter thinks this particular part is dismissive. It is important to know the range of things that can be interpreted as dismissive. 
3. The last questions asks whether programming part of the lesson contributes to the situation. Basically, whether this part is dismissive because we assume some knowledge of programming or if we assume some other knowledge. 

Regarding tests/checks: strictly speaking, the part in question deals with "human checking" (though, "human checking" can be called testing). Here is a nice article about it: https://www.satisfice.com/blog/archives/856

ldko:
@maxim-belkin thank you for explaining your reasoning behind the questions.

1.  **Please describe what these "tests" do.**  Calculate the min, max, mean, and standard deviation of the data before and after offsetting the mean and compare that to what the learner's knowledge may or may not expect of the values.
2. **Why do you think they're not "simple"?**   It should not be assumed that someone has a "simple" time with running the code and analyzing the before and after values. Even after the checks, the text explains that the the function may still be wrong.
3. **How programming (or Python-related) part of this lesson contributes to the complexity?** We shouldn't assume any part of programming or Python is "simple" for everyone. For example, some novices may find complexity in putting multiple function calls as arguments in a `print` statement or as in the case of:
```
print('difference in standard deviations before and after:',
      numpy.std(data) - numpy.std(offset_data))
```
including what looks like subtracting two function calls as an argument of the `print` function. It is expected that some complexity will be present for some learners in most any code, and while we don't need to remove all possible complexity, I don't think we should say any of it is "simple".

Regarding this procedure you are trying to establish for qualifying dismissive language, after answering the questions here, I feel like this procedure is more than what should be required.

maxim-belkin:
Thank you, Lauren. Your answers (specifically answers to questions 2 and 3) helped me understand what the issue might be. But before talking about that, here is my take on the first question:

**What these "tests" do**

 - show the minimum, maximum, and average values of data stored in a variable `data`.
 - apply function `offset_mean` to `data` and store it in a new variable called `offset_data`
 - show the minimum, maximum, and average values of data stored in a variable `offset_data`
 - the **human checking** part -- which is the actual "test" here -- happens after the last command when the learner compares its output with what they might expect to see based on the output of the first command.

**What knowledge of Python these steps require**

- use of `print()` function with multiple arguments
- how to execute a function
- how to assign a value to a variable

At the point in the lesson, all of these steps have been discussed and applied several times.

---

Now, going back to your answers and the original question.

Your answers hinted me that you interpret "simple" as an attribute of actions expected from learner. And I disagree with that. Here, "simple" is a qualifications of the testing procedure that we do: look at 3 numbers (min, mean, max) related to the data before and after processing it with a function. Note, in the context of the episode we're using this testing procedure instead of checking each value in the array to understand whether the function works correctly.

I, of course, agree with you in that evaluating obtained output might not be easy. I also agree that wrapping your head around using functions inside of functions might not be easy for new programmers. But these challenges stand separately from the employed testing procedure.

Regarding "seems almost right" and "It’s still possible that our function is wrong": these phrases make sense if we attribute "simple" to the testing procedure and not to how difficult it is for learners to execute individual steps. If we remove/change "simple", we have to update the wording of these sentences.

In summary, I suggest we do the following

1. As a stopgap measure: remove "simple" as suggested here or replace it with an antonym for "rigorous" that would be more difficult to associate with an individual.
2. (In a separate PR) Improve the paragraph. Options:
  - explicitly state the procedure before showing the commands.
  - break down the "tests" into individual commands and discuss each of them.

Regarding the three questions: we can tune them (feel free to make suggestions). But, honestly, I don't know what I would drop from them because I needed all of your answers to connect you point of view with mine.

ldko:
I don't think our point of views are connected despite the questions. I don't think removing the word "simple" in itself requires any other changes to the text. Any opinions on this all @annefou?

annefou:
Just to be honest with both of you @ldko and @maxim-belkin I think the discussion is not worth. I do not have any personal opinion on wether this test is "simple" or not but the re-wording does not change the meaning of the sentence (probably my english is not good enough to spot any differences!).  So I would not have any problem to accept this change.

maxim-belkin:
Thanks, Anne.

Lauren: please feel free to merge this PR as it stands. I don't approve this change in its current form for the reasons I explained above (+ I think these reports indicate other problems with the text in the lesson) but I accept your and Anne's decision.

PR #693, opened 2019-08-16
Update 08-defensive.md

wachiuphd:
As written, the assertion that checks that each value can be converted to an integer will fail if one of the elements is 0, since int(0) is 0 which is treated as False.  Zero is a legitimate value for a number of cars.

### DIFF
--- a/_episodes/08-defensive.md
+++ b/_episodes/08-defensive.md
@@ -520,7 +520,7 @@ This violates another important rule of programming:
 > def get_total(values):
 >     assert len(values) > 0
 >     for element in values:
->         assert int(element)
+>         assert isinstance(element, int)
 >     values = [int(element) for element in values]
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Thank you, @wachiuphd, for pointing out the problem!

`assert int(0)` indeed fails because `0` is treated as `False`. There is another problem with `assert int(element)`too: `int('123')` is `True` whereas `int('0')` is still `False`.

However, we don't talk about `isinstance` in our lesson at the present time and introducing it in a challenge without any explanation doesn't seem like the right thing to do.

How about changing the assert statement to the following?
```py
assert int(element) or element == 0
```

knarrff:
Can we have someone take a look at this please? The current proposal is a lot better (introducing is_instance) than the current state (a bug in code that should be ok).

PR #695, opened 2019-08-20
Update 01-numpy.md: clarifying text for axis=0,1 when averaging

davidbenncsiro:
I've suggested some clarifying text (see diffs) for axis=0 and axis=1 when computing average.

### DIFF
--- a/_episodes/01-numpy.md
+++ b/_episodes/01-numpy.md
@@ -668,7 +668,7 @@ operation across an axis:
 
 To support this functionality,
 most array functions allow us to specify the axis we want to work on.
-If we ask for the average across axis 0 (columns in our 2D example),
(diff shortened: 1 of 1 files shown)


davidbenncsiro:
Wait! I think we do indeed mean **rows** here after all.

I read this wrong but some clarification may be useful.

Can I suggest a clarifying addition? Instead of:
~~~
If we ask for the average across axis 0 (rows in our 2D example)
~~~
something like this:
~~~
If we ask for the average across axis 0 (down each column for all rows in that column, as shown by the arrows in the figure for our 2D example)
~~~

I've made a similar suggestion for axis=1.

I'm not sure whether I've clarified or confused in the end, so it's okay if you don't want to accept my suggestion of course.

davidbenncsiro:
No worries. That sounds quite reasonable @maxim-belkin 

Thanks.

PR #696, opened 2019-08-23
Update 03-lists.md

davidbenncsiro:
0 minutes for exercises seems a bit optimistic. :)

I'm not sure of the right number here but perhaps 10 mins?

I've just noticed that most episodes except for the first only show teaching time. Is the assumption that exercises are included in that time?

### DIFF
--- a/_episodes/03-lists.md
+++ b/_episodes/03-lists.md
@@ -1,7 +1,7 @@
 ---
 title: Storing Multiple Values in Lists
 teaching: 30
-exercises: 0
+exercises: 10
 questions:
 - "How can I store many values together?"
 objectives:
--- a/_episodes/03-lists.md
+++ b/_episodes/03-lists.md
@@ -1,7 +1,7 @@
 ---
 title: Storing Multiple Values in Lists
 teaching: 30
-exercises: 10
+exercises: 15
 questions:
 - "How can I store many values together?"
 objectives:


maxim-belkin:
> 0 minutes for exercises seems a bit optimistic. :)
>...
> I've just noticed that most episodes except for the first only show teaching time. Is the assumption that exercises are included in that time?

"0 minutes for exercises" is our legacy. We do need to update all episode to list more or less accurate time for exercises.

10 minutes for 4 exercises is definitely more realistic though is still optimistic (at least when I teach I spend more than 10 minutes). I think 3-5 mins per exercise, 12-20 min total, is more realistic. Perhaps, 15 minutes?

annefou:
I agree 15-20 mn is indeed more realistic.

maxim-belkin:
What do you think, @davidbenncsiro? Would you mind updating the estimate to 15 or 20 minutes?

davidbenncsiro:
Thanks @annefou and @maxim-belkin, I've updated it to 15 mins.

I suppose at some point, the same question can be answered for the other episodes. Shall I create a PR for those as well?

maxim-belkin:
> Thanks @annefou and @maxim-belkin, I've updated it to 15 mins.

Thank you, David!

> I suppose at some point, the same question can be answered for the other episodes. Shall I create a PR for those as well?

Yes, please! And thanks again!

PR #697, opened 2019-08-25
Update 06-func.md: exercise time

davidbenncsiro:
Following on from the PR #691, the exercises for this episode need around 15 to 20 mins as well I think. The first 3 are simpler than those that follow, so around 2 mins vs 3 mins or so for the two groups. What's your experience been @maxim-belkin and @annefou ? I'm continuing to prepare for my first Python swc teaching session after checkouts here in Adelaide (AU) this week.

### DIFF
--- a/_episodes/06-func.md
+++ b/_episodes/06-func.md
@@ -1,7 +1,7 @@
 ---
 title: Creating Functions
 teaching: 30
-exercises: 0
+exercises: 15
 questions:
 - "How can I define new functions?"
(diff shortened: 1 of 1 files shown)


annefou:
@davidbenncsiro nice inputs (same applies with all your PRs where you add timing for exercises). 
Unfortunately, your changes are conflicting with the current version. Would you be able to rebase?
Let us know if you need help.
Thanks for your contributions!

PR #703, opened 2019-08-26
Update 08-defensive.md

davidbenncsiro:
The line to be corrected from:
~~~
scaled = float(dx) / dy
~~~
to:
~~~
scaled = float(dy) / dx
~~~
is on line number 14, not 11, if the `def` is on line 1.


maxim-belkin:
Thanks, @davidbenncsiro!

PR #704, opened 2019-08-26
08-defensive.md: fix styling of keyboard shortcuts

maxim-belkin:
According to the style guide we have to use `<kbd></kbd>`...


maxim-belkin:
Self-merging....

PR #705, opened 2019-08-26
Update 08-defensive.md: exercise time

davidbenncsiro:
The time required for exercises in this episode is I think between 5 and 10 minutes. Of course, times required for all episodes' exercises really depends upon the experience of participants. I'm sure I'm not saying anything you don't know. :)

### DIFF
--- a/_episodes/08-defensive.md
+++ b/_episodes/08-defensive.md
@@ -1,7 +1,7 @@
 ---
 title: Defensive Programming
 teaching: 30
-exercises: 0
+exercises: 10
 questions:
 - "How can I make my programs more reliable?"
 objectives:


maxim-belkin:
For three exercises 10 minutes is not realistic. However, I'm tempted to remove the last exercise because:
a) its wording is not clear: "Fix" may have many different things
b) we haven't introduced `min` and `max` functions (which we might have to "fix")

Therefore, I suggest we:
1) Merge this PR
2) Move the last exercise to `_extras/extra_exercises.md`

After that, we should discuss how we can introduce `min` and `max` functions.

davidbenncsiro:
@maxim-belkin, thanks. Do we need to open a new issue for your idea above of moving the last exercise to _extras/extra_exercises.md and discussion of how we can introduce min and max functions?

maxim-belkin:
I moved that exercise to "extras" in dc112ff7.
We need to discuss the second part of the suggestion in a separate issue (or a PR, if you've got some ideas already).

PR #706, opened 2019-08-26
09-debugging.md: number of possible interactions between bugs in code

davidbenncsiro:
I may misunderstand the intent here, but if by "can interact in..." in "N things can interact in N<sup>2</sup> different ways", we mean something like "can be arranged in...", i.e. permutations, then should we say "N!" here instead (e.g. A, B, C => ABC, ACB, BAC, BCA, CAB, CBA, i.e. 3! ways)?

For example, if we have 3 monadic functions f, g, and h that can be freely composed, then we could have 3! ways of composing them without repetition: f(g(h)), f(h(g)), g(f(h)), g(h(f)), h(f(g)), h(g(f)).

### DIFF
--- a/_episodes/09-debugging.md
+++ b/_episodes/09-debugging.md
@@ -162,7 +162,7 @@ however small,
 we should re-run our tests immediately,
 because the more things we change at once,
 the harder it is to know what's responsible for what
-(those N<sup>2</sup> interactions again).
+(those N! interactions again).
 And we should re-run *all* of our tests:
 more than half of fixes made to code introduce (or re-introduce) bugs,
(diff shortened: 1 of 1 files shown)


maxim-belkin:
I think you're right, @davidbenncsiro! I think the N<sup>2</sup> is an estimate of the number of possible interactions between 2 things each of which has N possible values. Because we're talking about possible typos/errors/etc. on different lines of code, `N!` seems like a more reasonable estimate... So, I approve.

Just for reference, related commits are: f31f74b4 and 39b5b9

davidbenncsiro:
Thanks @maxim-belkin !

PR #708, opened 2019-08-26
Update 09-debugging.md: exercise time

davidbenncsiro:
5 or 10 minutes to complete exercises seems reasonable for this episode. The first exercise could take awhile so perhaps 10 minutes is closer to the mark.

### DIFF
--- a/_episodes/09-debugging.md
+++ b/_episodes/09-debugging.md
@@ -1,7 +1,7 @@
 ---
 title: Debugging
 teaching: 30
-exercises: 0
+exercises: 10
 questions:
 - "How can I debug my program?"
 objectives:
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Any interaction between learners takes time. I think 20 minutes would be a more realistic estimate for these two exercises.

annefou:
I definitely agree with you @maxim-belkin 
I never managed to spend less than 20 mn on exercises!

maxim-belkin:
Thank you, Anne.
David, could you please change exercise time to 20 minutes?

maxim-belkin:
David, I went ahead and changed the time estimate to 20 minutes. I hope you don't mind.

davidbenncsiro:
Thanks @maxim-belkin and apologies for the delay in getting to this. I've been a bit busy over the last several days. Happy for you to have made this change. There are other outstanding PRs that I will get too soon. Thanks again!

PR #710, opened 2019-08-26
04-files.md: reword 'Plotting Differences' callout box

maxim-belkin:
It all started with a typo fix (leftmost plot*s*)...


maxim-belkin:
Self-merging...

PR #711, opened 2019-08-26
setup.md: Return not Enter

maxim-belkin:
According to [this table in the lesson example](https://carpentries.github.io/lesson-example/06-style-guide/index.html#keyboard-key)
we ought to use `Return` instead of `Enter`.


maxim-belkin:
Self-merging...

PR #712, opened 2019-08-27
Update 06-func.md: visualize() vs analyze()

davidbenncsiro:
The function on line 152 could arguably be called visualize(). If such a change is made, other references to analyze() would need to be changed. Just a thought.

### DIFF
--- a/_episodes/06-func.md
+++ b/_episodes/06-func.md
@@ -149,7 +149,7 @@ we can make our inflammation analysis easier to read and easier to reuse.
 First, let's make an `analyze` function that generates our plots:
 
 ~~~
-def analyze(filename):
+def visualise(filename):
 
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Thank you, David. I agree with this suggestion. In the lesson we say:

> First, let's make an `analyze` function **that generates our plots**

calling it `analyze` is OK (we "analyze" data by looking at it), but not ideal (the function does visualization, not analysis). So, let's make this change -- please change all other occurrences of `analyze` to `visualize` (there are 4 places where we mention `analyze` function, all in `06-func.md`).

davidbenncsiro:
I've committed the 4 changes @maxim-belkin

maxim-belkin:
Excellent! Thank you, David!

PR #713, opened 2019-09-09
etherpad and google docs exporters

Denubis:
This is related to my code in https://github.com/carpentries/instructor-training/pull/1010, but added a themed one for google docs exports because I need it for tomorrow. 

Pinging @maxim-belkin @fmichonneau for the conversation here so that we're not tangling up the instructor-training exporter with stuff that better belongs in lesson templates.


maxim-belkin:
Thank you for the PR, @Denubis. Please submit it to [carpentries/styles](https://github.com/carpentries/styles) repo. Once we polish it there, all lessons will receive it as part of the style update procedure.

Thanks again!!!

PR #714, opened 2019-09-10
08-defensive.md: move the last exercise to 'Extras'

maxim-belkin:
Related to: #708 

The last exercise has a few issues:

a. it uses unclear language ("Fix" may refer to many things)
b. it uses functions not (currently) introduced in the lesson -- builtin `max` and `min`

Both issues can be fixed. Until then, we're moving this exercise to "Extras"

### DIFF
--- a/_episodes/08-defensive.md
+++ b/_episodes/08-defensive.md
@@ -538,27 +538,4 @@ This violates another important rule of programming:
 > {: .solution}
 {: .challenge}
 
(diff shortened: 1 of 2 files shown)


annefou:
Yes, I fully agree to move this exercise to 'extras'. To be honest, I have never done this exercise during a python workshop. Thanks!

maxim-belkin:
Thanks, Anne!

PR #715, opened 2019-09-13
Update 03-lists.md

ConradGGit:
Changed instructions on copying lists.  Instead of using 
 ```python
my_salsa = list(salsa)
```
changed to 
```python
my_salsa = copy.deepcopy(salsa)
```

This is because using `list()` is less robust.  The goal was to generate a new list which could be modified, without changing the original list.

However, `list()` will not achieve this if the original list contains a nested list.  Whereas `deepcopy()` will work for any number of nested lists.

The following code illustrates this difference.

```python
TestList_1 = [1,"apple", ["NestedListItem1","NestedListItem2"]]
print('The original list:', TestList_1)
print('Now we use the list() function to create a new list, and change a non-nested element of the list.')
NewList_1 = list(TestList_1)
NewList_1[0] = 2
print('The new list:',NewList_1)
print('In this case the original list remains unchanged.')
print('The original list:',TestList_1)
print('However, now we change a nested element of the new list.')
NewList_1[2][0]= "ChangedNestedListItem1"
print('The new list:',NewList_1)
print('In this case the old list gets changed.')
print('The old list:',TestList_1)

import copy
print('This time do the same thing, but use copy.deepcopy() instead of list().')
TestList_1 = [1,"apple", ["NestedListItem1","NestedListItem2"]]
print('The original list:', TestList_1)
print('Now we use the copy.deepcopy() function to create a new list, and change a non-nested element of the list.')
NewList_1 = copy.deepcopy(TestList_1)
NewList_1[0] = 2
print('The new list:',NewList_1)
print('In this case the original list remains unchanged.')
print('The original list:',TestList_1)
print('However, now we change a nested element of the new list.')
NewList_1[2][0]= "ChangedNestedListItem1"
print('The new list:',NewList_1)
print('In this case the old list also remains unchanged.')
print('The old list:',TestList_1)
```

### DIFF
--- a/_episodes/03-lists.md
+++ b/_episodes/03-lists.md
@@ -149,7 +149,8 @@ does not.
 >
 > ~~~
 > salsa = ['peppers', 'onions', 'cilantro', 'tomatoes']
-> my_salsa = list(salsa)        # <-- makes a *copy* of the list
+> import copy
+> my_salsa = copy.deepcopy(salsa)        # <-- makes a *copy* of the list
 > salsa[0] = 'hot peppers'
 > print('Ingredients in my salsa:', my_salsa)
 > ~~~


maxim-belkin:
Thank you for the PR, @ConradGGit.

Yes, `copy.deepcopy()` is definitely more "robust" that `list()` when it comes to copying lists. However, this is definitely a one step further that we want to talk about in this introduction to Python. We can add a note to the instructor guide (to help instructors prepare for such questions), but I don't think we have the time in the main lesson to talk about `copy.deepcopy`.

@annefou, what do you think?

annefou:
your suggestion @ConradGGit is great but too complex for an introduction to python. So I agree with you @maxim-belkin. Adding some information in the instructor notes is a very good idea.

ConradGGit:
Thanks for your feedback Anne and Maxim I really appreciate it and I am more than happy to put this in the instructor notes as you suggest.

Just so I understand though, would you mind letting me know why ```copy.deepcopy()``` is considered complicated?

As far as I can tell, the only additional complication compared to ```list()```  is that we would need to do ```import copy```.  But we have already covered importing libraries several times by this point in the course so I would have thought all students would be comfortable with this.  I guess it raises the issue of nested lists, but again this doesn't seem that hard to explain.

As I say, I am happy to go along with the consensus and put this in the notes, but I just thought I would ask the question so I can understand for future reference :)

maxim-belkin:
It's not that `copy.deepcopy` is complicated, it's the sheer amount of information that learners have to remember. At the moment, we can't (responsibly) add one more layer. You can argue that `list()` is also "a layer of information" and... I would have difficult time arguing with it because it certainly is...

I'm marking this PR as "discussion" to get opinions from the community.

ldko:
I see why it might be desirable to not teach `copy.deepcopy` with the quantity of material being taught already. However, it seems to me that many of the learners won't remember how they need to go about copying a `list` (whether by `list()` or `copy.deepcopy`) or doing most things they are seeing only once in a two-day workshop and would need to look back at the lesson or elsewhere for the function/syntax in the future anyway. If it were me learning, I would probably remember that there was something weird that I need to do to copy a list, but not necessarily how to do it. In such a case, I would find it most helpful to have included in the material the best practice/way that always works for copying the list. Also, I don't think it is a bad thing to remind learners about how they frequently may want to `import` functionality from various modules to do one thing or another.

Since the line 'my_salsa = list(salsa)' is in between a discussion of mutability and nested lists. To me telling the the learner how to make a copy of a `list` then immediately teach them a `list` it won't work on is like giving the learner a flashlight with no batteries and sending them into a dark room.

maxim-belkin:
Thanks, Lauren.

cmacmackin:
I'm incline to agree that it is better to use `copy.deepcopy` as it is the best practice. Given that `list` is not used anywhere else and learners are not really introduced to the concept of constructors, that isn't really any easier or more meaningful to them than `copy.deepcopy`. If anything I'd say it's less so, as it is fairly clear what something with the word "copy" in its name is going to do. However, to the extent that they would understand the purpose of the list constructor it just begs the question of why they don't need that when creating a list in the first place. The whole issue of constructors or converting iterables to lists would be well beyond the scope of the lesson, so best to avoid it entirely. Using `copy.deepcopy` has the further benefit that instructors could quickly mention it can be used to copy _anything_, not just lists.

PR #724, opened 2019-10-01
Fix capitalization

matuskalas:
Jupyter Notebook with capital N
Anaconda distribution with small d (not in title)
"plain vanilla" in small also in option


maxim-belkin:
Thank you, Matúš!

PR #726, opened 2019-10-11
Update 02-loop.md

kvndrs:
Carpentries instructor checkout lesson correction submission:
In the episode entitled, 'Repeating Actions with Loops'  of the lesson,  'Programming With Python', I clarified the meaning by indicating the name of the first episode. Also, I indicated that we could call up the dozen data sets previously created using the following commands, without stating the commands specifically.

---

---

### DIFF
--- a/_episodes/02-loop.md
+++ b/_episodes/02-loop.md
@@ -15,12 +15,12 @@ keypoints:
 - "Use `len(thing)` to determine the length of something that contains other values."
 ---
 
-In the last episode, we wrote Python code that plots values of interest from our first 
+As you recall from the first episode, 'Analyzing Patient Data', we wrote Python code that plots values of interest from our first 
 inflammation dataset (`inflammation-01.csv`), which revealed some suspicious features in it.
 
 ![Analysis of inflammation-01.csv](../fig/03-loop_2_0.png)
 
-We have a dozen data sets right now, though, and more on the way.
+We have a dozen data sets right now, though, and more on the way.  We could call up these data sets using the following commands. [Insert commands]. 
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Thank you, @kvndrs, for the pull request and welcome to The Carpentries.
Below I'll review this pull request. Keep in mind, however, that this review takes into account the current state of the lesson (the first episode is very long) and it (the review) will not affect your instructor checkout status. So, let's go.

When we switch to the second episode, we want to switch to the new topic by reminding people what we've done a few "moments" ago. The name of the episode is an additional piece of information which our learners, strictly speaking, don't need to remember. I, therefore, think we should keep the current wording and not mention the title of the previous episode. Alternatively, we can change "last" to "previous" or "first", or change the introductory part completely to something like "So far, we've written Python code that..."

Regarding the second part of the changes. You add the sentence 

>  We could call up these data sets using the following commands. [Insert commands]. 

Here you're using jargon ("call up") which, frankly speaking, I'm not aware of myself and, therefore, would argue that we don't have the time to introduce it in our lesson. If you're suggesting to convert this part into an exercise, then I would argue that this is too early into the episode to have any exercises.

So, my current verdict would be to either pass on this completely or make a minor change I mention above. Though, we would have to discuss it more and see if it promises any improvements to the episode/lesson.

Thanks for the pull request anyway! 🎉

ldko:
I'm closing this based on Maxim's comments and the shifting of the episodes that has since occurred. @kvndrs we appreciate you taking the time to make the suggestions and open the pull request.

PR #728, opened 2019-10-22
Added exercise on using isinstance()

Vilma-Agalioti:
I have added example and an exercise on defensive programming with the use of isinstance()
This adds with a 'how to' idea on defensive programming in the context of other types of objects, such as string and bytes, and not only numeric.

### DIFF
--- a/_episodes/08-defensive.md
+++ b/_episodes/08-defensive.md
@@ -561,4 +561,42 @@ This violates another important rule of programming:
 > {: .solution}
 {: .challenge}
 
+
+> ## Testing data type within a function with `isinstance()`
+>
+> `isinstance()` examines the type of an object.
+> We can use `isinstance()` to test if the object that is entered as an argument
+> in a function is allowed for this specific function.
+> For example, testing if 3.14 is float:
+> ~~~
+> isinstance(3.14, float)
+> ~~~
+> {: .language-python}
+>
+> ~~~
+> True
+> ~~~
+> {: .output}
+>
+> Write a function that prints a word given as an argument in a string format.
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Thank you, Vilma, for your contribution! And welcome to The Carpentries.

Unfortunately, I have to vote against merging in changes suggested in this PR for two reasons.

1.  In general, `isinstance` should be used to check if a particular variable is of one of several specified types (`isinstance(variable, (int, float, str))`) and in cases where we use classes which inherit from base classes (`class my_str(str): pass; isinstance(var, my_str);`). When we need a strict check that a specific variable is of a specific type (`float` or `int`), we still have to use `type(3.14) == float` or `type(3) == int`.

2. We can not add content/information without making changes to the main story line. We have to explain the need for this `isinstance()` in the text before we add the information about it.

My apologies, Vilma, if my review seems too harsh as this is your first contribution to our lesson.

Thanks for your review, Lauren. Looks like we need an opinion of our third maintainer to break the tie

CC @annefou

annefou:
Thanks @Vilma-Agalioti  for for contribution, 
The new exercise does not fit the existing objectives (as written at the top of the episode). It does not mean that it would not be useful to discuss `isinstance` as part of this episode but as is, it only adds new factual information. In that sense, I agree with @maxim-belkin : if we add `isinstance` in the episode, we need to motivate its usage (why and when do we need to check for type, etc.) and add prior information before introducing an exercise on this topic.

ldko:
@Vilma-Agalioti , based on the majority of maintainers deciding against introducing `isinstance()` at this time, I am going to close this pull request. We do appreciate you offering this contribution, thank you!

PR #729, opened 2019-10-23
Update some docstring sections to use numpy standard.

Carreau:
You will also see that some of the information was incorrect due to
multiple update. The function used to be called `center` and to return
a list instead of a numpy array. So the help and example were incorrect.

The use of the numpy standard is a good practice; as most projects use
it. Even if we don't expect users to remember it; at least they are made
aware that there is standard and can look for it.

This also mirror the documentation style of loadtxt that is seen later
in the lesson.

And finally sphinx can understand numpydoc and generate nice HTML pages.
So all in all I believe this is a nice detail to add to the lesson.

Please delete this line and the text below before submitting your contribution

### DIFF
--- a/_episodes/06-func.md
+++ b/_episodes/06-func.md
@@ -375,18 +375,6 @@ def offset_mean(data, target_mean_value):
     '''Return a new array containing the original data
        with its mean offset to match the desired value.
 
-    Parameters
-    ----------
-    data : numpy array, or list
-        The data we want to correct the average
-    target_mean_value: float or int
-        The target mean value we want for our new array
-
-    Returns
-    -------
-    numpy array: 
-        Same dimension as `data`, with mean value `target_mean_value`
-
     Examples
     --------
     >>> offset_mean([1, 2, 3], 0)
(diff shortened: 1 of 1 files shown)


_episodes/06-func.md
```diff
@@ -374,19 +374,58 @@ we can break the string across multiple lines:
...
 {: .language-python}
 
+We're also the [numpy docstring
+standard](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard),
+which is common across the scientific Python ecosystem and understood by many
+tools. 
+
 ~~~
-Help on function center in module __main__:
+Help on function offset_mean in module __main__:
```
ldko: This is a change we should include. Would you be willing to submit this naming correction in a separate PR?

maxim-belkin:
Yes, I think extras is the right place at the moment. At least for now. If / when we overhaul this part of the lesson, we might include it into the main story line. But right now -- it is kind of TMI :)

👋 Matthias! Glad to see you here!

annefou:
I agree with you @ldko and @maxim-belkin . Adding this additional step in extras is best. It is good to discuss best software practices but will be probably a bit too overwhelming for complete novice. 
Thanks.

Carreau:
> This is a change we should include. Would you be willing to submit this naming correction in a separate PR?

Done as #755 

>  Adding this additional step in extras is best. It is good to discuss best software practices but will be probably a bit too overwhelming for complete novice.
Thanks.

Fair point, though I will counter with the fact that the current docstring is counter productive. 
```
Return a new array containing the original data
with its mean offset to match the desired value.
Example: offset_mean([1, 2, 3], 0) => [-1, 0, 1]
```

It give a bad example to learners, introduce non standard syntax `=>` and is technically incorrect (the return value is not [-1, 0,1] but an array of float); which I believe we should avoid, can we go half way with something like:

```
    '''Return a new array containing the original data
       with its mean offset to match the desired value.
  
    Examples
    --------
    >>> offset_mean([1, 2, 3], 0)
    array([-1.,  0.,  1.])
    '''
```

It  does not add more information, is correct (the result is an array while currently is not) and give a chance to talk about the format.

annefou:
Your new proposal makes sense to me and as it does not add much more typing (which is usually a concern for novices).

Carreau:
> Your new proposal makes sense to me and as it does not add much more typing (which is usually a concern for novices).

Ok, updated to only reformat the Example section.

PR #733, opened 2019-11-03
01-numpy.md: Remove paragraph about '%matplotlib inline' magic

Iain-S:
As per issue 716, remove reference to `%matplotlib inline` as it is no longer required.


maxim-belkin:
Thank you, @Iain-S! I approve!

maxim-belkin:
Thank you, @Iain-S!

Issue #725, opened 2019-10-10
Practice what you preach: Variable names


SobhanOmranian:
In the lesson [Repeating Actions with Loops](http://swcarpentry.github.io/python-novice-inflammation/02-loop/index.html), we have a section named "What’s in a name?", in which it is stated that although the loop variable can be arbitrarily named, it is a good practice to give it a meaningful name in order to make its purpose more clear.

This is indeed an important, yet often missed point that can pay dividends in the long term from the readability aspect of the code. As most of the audience of this lesson are novices with fresh minds and ears, it is as important that the material consistently adheres to what it advocates. 

However, in numerous episodes throughout the lesson, we do not see proper naming for the (loop) variable names. For example, in the episode [Analyzing Data from Multiple Files](http://swcarpentry.github.io/python-novice-inflammation/04-files/index.html), there is the following piece of code (condensed for brevity):
```
filenames = sorted(glob.glob('inflammation*.csv'))
filenames = filenames[0:3]
for f in filenames:
    print(f)
    data = numpy.loadtxt(fname=f, delimiter=',')
```
As we can see, although the variable `filenames` is named carefully, the loop variable has lazily been abbreviated from `file` or `filename` to `f`.  While this might easily be okay in a more advanced course since the learners are more familiar with the concept of variables and naming conventions, it is in direct contradiction with the aforementioned good practice. A learner might wonder why such practice is not followed in the course. 

In fact, I don't seem to be able to find a good reason behind abbreviating such variable names. If saving time or space is the underlying reason, then typing `f` instead of `filename` hardly saves any of that. However, having a meaningful names such as `filename` or `file` does make the code more readable and is more aligned with the best practices promoted earlier. I hope the lesson maintainers could shed some light on the reasoning behind choosing such variable names.

This issue is propagated to various parts of the lesson. For instance another example would be in [Creating Functions](http://swcarpentry.github.io/python-novice-inflammation/06-func/index.html):
```
filenames = sorted(glob.glob('inflammation*.csv'))
for f in filenames[:3]:
    print(f)
    visualize(f)
    detect_problems(f)
```

I can provide a list of all the instances, if the consensus is that we are in need of more descriptive variable names in such cases.

ldko:
[...]

PR #734, opened 2019-11-04
Loop variable names

sparce:
Changes loop variable names to be more meaningful per issue #725 

Think I caught them all except for the one in the [functions episode](http://swcarpentry.github.io/python-novice-inflammation/06-func/index.html#readable-functions) that is supposed to be hard to interpret. 

(First time attempting a PR for the lessons, so feel free to let me know if there's anything I should be doing differently for next time)


sparce:
I should also add that I noticed some inconsistencies between the code in the lesson notes for [command line programs](http://swcarpentry.github.io/python-novice-inflammation/10-cmdline/index.html) and the actual scripts in the `code` directory.

For example, the lesson notes show the `readings_*.py` files using `numpy.mean(data, axis=1)`, while the scripts themselves use `data.mean(axis=1)`. 
Some of the solutions in the notes (eg. (changing flags)[http://swcarpentry.github.io/python-novice-inflammation/10-cmdline/index.html#changing-flags]) also just call `main()` directly at the end of the script, whereas all the script files check `if __name__ == '__main__':` first.

These differences don't matter for how the scripts run, but it's probably worth keeping the notes consistent with the scripts that learners will be downloading. Would it be better to start a separate issue for this, or should I just add corrections to this PR since I'm already working with the files in the code directory?

ldko:
@sparce thanks for noting the inconsistencies between the lessons and `code` directory. I agree with you that it is best to keep the sample code and instruction consistent. Changes for that should go wit a separate issue and PR. Thank you!

_episodes/04-files.md
```diff
@@ -162,11 +162,11 @@ where the maxima are a bit less regular, but the minima are consistently zero.
...
 > > composite_data = numpy.zeros((60,40))
 > >
-> > for f in filenames:
-> >     data = numpy.loadtxt(fname = f, delimiter=',')
+> > for filename in filenames:
+> >     data = numpy.loadtxt(fname = filename, delimiter=',')
 > >     composite_data += data
 > >
-> > composite_data/=len(filenames)
+> > composite_data /= len(filenames)
```
maxim-belkin: @ldko and @annefou, we need to check that we mention compound operators in the introduction

maxim-belkin:
Thank you, Stephen.
Are you comfortable splitting the PR into several smaller ones, updating one file per PR?

_episodes/04-files.md
annefou: We do not mention compound operators in the introduction. But we do it here. To be honest, I sometimes do not mention it and use
composite_data = composite_data + data
etc.
Maybe we should add a note in the instructor guide; what do you think?

sparce:
@maxim-belkin Is there a way to break it up from here, or would I need to start again with a separate branch and PR for each episode? I'm happy to handle it either way.

maxim-belkin:
I see that you're adept at using GitHub, so may I ask you to follow the guidelines in our Wiki: https://github.com/swcarpentry/python-novice-inflammation/wiki/Splitting-(Breaking-down)-Pull-Requests and let me know if the instructions there work for you?

_episodes/04-files.md
maxim-belkin: @nspies, please ignore this discussion when doing the PR split -- we'll address this in a separate PR. You changes are good.

maxim-belkin:
Thank you for splitting this PR into multiple ones, Stephen!
Working with small "limited-in-scope" commits is a lot easier and usually results in faster turn-around times.

PR #744, opened 2019-11-10
Update loop variable names in code files

sparce:
Split from PR #734

### DIFF
--- a/code/readings_01.py
+++ b/code/readings_01.py
@@ -6,5 +6,5 @@ def main():
     script = sys.argv[0]
     filename = sys.argv[1]
     data = numpy.loadtxt(filename, delimiter=',')
-    for row_mean in data.mean(axis=1):
+    for row_mean in numpy.mean(data, axis=1):
         print(row_mean)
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Could you please synchronize this file with what we have in the episode? Here is what we've got there:

```py
import sys                                                                                               
import numpy                                                                                             
                                                                                                         
                                                                                                         
def main():                                                                                              
    script = sys.argv[0]                                                                                 
    filename = sys.argv[1]                                                                               
    data = numpy.loadtxt(filename, delimiter=',')                                                        
    for row_mean in numpy.mean(data, axis=1):                                                            
        print(row_mean)  
```

sparce:
Can do. I misinterpreted a comment on the last PR and thought that you wanted those changes separately. :)

maxim-belkin:
Thank you, Stephen! I think it's best to fix all discrepancies in these scripts at once as the entire procedure can be dubbed "Syncing up with the lesson". I hope Lauren will forgive me for that.

PR #745, opened 2019-11-10
Update loop variable names in code files

sparce:
Split from PR #734


maxim-belkin:
Could you please synchronize the contents of this file with what we have in the episode?
There should be: (a) only one empty line between imports and `def main` and `numpy.mean(data...` instead of `data.mean`.

maxim-belkin:
Thank you, Stephen!

PR #747, opened 2019-11-10
Update loop variable names for code file

sparce:
Split from PR #734

### DIFF
--- a/code/readings_04.py
+++ b/code/readings_04.py
@@ -11,11 +11,11 @@ def main():
         data = numpy.loadtxt(filename, delimiter=',')
 
         if action == '--min':
-            values = data.min(axis=1)
+            values = numpy.min(data, axis=1)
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Could you please sync it with what we have in the episode?
Here is what we've got there (note `numpy.***` vs `data.***`):

```py
import sys
import numpy

def main():
    script = sys.argv[0]
    action = sys.argv[1]
    filenames = sys.argv[2:]

    for filename in filenames:
        data = numpy.loadtxt(filename, delimiter=',')

        if action == '--min':
            values = numpy.min(data, axis=1)
        elif action == '--mean':
            values = numpy.mean(data, axis=1)
        elif action == '--max':
            values = numpy.max(data, axis=1)

        for val in values:
            print(val)

if __name__ == '__main__':
   main()
```

PR #750, opened 2019-11-10
Update loop variable names in code file

sparce:
Split from PR #734


maxim-belkin:
And again, could you please sync this with what we have in the episode? The contents of this file in the lesson is shown in the "Changing Flags" challenge. There we use `numpy.***` functions rather than `data.***` methods.

PR #751, opened 2019-11-10
Update loop variable names

sparce:
Split from PR #734


maxim-belkin:
Could you please sync the content of the file with what we have in the episode? (`numpy.***` instead of `data.***`, etc)

PR #757, opened 2019-11-18
Fix 'Computing the Value of a Polynomial' exercise

marijane:
The Output block in this exercise contains two assignments to the variable y, which neither looks like an Output nor does it follow from the preceding Python block when run.

I believe the first assignment to y was intended to be part of the Python block, and I also believe the intent was to print the value of y, which is 97, instead of assigning 97 to it in the Output block.

This commit does the following:
- moves the first y assignment from the Output block into the Python block
- adds a print(y) statement to the end of the Python block
- removes the assignment from the y = 97 line, so that the Output block matches the output of the edited Python block when it is run.

---

---

### DIFF
--- a/_episodes/03-loop.md
+++ b/_episodes/03-loop.md
@@ -390,12 +390,13 @@ so we should always use it when we can.
 > ~~~
 > x = 5
 > coefs = [2, 4, 3]
+> y = coefs[0] * x**0 + coefs[1] * x**1 + coefs[2] * x**2
+> print(y)
 > ~~~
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Thanks, @marijane. I think the change makes sense and this typo dates back to 2016 (bb8857f) when this exercises was first added to the lesson.

@annefou and @ldko: I think this piece misses a few words about cubic polynomial (e.g. "here is how we can compute a cubic polynomial" or something along these lines) -- what do you think?

PR #760, opened 2019-11-25
Clarification of 1.0 v 1.

coolernato:
The previous explanation implies that Python will always print 1.0 as "1." when there's nothing after the decimal point. However, this is specific to numpy. For example, the statement "print(1.0)" will print "1.0" and not "1.". I changed the description to refer to a numpy array instead of Python in general. I also clarified the space saved was on the screen (rather than in memory, for example).

### DIFF
--- a/_episodes/02-numpy.md
+++ b/_episodes/02-numpy.md
@@ -89,8 +89,8 @@ that output is the data we just loaded.
 By default,
 only a few rows and columns are shown
 (with `...` to omit elements when displaying big arrays).
-To save space,
-Python displays numbers as `1.` instead of `1.0`
+To save space on the screen
+when printing a numpy array, numbers will be displayed as `1.` instead of `1.0`
 when there's nothing interesting after the decimal point.
 
 > ## Importing libraries with shortcuts
--- a/_episodes/02-numpy.md
+++ b/_episodes/02-numpy.md
@@ -89,9 +89,7 @@ that output is the data we just loaded.
 By default,
 only a few rows and columns are shown
 (with `...` to omit elements when displaying big arrays).
-To save space on the screen
-when printing a numpy array, numbers will be displayed as `1.` instead of `1.0`
-when there's nothing interesting after the decimal point.
+To save space on the screen when displaying a NumPy array, numbers are shown as `1.` instead of `1.0` when there's nothing interesting after the decimal point.
 
 > ## Importing libraries with shortcuts
 >
--- a/_episodes/02-numpy.md
+++ b/_episodes/02-numpy.md
@@ -89,7 +89,7 @@ that output is the data we just loaded.
 By default,
 only a few rows and columns are shown
 (with `...` to omit elements when displaying big arrays).
-To save space on the screen when displaying a NumPy array, numbers are shown as `1.` instead of `1.0` when there's nothing interesting after the decimal point.
+Note that, to save space when displaying NumPy arrays, Python does not show us trailing zeros, so "1.0" becomes "1.".
 
 > ## Importing libraries with shortcuts
 >
--- a/_episodes/02-numpy.md
+++ b/_episodes/02-numpy.md
@@ -89,7 +89,7 @@ that output is the data we just loaded.
 By default,
 only a few rows and columns are shown
 (with `...` to omit elements when displaying big arrays).
-Note that, to save space when displaying NumPy arrays, Python does not show us trailing zeros, so "1.0" becomes "1.".
+Note that, to save space when displaying NumPy arrays, Python does not show us trailing zeros, so `1.0` becomes `1.`.
 
 > ## Importing libraries with shortcuts
 >


_episodes/02-numpy.md
```diff
@@ -89,8 +89,8 @@ that output is the data we just loaded.
 By default,
 only a few rows and columns are shown
 (with `...` to omit elements when displaying big arrays).
-To save space,
-Python displays numbers as `1.` instead of `1.0`
+To save space on the screen
+when printing a numpy array, numbers will be displayed as `1.` instead of `1.0`
 when there's nothing interesting after the decimal point.
```
ldko: @coolernato I think the points you are clarifying are worthwhile ones to make. Thanks for proposing the change. A few things I am thinking:

1. We should capitalize "NumPy" here (based on other usage in the episode).
2. I wonder about using the word "printing" , since in this case we aren't using `print`.
3. Not sure about using the tense of "will be".

What do you think of the wording:
```
To save space on the screen when displaying a NumPy array, numbers are shown as `1.` instead of `1.0` when there's nothing interesting after the decimal point.
```

coolernato: @ldko Yes, I agree with your points and think your re-phrasing makes sense. I will update the request to use that phrasing.

maxim-belkin:
Thanks for spotting the problem, Chris, and improving Chris' suggestion, Lauren!

Yes, we need to reword this part of the episode. How strongly do you both feel about these two pieces?
 
> To save space on the screen when displaying a NumPy array

> when there's nothing interesting after the decimal point.

Can we remove them? They both justify / explain the middle part (`numbers are shown as 1. instead of 1.0`) but I'm not positive they're necessary.

1. "there's nothing interesting after the decimal point" is a bit vague
2. the actual reason we see `1.` instead of `1.0` (or something else) is the `floatmode` setting which we can control with `numpy.set_printoptions()`: by default, NumPy displays elements of an array using the number of fractional digits necessary to uniquely represent this specific element (`numpy.get_printoptions()['floatmode']` returns `maxprec` -- see help page for `numpy.set_printoptions`).

What if we say something like this?

> Note, that when displaying NumPy arrays, Python does not show us trailing zeros, so `1.0` becomes `1.`

Do we nee to know the "reasons" why NumPy does what it does when displaying arrays?

coolernato:
@maxim-belkin I agree the "there's nothing interesting after the decimal point" is a bit vague. Your rewording regarding trailing zeroes is much clearer.

Regarding your second point, I would say this is the mechanism by which the format is controlled. I agree that going into detail about this mechanism would be too much. However, the reason it is by default set to "1." instead of "1.0" is to save space on the display. I think that there's value to pointing this out - otherwise learners will likely wonder why this behaviour (which many learners would not expect) occurs.

I would propose the following (or some reworded variant):

"Note, that to save space when displaying NumPy arrays, Python does not show us trailing zeros, so 1.0 becomes 1."

If you feel strongly that the explanation is unnecessary I would be happy to leave it out, as either version corrects the initial problem I identified.

slel:
How about something like this.

Note the `1.` for `1.0`: NumPy arrays skip trailing zeros
to save space, but keep the decimal point to tell apart
arrays of integers from arrays of floating-point numbers.

coolernato:
@slel The point that the period remains to differentiate it from an int is valid. I worry though that it might be going into too much detail?

If this were to be mentioned, it might be best splitting it into two sentences:

'Note, that to save space when displaying NumPy arrays, Python does not show us trailing zeros, so "1.0" becomes "1.". The period remains after the "1" to show the number is a float and not an int.'

This is my first pull request (on The Carpentries or elsewhere). When there are a multitude of valid options (as I believe we have here), who is it that makes the final decision of how to resolve an issue?

slel:
You are right, no need to mention ints here. I'm fine with any phrasing.
You and the lesson maintainers should settle on something snappy.

ldko:
I agree with both @maxim-belkin and @coolernato that the wording "nothing interesting after the decimal point" is worth replacing and with @coolernato that it is useful to give some reasoning of why there is nothing shown after the decimal point. If you give no explanation at all, you are stating only what is already shown by the output, so why state it at all?

```
"Note, that to save space when displaying NumPy arrays, Python does not show us trailing zeros, so 1.0 becomes 1."
```
sounds good to me, but I would either remove the word "that" or move the comma from before "that" to after "that".

@coolernato, regarding your question about who makes the final decision, I believe that, ideally, two of the maintainers agree on a resolution with the submitter, and we will merge the changes. I think we are close here!

coolernato:
@ldko A fine point regarding that comma. I've changed the phrasing to 'Note that, to save space when displaying NumPy arrays, Python does not show us trailing zeros, so "1.0" becomes "1.".' I think that's a simple, accurate and to the point explanation of the behaviour.

maxim-belkin:
I'd suggest using back-ticks instead of quotation marks to avoid any confusion.
Otherwise -- 👍

coolernato:
Fair point. Updated.

maxim-belkin:
Thank you!

PR #763, opened 2019-11-26
Update 04-lists.md

AnnAnnFryingPan:
I think it's good to repeat the explanation of the behaviour caused by the (attempted) copying and then modifying of a list in place. From teaching this several times, I think it would be useful to notify the learner that the use of append here to modify a 'copied' list is a second, similar example to what we saw earlier with the salsa list.

### DIFF
--- a/_episodes/04-lists.md
+++ b/_episodes/04-lists.md
@@ -268,7 +268,7 @@ odds after reversing: [11, 7, 5, 3]
 While modifying in place, it is useful to remember that Python treats lists in a slightly
 counter-intuitive way.
 
-As we saw earlier, when we modified the (salsa) list item in-place, if we make a list, (attempt to) copy it and then modify this list we can cause all sorts of trouble. This also applies to modifying the list using the above functions:
+As we saw earlier, when we modified the `salsa` list item in-place, if we make a list, (attempt to) copy it and then modify this list, we can cause all sorts of trouble. This also applies to modifying the list using the above functions:
 
 ~~~
 odds = [1, 3, 5, 7]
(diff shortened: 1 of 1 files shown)


ldko:
Thanks @AnnAnnFryingPan , I see what you mean about making it clear that this is a second example of "copying" a list and what can happen. Your proposed change is a valid solution and like you say, it is beneficial to repeat the point about modifying a list in place. I wonder though, with the lesson being overly full, what others think about removing one of the explanations of using `list()` (I would lean toward removing the `salsa` example). @annefou @maxim-belkin ?

If we want to keep both examples, I think your wording changes are good to include with the minor edit of adding back a comma before the clause `we can cause all sorts of trouble`.

annefou:
What I find a bit confusing is the fact that we have "Nested lists" and "Heterogeneous lists" between these two examples. So as you said @AnnAnnFryingPan it is then a good idea to remind learners that we are repeating what we said with another example. To make the episode a bit shorter, we could have at least "Nested lists" as complementary material. I usually skip it when teaching this part.

_episodes/04-lists.md
```diff
@@ -268,7 +268,7 @@ odds after reversing: [11, 7, 5, 3]
 While modifying in place, it is useful to remember that Python treats lists in a slightly
 counter-intuitive way.
 
-If we make a list and (attempt to) copy it then modify in place, we can cause all sorts of trouble:
+As we saw earlier, when we modified the (salsa) list item in-place, if we make a list, (attempt to) copy it and then modify this list we can cause all sorts of trouble. This also applies to modifying the list using the above functions:
```
ldko: @AnnAnnFryingPan , after discussing with the other maintainers, we would like to incorporate your suggestion to refer back to the earlier time in the episode where modifying in-place is covered. Before we do so, would you mind changing `(salsa)` to `` `salsa` `` (in back ticks instead of parentheses) , and add a comma back in before `we can cause all sorts of trouble`? Thank you for your contribution!

AnnAnnFryingPan: @Idko,  Thanks, I have done the requested updates.

PR #766, opened 2019-12-01
01-intro.md: New fundamentals exercise

zchilton1:
As part of my checkout, I created a short exercise that could be added to the Python Fundamentals episode.
-Zack Chilton

### DIFF
--- a/_episodes/01-intro.md
+++ b/_episodes/01-intro.md
@@ -220,4 +220,29 @@ weight in kilograms is now: 65.0
 > {: .solution}
 {: .challenge}
 
+> ## Seeing Data Types
+>
+> Python can show a variables data type using type(variable). 
+> What are the data types of the following variables?
+>
+> ~~~
+> planet = 'Earth'
+> type(planet)
+> apples = 5
+> type(apples)
+> distance = 10.5
+> type(distance)
+> ~~~
+> {: .language-python}
+>
+> > ## Solution
+> > ~~~
+> > <class 'str'>
+> > <class 'int'>
+> > <class 'float'>
+> > ~~~
+> > {: .output}
+> {: .solution}
+{: .challenge}
+
 {% include links.md %}
--- a/_episodes/01-intro.md
+++ b/_episodes/01-intro.md
@@ -220,7 +220,6 @@ weight in kilograms is now: 65.0
 > {: .solution}
 {: .challenge}
 
-> ## Seeing Data Types
 >
 > Python can show a variable's data type using type(variable). 
 > What are the data types of the following variables?
--- a/_episodes/01-intro.md
+++ b/_episodes/01-intro.md
@@ -233,8 +233,11 @@ weight in kilograms is now: 65.0
 >
 > > ## Solution
 > > ~~~
+> > type(planet)
 > > <class 'str'>
+> > type(apples)
 > > <class 'int'>
+> > type(distance)
 > > <class 'float'>
 > > ~~~
 > > {: .output}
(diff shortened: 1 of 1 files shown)


_episodes/01-intro.md
```diff
@@ -220,4 +220,29 @@ weight in kilograms is now: 65.0
 > {: .solution}
 {: .challenge}
 
+> ## Seeing Data Types
+>
+> Python can show a variable's data type using type(variable).
```
ldko: Since we will move introduction of `type()` to the main part of the episode, would you please remove the line above.

_episodes/01-intro.md
```diff
@@ -220,4 +220,29 @@ weight in kilograms is now: 65.0
...
 {: .challenge}
 
+> ## Seeing Data Types
+>
+> Python can show a variable's data type using type(variable). 
+> What are the data types of the following variables?
+>
+> ~~~
+> planet = 'Earth'
+> type(planet)
```
ldko: Would you please move the `type(planet)` call here in line 230 and the other `type()` calls given in 232 and 234 to the Solution section, to encourage the learners to recall `type()` on their own.

maxim-belkin:
Gentle ping.

zchilton1:
Sorry for the response delay. I removed the exercise name line and pushed that change. Please let me know if any further changes are required.

_episodes/01-intro.md
```diff
@@ -220,4 +220,28 @@ weight in kilograms is now: 65.0
 > {: .solution}
 {: .challenge}
 
+>
```
ldko: We actually wanted to keep your exercise title `> ## Seeing Data Types`. Please bring it back :).

_episodes/01-intro.md
ldko: Though we (the maintainers) still need to open a PR to introduce `type()` earlier in the episode, we would like you to remove the line `> Python can show a variable's data type using type(variable).` , so when we have both PRs merged, `type()` won't be introduced a second time.

_episodes/01-intro.md
ldko: We would still like for you to move the `type()` call lines to the solution, so learners doing the exercise can remember that they need to use `type()` on their own. i.e.:
```
> ~~~
> planet = 'Earth'
> apples = 5
> distance = 10.5
> ~~~
> {: .language-python}
>
> > ## Solution
> > ~~~
> > type(planet)
> > type(apples)
> > type(distance)
> > ~~~
> > {: .language-python}
> >
> > ~~~
> > <class 'str'>
> > <class 'int'>
> > <class 'float'>
> > ~~~
> > {: .output}
```

zchilton1:
I think my recent commit handles the requested changes. Let me know if anything is missing / off.

_episodes/01-intro.md
ldko: Hi Zack, thanks for the additional work--one further change would be appreciated: In your most recent commit https://github.com/swcarpentry/python-novice-inflammation/pull/766/commits/d097ea845347f9e1f9b24d7ab9ebf57002e99117 you removed the `type()` calls from the challenge question as requested, but we would still like to see the calls shown in the solution before showing the output. This should be possible using the syntax given in the above thread message. Thank you again!

maxim-belkin (approved):
👍🏻   Love it!

ldko (approved):
@zchilton1 we finally get the introduction of `type` into the main text of the intro episode, so this is ready to include now. Thanks very much for your work on this!

PR #771, opened 2019-12-17
Split 02-numpy.md

maxim-belkin:
Extracting visualization content into its own episode.

Once approved, this PR has to be merged with a merge commit.


ldko:
README.md should also be updated to reflect the new episode and re-numbered episodes.

_episodes/03-matplotlib.md
```diff
@@ -0,0 +1,256 @@
+---
+title: Analyzing Patient Data
```
ldko: This title should be changed.

maxim-belkin:
Thanks, Lauren!

_episodes/03-matplotlib.md
maxim-belkin: Updated!

_episodes/03-matplotlib.md
```diff
@@ -0,0 +1,256 @@
...
+teaching: 30
+exercises: 20
+questions:
+- "How can I visualize tabular data in Python?"
+- "How can I group several plots together?"
+objectives:
+- "Plot simple graphs from data."
+- "Group several graphs in a single figure."
+keypoints:
+- "Use the `pyplot` library from `matplotlib` for creating simple visualizations."
```
annefou: below we say "pyplot module from matplotlib library". Maybe we could use the same here. What do you think?

maxim-belkin: Thanks, Anne. I think it makes sense to use the same wording.

maxim-belkin:
Anne, Lauren: Thank you for reviewing this. Please post any additional comments if you have them. I'd like to merge this by the COB today if you think it is ready.

PR #775, opened 2020-01-20
Moved data description to lesson 2. 

rraffiu:
Based on my recent teaching experience of these lessons I think the data description is out of place. It should be in lesson two.

### DIFF
--- a/_episodes/01-intro.md
+++ b/_episodes/01-intro.md
@@ -3,6 +3,7 @@ title: Python Fundamentals
 teaching: 20
 exercises: 10
 questions:
+- "What is Python?"
 - "What basic data types can I work with in Python?"
 - "How can I create a new variable in Python?"
 - "Can I change the value associated with a variable after I create it?"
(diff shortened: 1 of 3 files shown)


maxim-belkin:
Thank you for your contribution, Rafi, and congratulations on your first pull request to our lesson.

We (maintainers of the lesson) discussed this at our previous meeting and I'm glad to see the first contribution related to the order of information in our lesson since "the big split" (the split of the first episode into three separate ones).

Anyways, here are my comments:

1. You introduce a section called "Python" with only two sentences in it -- I don't think we need a section for two sentences.
2. Portability is a questionable strength -- Python code relies on whitespaces and when these are not handled properly (e.g. when one does not use proper HTML tags), Python code breaks.
3. Unless we explain what "interpreted" means, I'd say we don't need this qualifier. I like "general-purpose", but it would be great to support this statement with examples of areas where it is used.
4.  Suggested location for the `## Inflammation data` section seems to interrupt the flow of (information in) the episode slightly.
5. Addition of the `Python` section is not related to the move of the `Inflammation data` section, so I'd suggest we do these two operations in two separate pull requests.

Thanks again!

rraffiu:
Thanks Maxim,

I have removed the python intro/section. And I am not a professional programming expert (rather a physicist who uses python to do his research) so I don't know the technically correct language. Here is what I feel as an instructor and python user:
1. This lesson needs a few lines introducing python.
2. A couple of line why is it so popular across disciplines (particularly non-specialists like myself). 

I will try to come up with something along these lines in a separate commit and you can help me iron out any wrinkles. 

Regarding the data description place. It may not be at the best place but it certainly is at a better place than before. This again is an instructors gut feeling. 

Thanks.

PR #777, opened 2020-01-24
02-numpy: clarify numpy.diff() part

elcorto:
Using a variable called `npdiff` to hold an array is highly confusing,
given that the function to be learned here is called {numpy,np}.diff().

Also, stress that the array returned by `diff()` is shorter by one.

### DIFF
--- a/_episodes/02-numpy.md
+++ b/_episodes/02-numpy.md
@@ -656,6 +656,9 @@ which is the average inflammation per patient across all days.
 
 > ## Change In Inflammation
 >
+> The patient data is _longitudinal_ in the sense that each row represents a
+> series of observations relating to one individual.  This means that
+> the change in inflammation over time is a meaningful concept.
 > Let's find out how to calculate changes in the data contained in an array
 > with NumPy.
 >
--- a/_episodes/02-numpy.md
+++ b/_episodes/02-numpy.md
@@ -656,21 +656,20 @@ which is the average inflammation per patient across all days.
 
 > ## Change In Inflammation
 >
-> This patient data is _longitudinal_ in the sense that each row represents a
-> series of observations relating to one individual.  This means that
-> the change in inflammation over time is a meaningful concept.
+> Let's find out how to calculate changes in the data contained in an array
+> with NumPy.
 >
-> The `numpy.diff()` function takes a NumPy array and returns the
-> differences between two successive values along a specified axis. For
-> example, consider a one-dimensional NumPy array of length 5 that looks like
-> this:
+> The `numpy.diff()` function takes an array and returns the differences
+> between two successive values. First we consider a one-dimensional
+> array of length 5. This could be part of some row `i` of our inflammation data,
+> i.e. `row_start = data[i,:5]`.
 >
 > ~~~
-> a = numpy.array([ 0,  2,  5,  9, 14])
+> row_start = numpy.array([ 0,  2,  5,  9, 14])
 > ~~~
 > {: .language-python}
 >
-> Calling `numpy.diff(a)` would do the following calculations
+> Calling `numpy.diff(row_start)` would do the following calculations
 >
 > ~~~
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Thanks, Steve. I agree with your reasoning and like the changes you make to the text. I wonder, though, if `a` is a good name for a variable. Shall we use a full word like `array` or something else? My co-maintainers, Anne and Lauren, might have some ideas.

Anyway, thanks for the pull request!

ldko:
Hi Steve, I agree with Maxim--I like the clarifications your edits contribute to the text but would prefer to avoid the short variable name `a` and try for something slightly more meaningful (realizing we are giving a generic example which makes it difficult to be very specific). I would, however, also prefer to avoid calling the variable `array` because of `numpy` having an `array` function and the output in the exercise appearing as `array([2, 3, 4, 5])` which could confuse some learners into making a connection between the variable name `array` and the function. I would be okay with something like `values` or `data` or `value_array` etc.

elcorto:
Valid points regarding `a` as variable name. I was thinking in numpy docstring terms where short variable names are often used, but of course in this context another naming is needed. I just pushed an update.

_episodes/02-numpy.md
```diff
@@ -656,38 +656,42 @@ which is the average inflammation per patient across all days.
 
 > ## Change In Inflammation
 >
-> This patient data is _longitudinal_ in the sense that each row represents a
-> series of observations relating to one individual.  This means that
-> the change in inflammation over time is a meaningful concept.
```
ldko: I am not sure about completely losing this sentence. I like how it set up why it is useful to learn how to use the `diff` function. What are your thoughts on removing it?

elcorto: As mentioned in the [commit message](https://github.com/swcarpentry/python-novice-inflammation/pull/777/commits/20d573e7e828a950b19ed8fe275b533f30ae6f8c):
```
The text part about "patient data is _longitudinal_" is redundant since
that answers the question posed later regarding the usage of the `axis`
keyword. It intoduces the usage of `axis`, while the following intro to
diff() first uses a 1d array.
```
But you are right, it motivates learning about `diff` in the first place, even though the parts about (i) "we can take diffs along axes" and (ii) "taking the diff along axis 1 is useful" is repeated later on. 

In any case, I'd leave it in then since the main point of this PR is the variable name and I have no better introductory line :).

maxim-belkin:
Thanks, Steve. I had the same concern as Lauren so I'm glad you reverted that change.

In any case, a question to both of you: what if we name our variable `patient3_week1` (or something like that)? I know this is not the best name for a variable but... it makes it easy to understand what information the variable contains.

elcorto:
I didn't do that since the original example uses an artificial 1d array of length 5 (enough for explaining what `diff()` does) instead of pulling a row (length 40) from the `data` array read in earlier in the episode, probably by design. My compromise was adding `row_start = data[i,:5]` in the text to make a connection from the following 1d example to the 2d array `data`, which wasn't present before.

ldko:
I am ok with using `row_start`, since the data doesn't actually map to a patient but hints at the possibility.

maxim-belkin:
`row_start` is OK... What I was trying to say is that this part of the lesson is currently independent of the lesson and I suggest to blend it in a little bit. For example, we can say that to see if there are any peculiarities in the inflammation of a particular patient over time, we can use `numpy.diff()` function. But we can certainly do it in a separate PR.

ldko:
If @elcorto wants to modify the exercise to reflect the actual patient data and further blend the exercise into the lesson as you suggest @maxim-belkin, I think that would be great, otherwise, I think it is ok to merge as-is and further blend it in a separate PR.

maxim-belkin:
Agreed.

elcorto:
Unfortunately, I won't have time to improve this episode more for now in the way you outlined. However, I found other issues that I hope to fix in later PRs.

As for this PR, I don't have any pending changes to add.

Issue #776, opened 2020-01-22
Analyzing Data from multiple files: generate composite statistics example


aliceminotto:
In https://github.com/swcarpentry/python-novice-inflammation/blob/gh-pages/_episodes/06-files.md the solution uses in place operators += and /=, which are only mentioned and explained in the next episode.
I'd suggest moving the "in place operators" example before this one here from the "making choices" lesson as it is.

ldko:
Hi Alice, thanks for catching this. As you indicate, we should avoid the in-place operators in the _Generate Composite Statistics_ exercise of the [_Analyzing Data from Multiple Files_ episode](http://swcarpentry.github.io/python-novice-inflammation/06-files/index.html), since they have not been introduced. Moving the _In-Place Operators_ exercise from the [_Making Choices_ episode](http://swcarpentry.github.io/python-novice-inflammation/07-cond/index.html) however would be problematic because the solution expects learners to use conditionals which would not have been taught yet. What do you think about instead replacing the use of in-place operators in the _Generate Composite Statistics_ exercise and its solution with basic operators (i.e. replace with `composite_data /= len(filenames)` with `composite_data = composite_data / len(filenames)` etc.)? Would you be interested in submitting a pull request for that?

aliceminotto:
Hi! Sure that would work, I'll do that! Cheers

maxim-belkin:
Closed in 6d32e70.

Thanks, Alice and Lauren.

PR #779, opened 2020-01-27
update exercise "generate composite statistics"

aliceminotto:
changes not to use in place operators, which are being introduced in a later lesson


aliceminotto:
see #776

PR #787, opened 2020-02-17
Clarifying info on average inflammation

alexdesiqueira:
The previous text did not mention, in a first moment, that the average was related to the patients' inflammation. Adding the word `inflammation` makes the text clearer.


alexdesiqueira:
Thank you Lauren!

Issue #781, opened 2020-02-04
Call for Contributions: image descriptions & titles


maxim-belkin:
### Overview 

We'd like to solicit contributions aimed at improving image descriptions (also known as "alternative text") in our lesson. The goal of this solicitation is to lower barriers to using Python Novice Inflammation lesson for learners with vision impairments and those who are unable to clearly see images for other reasons.

### Technical details

Our lesson is written in the Markdown language. Therefore, all images follow the same format:

```markdown
![image description](../fig/image.png "Optional title")
```

This call seeks contributions aimed at improving the text in square brackets above (`image description`).

### Current list of figures

- [x] 01-intro.md  >  `![Variables as Sticky Notes](../fig/python-sticky-note-variables-01.svg)`
- [x] 01-intro.md  >  `![Creating Another Variable](../fig/python-sticky-note-variables-02.svg)`
- [x] 01-intro.md  >  `![Updating a Variable](../fig/python-sticky-note-variables-03.svg)`
- [x] 02-numpy.md  >  `![Zero Index](../fig/python-zero-index.png)`
- [x] 02-numpy.md  >  `![Operations Across Axes](../fig/python-operations-across-axes.png)`
- [x] 03-matplotlib.md  >  `![Heatmap of the Data](../fig/inflammation-01-imshow.svg)`
- [x] 03-matplotlib.md  >  `![Average Inflammation Over Time](../fig/inflammation-01-average.svg)`
- [x] 03-matplotlib.md  >  `![Maximum Value Along The First Axis](../fig/inflammation-01-maximum.svg)`
- [x] 03-matplotlib.md  >  `![Minimum Value Along The First Axis](../fig/inflammation-01-minimum.svg)`
- [x] 03-matplotlib.md  >  `![The Previous Plots as Subplots](../fig/inflammation-01-group-plot.svg)`
- [x] 03-matplotlib.md  >  `![Plot with step lines](../fig/inflammation-01-line-styles.svg)`
- [x] 04-loop.md  >  `![Analysis of inflammation-01.csv](../fig/03-loop_2_0.png)`
- [x] 04-loop.md  >  `![loop_image](../fig/loops_image.png)`
- [x] 05-lists.md > `![The first element of a list. Adapted from @hadleywickham.](../fig/indexing_lists_python.png)`
- [x] 06-files.md  >  `![Analysis of inflammation-01.csv](../fig/03-loop_49_1.png)`
[...]

PR #792, opened 2020-02-23
provide detailed alt text to intro image 01.

blankdots:
Variables as Sticky Notes now has a detailed content: ../fig/python-sticky-note-variables-01.svg

- [x] `01-intro.md > Variables as Sticky Notes`

### DIFF
--- a/_episodes/01-intro.md
+++ b/_episodes/01-intro.md
@@ -139,7 +139,7 @@ weight in kilograms is now: 65.0
 > A variable is analogous to a sticky note with a name written on it:
 > assigning a value to a variable is like putting that sticky note on a particular value.
 >
-> ![Variables as Sticky Notes](../fig/python-sticky-note-variables-01.svg)
+> ![Value of 65.0 with weight_kg name stuck on it](../fig/python-sticky-note-variables-01.svg)
 >
 > This means that assigning a value to one variable does **not** change
 > values of other variables.


_episodes/01-intro.md
```diff
@@ -139,7 +139,7 @@ weight in kilograms is now: 65.0
 > A variable is analogous to a sticky note with a name written on it:
 > assigning a value to a variable is like putting that sticky note on a particular value.
 >
-> ![Variables as Sticky Notes](../fig/python-sticky-note-variables-01.svg)
+> ![Variables as Sticky Notes analogy example depicting weight_kg value as 65.0](../fig/python-sticky-note-variables-01.svg)
```
ldko: Per the webaim.org  suggestions, the `alt` attribute should be succinct, not add information not shown in the image, and does not need to repeat information given in the surrounding content. It also says it "is often a matter of personal interpretation" though, so it feels awkward to me to suggest changes to your interpretation, but based on the context of the image, do you think `value of 65.0 with weight_kg name stuck on it` (or some tweaked version of that) would be appropriate for more succinct text?

blankdots: Sure I will change, but I was aiming to illustrate also function of the image.

ldko: Regarding [function discussed by WebAIM](https://webaim.org/techniques/alttext/#functional) I interpret to mean function such as if the image is also a link or clickable, what does clicking on it do. We could mention sticky notes explicitly, though, if that is what you were thinking as far as function: `Value of 65.0 with sticky note on it reading weight_kg`.

blankdots: I think, your suggestion is fine and we can leave it at that, unless someone has other comments or suggestions.

maxim-belkin:
I think that currently proposed alternative text is very reasonable.

PR #793, opened 2020-02-23
provide detailed alt text to intro image 02.

blankdots:
Variables as Sticky Notes now has a detailed content: ../fig/python-sticky-note-variables-02.svg

Related to #781

Followed the recommendations at: https://webaim.org/techniques/alttext/#basics trying to be as specific as possible for the content and the function.

- [x] `01-intro.md > Creating Another Variable`

### DIFF
--- a/_episodes/01-intro.md
+++ b/_episodes/01-intro.md
@@ -157,7 +157,7 @@ weight in kilograms is now: 65.0
 > ~~~
 > {: .output}
 >
-> ![Creating Another Variable](../fig/python-sticky-note-variables-02.svg)
+> ![Value of 65.0 with weight_kg label stuck on it, and value of 143.0 with weight_lb label stuck on it](../fig/python-sticky-note-variables-02.svg)
 >
 > Let's now change `weight_kg`:
 >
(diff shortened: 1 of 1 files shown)


_episodes/01-intro.md
```diff
@@ -157,7 +157,7 @@ weight in kilograms is now: 65.0
 > ~~~
 > {: .output}
 >
-> ![Creating Another Variable](../fig/python-sticky-note-variables-02.svg)
+> ![Creating additional variable example: old variable weight_kg equals 65.0 and new variable weight_lb equals 143.0](../fig/python-sticky-note-variables-02.svg)
```
ldko: I think we could lose some of the starting text to make this shorter. What about something like `value of 65.0 with weight_kg name stuck on it and value of 143.0 with weight_lb name stuck on it`?

ldko:
Based on this [conversation](https://github.com/swcarpentry/python-novice-inflammation/pull/798#issuecomment-592126972), for consistency we should probably change "weight_kg name"  to "label weight_kg" and "weight_lb name" to "label weight_lb". @blankdots if you don't mind, in this same PR you could make the same change for intro image 01.

blankdots:
@ldko  Done.

ldko:
Thank you very much @blankdots

PR #795, opened 2020-02-23
provide detailed alt text to cond image 01.

blankdots:
Flowchart conditional image now has a detailed description: ../fig/python-flowchart-conditional.png

- [x] `07-cond.md > Executing a Conditional`

### DIFF
--- a/_episodes/07-cond.md
+++ b/_episodes/07-cond.md
@@ -48,12 +48,12 @@ done
 The second line of this code uses the keyword `if` to tell Python that we want to make a choice.
 If the test that follows the `if` statement is true,
 the body of the `if`
-(i.e., the set of lines indented underneath it) is executed.
+(i.e., the set of lines indented underneath it) is executed, and "greater" is printed.
 If the test is false,
-the body of the `else` is executed instead.
-Only one or the other is ever executed:
+the body of the `else` is executed instead, and "not greater" is printed.
+Only one or the other is ever executed before continuing on with program execution to print "done":
 
-![Executing a Conditional](../fig/python-flowchart-conditional.png)
+![A flowchart diagram of the if-else construct that tests if variable num is greater than 100](../fig/python-flowchart-conditional.png)
 
 Conditional statements don't have to include an `else`.
 If there isn't one,


_episodes/07-cond.md
```diff
@@ -53,7 +53,8 @@ If the test is false,
 the body of the `else` is executed instead.
 Only one or the other is ever executed:
 
-![Executing a Conditional](../fig/python-flowchart-conditional.png)
+![If-Else flow diagram for num variable testin if it is greather than 100.
+If true it prints 'greater' and 'done', if false it prints 'not greather' and 'done'.](../fig/python-flowchart-conditional.png)
```
ldko: Another possibility to consider is modifying the text before the figure slightly in order to reduce the `alt` attribute text. For example:

```
The second line of this code uses the keyword `if` to tell Python that we want to make a choice.
If the test that follows the `if` statement is true,
the body of the `if`
(i.e., the set of lines indented underneath it) is executed, and "greater" is printed.
If the test is false,
the body of the `else` is executed instead, and "not greater" is printed.
Only one or the other is ever executed before continuing on with program execution to print "done":

![If-Else flow diagram testing if variable num is greater than 100]
```

blankdots: good point, changed to that.

_episodes/07-cond.md
```diff
@@ -48,12 +48,12 @@ done
...
-(i.e., the set of lines indented underneath it) is executed.
+(i.e., the set of lines indented underneath it) is executed, and "greater" is printed.
 If the test is false,
-the body of the `else` is executed instead.
-Only one or the other is ever executed:
+the body of the `else` is executed instead, and "not greater" is printed.
+Only one or the other is ever executed before continuing on with program execution to print "done":
 
-![Executing a Conditional](../fig/python-flowchart-conditional.png)
+![If-Else flow diagram testing if variable num is greater than 100](../fig/python-flowchart-conditional.png)
```
maxim-belkin: What do you guys think about this alt. text ?
> A flowchart diagram of the if-else construct that tests if variable num is greater than 100

ldko: That sounds fine to me. In your version it is more clear that it is the if-else construct testing num > 100 vs. the diagram doing any actual testing.

blankdots: @maxim-belkin changed to that, thx!

ldko:
Thanks for doing all the follow up on this one, @blankdots

PR #796, opened 2020-02-23
provide detailed alt text to func image 01.

blankdots:
Python function structure image now has a detailed description: ../fig/python-function.svg

- [x] `08-func.md > The Blueprint for a Python Function`

### DIFF
--- a/_episodes/08-func.md
+++ b/_episodes/08-func.md
@@ -54,7 +54,7 @@ def fahr_to_celsius(temp):
 ~~~
 {: .language-python}
 
-![The Blueprint for a Python Function](../fig/python-function.svg)
+![Labeled parts of a Python function definition](../fig/python-function.svg)
 
 
 The function definition opens with the keyword `def` followed by the


_episodes/08-func.md
```diff
@@ -54,7 +54,10 @@ def fahr_to_celsius(temp):
 ~~~
 {: .language-python}
 
-![The Blueprint for a Python Function](../fig/python-function.svg)
+![The Blueprint for a Python Function.
+The function definition opens with the keyword 'def' followed 
+by the name of the function nd a parenthesized list of parameter names.
+The body of the function concludes with 'return' keyword followed by the return value](../fig/python-function.svg)
```
ldko: One of WebAIM's suggestions is not to be redundant. Since the description being added to the `alt` attribute here is already in the text underneath it, I don't think we need to repeat it. We could keep the original `alt` text or make it a little less abstract with something like "Labeled parts of a Python function definition".

PR #798, opened 2020-02-23
provide detailed alt text to intro image 03.

blankdots:
Additional variables as Sticky Notes now has a detailed content: ../fig/python-sticky-note-variables-03.svg

- [x] `01-intro.md > Updating a Variable`

### DIFF
--- a/_episodes/01-intro.md
+++ b/_episodes/01-intro.md
@@ -172,7 +172,7 @@ weight in kilograms is now: 65.0
 > ~~~
 > {: .output}
 >
-> ![Updating a Variable](../fig/python-sticky-note-variables-03.svg)
+> ![Value of 100.0 with label weight_kg stuck on it, and value of 143.0 with label weight_lb stuck on it](../fig/python-sticky-note-variables-03.svg)
 >
 > Since `weight_lb` doesn't "remember" where its value comes from,
 > it is not updated when we change `weight_kg`.


_episodes/01-intro.md
```diff
@@ -172,7 +172,7 @@ weight in kilograms is now: 65.0
 > ~~~
 > {: .output}
 >
-> ![Updating a Variable](../fig/python-sticky-note-variables-03.svg)
+> ![Updating the weight_kg variable with 100.0 as the value; weight_lb variable equals to 143.0](../fig/python-sticky-note-variables-03.svg)
```
ldko: Since the text before and after the image indicates `weight_kg` has updated and `weight_lb` has not, what do you think about making this follow the pattern of intro image 01 and 02:
"Value of 100.0 with weight_kg name stuck on it, and value of 143.0 with weight_lb name stuck on it"?

maxim-belkin:
I wonder if we should be using `label weight_*` or `word weight_*` instead of `weight_* name`.

ldko:
> I wonder if we should be using `label weight_*` or `word weight_*` instead of `weight_* name`.

I think "name" works in the context because of earlier in the episode mentioning  "a sticky note with a name written on it" and talking about variable names. However, f you and/or @blankdots feel strongly about using "label" or "word" before weight_* instead, I don't have a problem with that.

maxim-belkin:
I can't say I have a strong opinion here but "weight_* name" seems a bit weird because "name" is  a property of an object and in our case this object is a variable. So, when I read 'weight_kg name', it seems as if `weight_kg` object has a name which we don't mention.

blankdots:
used `label` instead of `name`.

PR #804, opened 2020-02-27
Add alt-text 06-files

neil-csiro:
#781 

Add alt-text to 06-files.


neil-csiro:
Committed to multiple images in the same episode.  Closing pull request and creating three new pull requests, for each image in this file individually.

PR #805, opened 2020-02-27
Add alt-text 06_files loop_49_1

neil-csiro:
#781 

Add alt-text for the first image, loop_49_1 in 06_files.


_episodes/06-files.md
```diff
@@ -85,7 +85,9 @@ inflammation-01.csv
 ~~~
 {: .output}
 
-![Analysis of inflammation-01.csv](../fig/03-loop_49_1.png)
+![Three plots are generated by the code applied to inflammation-01.csv. 
+The centre plot with the y-label max is triangular.  
+The other two plots are jagged triangles.](../fig/03-loop_49_1.png)
```
ldko: Since these plots are the same as the those in the three previous images, but side-by-side here, I wonder if it makes sense to skip re-describing the physical characteristics of the lines and say something like:  "Three side-by-side subplots show, from left to right, the average, maximum, and minimum values per day for the inflammation-01.csv data".

neil-csiro:
Thank you – I’ll update this in the next couple of days.

@ldko commented on this pull request.

In _episodes/06-files.md<https://github.com/swcarpentry/python-novice-inflammation/pull/805#discussion_r387410122>:

> @@ -85,7 +85,9 @@ inflammation-01.csv

 ~~~

 {: .output}



-![Analysis of inflammation-01.csv](../fig/03-loop_49_1.png)

+![Three plots are generated by the code applied to inflammation-01.csv.

+The centre plot with the y-label max is triangular.

+The other two plots are jagged triangles.](../fig/03-loop_49_1.png)

Since these plots are the same as the those in the three previous images, but side-by-side here, I wonder if it makes sense to skip re-describing the physical characteristics of the lines and say something like: "Three side-by-side subplots show, from left to right, the average, maximum, and minimum values per day for the inflammation-01.csv data

ldko:
Resolved via #908

PR #807, opened 2020-02-27
Add alt-text 06-files image loop_49_5

neil-csiro:
#781 

Add alt-text to above episode and image.


ldko:
This was resolved by #922. Thanks, @neil-csiro !

PR #808, opened 2020-02-27
Add alt-text to 10-defensive overlapping_ranges

neil-csiro:
#781 

Added alt-text to overlapping_ranges image in episode 10-defensive.

### DIFF
--- a/_episodes/10-defensive.md
+++ b/_episodes/10-defensive.md
@@ -277,7 +277,9 @@ The range of each time series is represented as a pair of numbers,
 which are the time the interval started and ended.
 The output is the largest range that they all include:
 
-![Overlapping Ranges](../fig/python-overlapping-ranges.svg)
+![Four overlapping lines represent four time series, as described in the preceeding text. 
+They begin and end at different points in time, but common to all four lines is time 
+values zero through two.](../fig/python-overlapping-ranges.svg)
 
 Most novice programmers would solve this problem like this:
 


_episodes/10-defensive.md
```diff
@@ -277,7 +277,9 @@ The range of each time series is represented as a pair of numbers,
 which are the time the interval started and ended.
 The output is the largest range that they all include:
 
-![Overlapping Ranges](../fig/python-overlapping-ranges.svg)
+![Four overlapping lines represent four time series, as described in the preceeding text. 
+They begin and end at different points in time, but common to all four lines is time 
+values zero through two.](../fig/python-overlapping-ranges.svg)
```
ldko: I like the explanation you have included to describe the range overlapping, though I would modify it slightly, perhaps something close to "Four horizontal lines with a number at each end represent time series. They begin and end at different points in time, but common to all is the range 0.0 through 2.0."

neil-csiro:
Thank you – I’ll update this in the next couple of days.

@ldko commented on this pull request.

In _episodes/10-defensive.md<https://github.com/swcarpentry/python-novice-inflammation/pull/808#discussion_r387416507>:

> @@ -277,7 +277,9 @@ The range of each time series is represented as a pair of numbers,

 which are the time the interval started and ended.

 The output is the largest range that they all include:



-Overlapping Ranges

+Four overlapping lines represent four time series, as described in the preceeding text.

+They begin and end at different points in time, but common to all four lines is time

+values zero through two.

I like the explanation you have included to describe the range overlapping, though I would modify it slightly, perhaps something close to "Four horizontal lines with a number at each end represent time series. They begin and end at different points in time, but common to all is the range 0.0 through 2.0."

ldko:
Closing this for #922. Thanks for submitting this @neil-csiro !

PR #809, opened 2020-03-02
Briefly introduce dictionaries before using them in the exercise

### LINKED ISSUE #783: Dictionaries not introduced but used in exercise in "Error and Exceptions" - & suggestiong
NB: Good first issue label (cannot be added because not in contributor list)

Exercise `Reading error Messages` - Lesson "Error and Exceptions"
(http://swcarpentry.github.io/python-novice-inflammation/09-errors/index.html)

- As dictionaries are not introduced previously in lesson (maybe a consequence of reducing the lesson) I suggest the following:

1. Moving this exercise at the end of the first lession section - before syntax errors  (will be used to ensure that students have understood the traceback - and it is time for an exercice) 
Then this will also provide a good introduction to syntax errors (as the key error is a syntax error, and will naturly flow with the Caturday joke)  
2. Provide a short introduction to exercise: so that students can read the exercise more easily without focussing on the fact that they have not encountered dictionaries before
 
This can be introduced (in a minimal way) by the instructor when introducing the exercise and should not take more than 1 min (avoiding extra material load for the course)

> In this exercise we introduce a data structure that you did not previously encounter: dictionary. 
In a dictionary: keys are associated to values: key-values pairs. Values can be of different types (ie:numbers, lists...) but keys must be unique and immutable (no lists). Dictionaries can be constructed using the following syntax:
`a_dictionary = {'key1' : value1,
                         'key2': value2}`

> Dictionaries are not ordered, consequently we cannot access values using indexes, but we can acess values using their corresponding key. Ex: 
`print(a_dictionary['key1'])`

All the best
Eve

tobiashuste:
Make sure that dictionaries are briefly introduced before using them in an exercise.

- The exercise "Reading Error Messages" is moved before the "Syntax
Errors" section, in order to ensure learners understood Tracebacks
before moving onwards.
- A dictionary is used in the "Reading Error Messages" exercise which has
not been introduced before. The learner should be able to get to know
dictionaries by this very brief explanation.

Closes #783

### DIFF
--- a/_episodes/09-errors.md
+++ b/_episodes/09-errors.md
@@ -127,6 +127,73 @@ as it is possible to create custom errors.
 In that case,
 hopefully the custom error message is informative enough to help you figure out what went wrong.
 
+> ## Reading Error Messages
+>
+> Read the Python code and the resulting traceback below, and answer the following questions:
+>
+> 1.  How many levels does the traceback have?
+> 2.  What is the function name where the error occurred?
+> 3.  On which line number in this function did the error occur?
+> 4.  What is the type of error?
+> 5.  What is the error message?
+>
+> ~~~
+> # This code has an intentional error. Do not type it directly;
+> # use it for reference to understand the error message below.
+> def print_message(day):
+>     messages = {
+>         "monday": "Hello, world!",
+>         "tuesday": "Today is Tuesday!",
+>         "wednesday": "It is the middle of the week.",
+>         "thursday": "Today is Donnerstag in German!",
+>         "friday": "Last day of the week!",
+>         "saturday": "Hooray for the weekend!",
+>         "sunday": "Aw, the weekend is almost over."
+>     }
+>     print(messages[day])
+>
+> def print_friday_message():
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Thanks for the PR, Tobias. We will discuss this PR at our next maintainers meeting.

maxim-belkin:
Hi Tobias,

Thanks for putting this together. I agree with you and @evezeyl that using dictionaries in an exercise without prior introduction is strange. However, "brief introductions" do increase cognitive load on learners and, therefore, we (maintainers) wonder if some other solution is necessary. The exercise in question does not have to have dictionaries because the key questions it poses are not related to the nature of this data structure. Would you agree to change the exercise to use list instead of dictionary? A list can contain names of the days of the week ("Monday", "Tuesday", ..."), first function prints name of the week day that corresponds to the specified day number and the second function calls the first function with index 7.

Thanks for you contribution again!
Maxim

CC @swcarpentry/python-novice-inflammation-maintainers

PR #802, opened 2020-02-26
#781 Provide alt text for most episodes

neil-csiro:
#781 

Alt-text suggestions have been provided for most episodes.  I have not updated Episodes 4 and 2 as they have more complex diagrams, and I would like feedback on my text on the simpler diagrams before attempting the last two episodes.  Thanks.  I'm an inexperienced github/carpentries project user, so please advise if I should adjust my process for making these changes.

PR #810, opened 2020-03-03
added more detail to alt text of image inflammation-01-imshow.svg

jdilger:
Updated the alt text for inflammation-01-imgshow.svg to better reflect what is shown. 
#781

### DIFF
--- a/_episodes/03-matplotlib.md
+++ b/_episodes/03-matplotlib.md
@@ -27,7 +27,7 @@ matplotlib.pyplot.show()
 ~~~
 {: .language-python}
 
-![Heatmap of the Data](../fig/inflammation-01-imshow.svg)
+![Heatmap Of Inflammation Data Created With Matplotlib](../fig/inflammation-01-imshow.svg)
 
 Blue pixels in this heat map represent low values, while yellow pixels represent high values.  As we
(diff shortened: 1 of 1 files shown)


ldko:
Hi John, Thank you for proposing greater detail for alt text on this image. In reading https://webaim.org/techniques/alttext/, it discourages including information in the description that the image itself does not contain and to avoid redundancy with the text around the image. Thus for me, "Created With Matplotlib" is not needed because it is not something the image itself indicates, and the text leading up to the image indicates the heat map is generated with Matplotlib. 

@neil-csiro also proposed alt text for the heat map image in [another PR](https://github.com/swcarpentry/python-novice-inflammation/pull/802/files#diff-bdf0ae402c219a3ef52c5a69b717ad6aR30) that isn't yet incorporated.

@jdilger perhaps we could incorporate some of what @neil-csiro suggested and change the text in your PR to something like "Heat map with y axis from zero to sixty and x axis from zero to forty. Moving left to right, pixels the full height of the map change color from dark blue to blue green to yellow then back to blue green then dark blue."

ldko:
The alt text for inflammation-01-imshow.svg was updated in #888 .

PR #811, opened 2020-03-09
fix SyntaxError in function exercise

andreas-h:
otherwise, `non-default argument follows default argument` is raised


andreas-h:
stupid me, that was the point of the exercise. sorry for the noise.

ldko:
Hey @andreas-h , it's no problem! Thanks for taking the time to put in a pull request anyway.

PR #816, opened 2020-03-24
Add notebook to references

twei55:
I would suggest to link the term notebook in the second episode to an entry in the reference section which I added in this Pull request. The term has not been introduced before and some students might not know what it refers to.

### DIFF
--- a/_episodes/02-numpy.md
+++ b/_episodes/02-numpy.md
@@ -82,7 +82,7 @@ a line. These both need to be character strings (or [strings]({{ page.root }}/re
 for short), so we put them in quotes.
 
 Since we haven't told it to do anything else with the function's output,
-the notebook displays it.
+the [notebook]({{page.root}}/reference/#notebook) displays it.
 In this case,
 that output is the data we just loaded.
 By default,
--- a/_episodes/02-numpy.md
+++ b/_episodes/02-numpy.md
@@ -82,7 +82,7 @@ a line. These both need to be character strings (or [strings]({{ page.root }}/re
 for short), so we put them in quotes.
 
 Since we haven't told it to do anything else with the function's output,
-the [notebook]({{page.root}}/reference/#notebook) displays it.
+the [notebook]({{ page.root }}/reference/#notebook) displays it.
 In this case,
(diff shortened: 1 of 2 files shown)


reference.md
```diff
@@ -161,6 +161,10 @@ mutable
 :   Changeable. The value of mutable data can be altered after it has been
     created. See [immutable](#immutable)."
 
+notebook
+:   Interactive computational environment in your web browser, in which you can write and execute Python code.
+    Examples are IPython or Jupyter notebooks
+
```
ldko: 1. To me "in your web browser" is slightly vague. Does it sound any more clear to you to say "Interactive computational environment accessed via your web browser"? 
2. Should we also indicate they can contain text and visualizations? Something like "write and execute Python code plus display text and visualizations"?
3. To follow the style of other glossary entries, I would end the text with a full stop (`.`).

_episodes/02-numpy.md
```diff
@@ -82,7 +82,7 @@ a line. These both need to be character strings (or [strings]({{ page.root }}/re
 for short), so we put them in quotes.
 
 Since we haven't told it to do anything else with the function's output,
-the notebook displays it.
+the [notebook]({{page.root}}/reference/#notebook) displays it.
```
ldko: For consistency of style, we should add spaces around `page.root` (i.e. `{{ page.root }}`).

reference.md
maxim-belkin: Perhaps we could base our definition off the one used by the Jupyter project itself:

_from https://jupyter-notebook.readthedocs.io/en/latest/notebook.html#main-features-of-the-web-application_

> a browser-based tool for interactive authoring of documents which combine explanatory text, mathematics, computations and their rich media output.

maxim-belkin:
I concur with Lauren -- I think this is a good addition.
Thanks, Tobias!

reference.md
twei55: @ldko @maxim-belkin Thanks for your feedback and suggestions. I tried to combine both of your ideas, maybe you can have second look. As I am not a native speaker, I would trust your recommendations ;-)

Issue #664, opened 2019-07-18
Lesson 1 - definition of 'libraries', 'modules' and 'functions'


HeatherAn:
In Lesson 1 - Analyzing Patient Data, 'Loading data into Python' section, when referring to 'libraries', perhaps it would be better to state from the beginning what a 'library', a 'module' and a 'function' are. 

The lesson would benefit if this distinction is made early on. Something along the lines of  'Python has functions that come by default (like print() for example), but there are also a lot of very useful functions in so-called libraries. Each library has functions that do specific tasks organized in different units called modules. Thus we can use the functions that come by default, or we can use the functions within libraries. Whenever we want to use functions within a given library, then we have to first import that library, and then call the functions using the dotted notation (library.function).'

PR #817, opened 2020-03-29
Update 03-matplotlib.md

hanlei9876:
In correspondence to issue #664

Add a short explanation to clarify the differences between libraries and modules

I agree with the point of clarifying the differences between the concepts of 'libraries' and 'modules'. Because if not explaining them, there might be confusions to audiences. But I feel that the main thread of this lesson is to talk about Python fundamental concepts and dealing with datasets, and the terminologies of 'libraries' and 'modules' are not that much important. So, as a tradeoff, I choose to add a quick note of clarification below the first appearance of the concept 'module', so the audience won't get confused here while still following the main thread of this lesson.

---

---

### DIFF
--- a/_episodes/03-matplotlib.md
+++ b/_episodes/03-matplotlib.md
@@ -27,6 +27,13 @@ matplotlib.pyplot.show()
 ~~~
 {: .language-python}
 
+## Differences between libraries and modules
+>
+> Above, we import function 'imshow(data)' from module 'pyplot' instead of from 'matplotlib'. There are many libraries for 
+> Python users, offering various functions to help us perform tasks. Some functions are stored in libraries so we can call 
+> them directly. For example, we use `numpy.loadtxt(...)` in last episode to call function 'loadtxt()' from library 'numpy'. > But some libraries have functions organized in different units called modules. In this case, we have to first import that > library 'matplotlib', and then import the module 'pyplot' and call the functions 'imshow' inside by using the sequential dotted notations as 'matplotlib.pyplot.imshow()'.
+{: .callout}
+
 ![Heatmap of the Data](../fig/inflammation-01-imshow.svg)
 
 Blue pixels in this heat map represent low values, while yellow pixels represent high values.  As we


maxim-belkin:
Hi, @hanlei9876! Thank you for the pull request and congratulations on your first contribution to our lesson. 🎉  I share my opinion below but keep in mind that there are other maintainers and our opinions may differ, so this is not the final decision.

So, here is what I think:

1. We don't need to know the differences between libraries/modules/packages/etc. to execute the `import` statements. The main point is that we have to *import* something to later use functions from it.

2. Explaining these differences would increase the cognitive load on learners -- the amount of information learners have to put into their working memories in order to understand something.

3. In the grand scheme of things, our goal isn't to teach learners peculiarities of Python terminology but to analyze data in Python. Deviating from the main story line to explain differences between modules and packages does not seem to be necessary.

I guess it is clear where I'm heading with all of this: I don't think we can accommodate proposed changes in our lesson. Nonetheless, I'm looking forward to learning what other maintainers (Lauren (@ldko) and Anne (@annefou)) think about it.

Again, thank you for your contribution!

ldko:
Thanks for taking the time to submit a pull request @hanlei9876. :) It's good to have you participating.

I agree with you that focusing on the difference in terminologies of 'libraries' and 'modules' isn't that important and thus may divert from the main story line, which leads me to agree with @maxim-belkin that we won't be able to include this text in the episode. However, I think you also make a valid point that undefined terminology in the lesson can be confusing to audiences.

Since the matplotlib episode is where the term "module" is introduced, a different option than adding content to the episode is to link the first use of the term "module" to a glossary definition that we add. This way, if people want clarification, they have a way to quickly access a definition, but it does not slow the pace of the episode nor add cognitive load to new learners who aren't ready for it. Basing off of the [Python's explanation of a module](https://docs.python.org/3/tutorial/modules.html), perhaps a definition we might give could be:

> A file containing executable statements and Python definitions. The definitions in a module, including functions and variables, may be imported into other modules, scripts, and interactive interpreters for use.


Creating and linking to a glossary entry like this would be similar to what we are working on for the term "notebooks" in pull request #816.

What do you all think of this idea @maxim-belkin @hanlei9876 ?

hanlei9876:
Thank you very much for your fresh ideas, @maxim-belkin @ldko.

After carefully reading through your opinions and suggestions. Now I agree with you both on that introducing the terminologies of 'libraries' and 'modules' will increase the cognitive load to new students. 

Considering the undefined terminology in the lesson can be confusing to audiences, so the suggestion by @ldko is a very good practice to link the first use of the term "module" to a revised but more efficient glossary definition by @ldko.

Once again, really appreciate the wonderful feedback by you two @maxim-belkin @ldko.

PR #818, opened 2020-04-02
update explanation of the slicing element string

zhaophotonicslab:
To slice the last three characters, one should use element[-3 : ], which might be a more universal way for any string with length >=3. 
Using element[3:6] refers to the 4th to the 6th characters by its literal meaning.

---

---

### DIFF
--- a/_episodes/02-numpy.md
+++ b/_episodes/02-numpy.md
@@ -548,15 +548,26 @@ which is the average inflammation per patient across all days.
 > {: .solution}
 >
 > How can we rewrite the slice for getting the last three characters of `element`,
-> so it will work on any character string at least three characters in length?
+> so that it works even if we assign a different string to `element`?
+> Test your solution with the following strings: `carpentry`, `clone`, `hi`. 
 >
 > > ## Solution
 > > ~~~
+> > element = 'oxygen'
+> > print('last three characters:', element[-3:])
+> > element = 'carpentry'
+> > print('last three characters:', element[-3:])
+> > element = 'clone'
+> > print('last three characters:', element[-3:])
+> > element = 'hi'
 > > print('last three characters:', element[-3:])
 > > ~~~
 > > {: .language-python}
 > > ~~~
 > > gen
+> > try
+> > one
+> > hi
 > > ~~~
 > > {: .output}
 > {: .solution}
--- a/_episodes/02-numpy.md
+++ b/_episodes/02-numpy.md
@@ -547,14 +547,14 @@ which is the average inflammation per patient across all days.
 > > effectively removing the first and last letters from 'oxygen'
 > {: .solution}
 >
-> How can we rewrite the slice for getting the last three characters of `element` does,
-> so it will work on any character string at least three characters in length.
+> How can we rewrite the slice for getting the last three characters of `element`,
+> so it will work on any character string at least three characters in length?
 >
 > > ## Solution
 > > ~~~
 > > print('last three characters:', element[-3:])
-> ~~~
-> {: .language-python}
+> > ~~~
+> > {: .language-python}
 > > ~~~
 > > gen
 > > ~~~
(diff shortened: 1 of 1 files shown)


ldko:
Hi @zhaophotonicslab, congratulations on opening your first Software Carpentry PR! You make a valid point here that if you want a general slice of the last 3 characters of a string you should use `[-3 : ]`. However, I think where you are making this change, the negative indexing has not yet been introduced. Since in the case of "oxygen", `element[3:6]` does also give you the last three characters, I suggest we leave that as is, and then add a final part to the Challenge (after the solution for explaining what `element[1:-1]` does) that asks something like "How can we rewrite the slice for getting the last three characters of `element`, so that it works on any character string of at least three characters"?

zhaophotonicslab:
Dear Lauren,

Thanks for your prompt response. It is a good idea to put it as an exercise to the Challenge. I agree with you that introducing negative indexing may bring confusions or take more time to explain to students.

ldko:
Hi @zhaophotonicslab, since negative indexing is brought up in the Challenge shortly after where you have made an edit, after that point, we have a good opportunity to show the way you suggest of getting the last three characters. Would you like to modify this Pull Request to change line 508 back to what it was then ask  "How can we rewrite the slice for getting the last three characters of `element`, so it will work on any character string at least three characters in length"? around line 550?

zhaophotonicslab:
Hi Lauren,

Thanks for the suggestions. I agree it is better to put it in the challenge part. I have modified my branch as suggested. Do I need to create a new pull request?

The commit ID to my own branch is 1334999155cb8596377b77e7b4f6072ae5d182e8

I might have also accidently committed to swcarpentry gh-page branch. If I did, please ignore that.

Thanks again for your response.

_episodes/02-numpy.md
```diff
@@ -546,6 +546,20 @@ which is the average inflammation per patient across all days.
...
 > {: .solution}
+>
+> How can we rewrite the slice for getting the last three characters of `element` does,
+> so it will work on any character string at least three characters in length.
+>
+> > ## Solution
+> > ~~~
+> > print('last three characters:', element[-3:])
+> ~~~
+> {: .language-python}
```
ldko, line 557: There is a slight syntax error here that is causing a failed check on the Pull Request. I think we can fix it with another greater than symbol on the two above lines:
```
> > ~~~
> > {: .language-python}
```

_episodes/02-numpy.md
```diff
@@ -546,6 +546,20 @@ which is the average inflammation per patient across all days.
 > > Creates a substring from index 1 up to (not including) the final index,
 > > effectively removing the first and last letters from 'oxygen'
 > {: .solution}
+>
+> How can we rewrite the slice for getting the last three characters of `element` does,
+> so it will work on any character string at least three characters in length.
```
ldko, line 551: Can you remove the word "does" on line 550 and change the period (`.`) to a question mark (`?`) at the end of line 551?

zhaophotonicslab:
Hi @Idko,

Thanks for pointing out the errors. I have fixed the syntax errors as suggested.

_episodes/02-numpy.md
```diff
@@ -546,6 +546,20 @@ which is the average inflammation per patient across all days.
 > > Creates a substring from index 1 up to (not including) the final index,
 > > effectively removing the first and last letters from 'oxygen'
 > {: .solution}
+>
+> How can we rewrite the slice for getting the last three characters of `element`,
+> so it will work on any character string at least three characters in length?
```
maxim-belkin, line 551: Lauren, what do you think about this:

> How can we rewrite the slice for getting the last three characters of `element` so that it works even if we assign a different string to `element`? Test your solution with the following strings: `carpentry`, `clone`, `hi`.

ldko: I think that change is fine to make. It might be nice having `hi` as an example, since it will show learners that it doesn't error to take a slice bigger than the string.

zhaophotonicslab:
That’s a good idea. Thanks to @maxim-belkin and @Idko. The changes have been made.

Issue #769, opened 2019-12-04
Move "mystery functions in IPython" box to earlier in lesson


scastanedabarba:
In the "Analyzing patient data" lesson, we start using functions in IPython really early and mention the parameters that numpy.loadtext has. I think it would be helpful to move "Mystery functions in IPython" to right after this point so that we can explain how we know that this takes exactly 2 parameters and that they need to be strings. At the very least I think it would be good to insert a sentence saying something like;

Later in the lesson, we will learn how to access explanations of functions and we will also explore how we can determine what other functions are available within a library.

twei55:
I work with Python for a couple of years but never heard about "mystery functions". I asked colleagues of mine that also never heard of it. I used Google to search for some definition but it seems only to be used in the context of software carpentries. Is this a wording used in a specific community? Otherwise I would suggest to rename it something like in-line help.

PR #822, opened 2020-05-14
Add a reference to the API documentation and introduce the concept

cgaylord-vt:

---

---

### DIFF
--- a/_episodes/03-matplotlib.md
+++ b/_episodes/03-matplotlib.md
@@ -124,7 +124,7 @@ explore how to use various Python libraries, like `matplotlib`. These details ar
 contained in the "API Documentation" for `matplotlib`. Because `matplotlib` is
 not part of the core Python language, it has its own [documentation web site]
 (https://matplotlib.org/). In particular, if we search, we find the specifics
-for [add_subplot]
+for `add_subplot`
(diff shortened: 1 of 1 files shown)


cgaylord-vt:
In my experience, when novices first see something like add_subplot they find it to be magic: "how did you know that's how to use it", etc. I think this is a good place to introduce the concept that there are reference materials to refer to when writing code, both to understand the process of finding functionality as well as specifics of how that functionality works.

ldko:
Hi @cgaylord-vt thank you for your contribution of opening this pull request. I agree with your noted experience of novices wondering how you know how to use a function or method. I think your explanation ties in with the _Mystery Functions in IPython_ callout that is in [episode 2](https://swcarpentry.github.io/python-novice-inflammation/02-numpy/index.html) where the solution given there is to get information via `help` or `?`. I think it might be helpful to bring up documentation here. There is also an issue open (#769) about possibly moving this explanation earlier and a comment about changing the name of the callout. I am wondering what you @cgaylord-vt and other maintainers @maxim-belkin @annefou  think about putting the explanation of documentation with this content and possibly moving it earlier (and renaming).

Regarding the text you have included about documentation, I think if we mention "API" we perhaps need to link it to an entry in the glossary, as I suspect this is a term that novices may not know. Let's see if we can get other feedback about this. Thanks again for your work on this @cgaylord-vt !

Issue #823, opened 2020-05-26
02-numpy: inconsistent text/images for array indexing and method calls


jdkent:
## Array Indexing

The current text indexes arrays like: `data[0, 0]`, but the image indexes the array like `data[0][0]`.
These methods are equivalent (AFAIK), but this was not mentioned in the text.

![](https://github.com/swcarpentry/python-novice-inflammation/raw/gh-pages/fig/python-zero-index.png)

## Method Calls

The current text calls functions from `numpy` to operate on `data` (`numpy.max(data, axis=1)`),
but the image calls the `.max()` method from `data` (`data.max(axis=1)`), and I did not find explicit mention of their equivalency in the text.
![](https://swcarpentry.github.io/python-novice-inflammation/fig/python-operations-across-axes.png)

Thanks for all your work on these lessons, let me know if I'm missing something that makes this a non-issue.

ldko:
Hi @jdkent,  thanks for pointing out the inconsistencies. I think in both cases, it would be desirable to update the images to match the text. Regarding the `data.max` vs. `numpy.max` function or method issue, previously the method was also used in the text, but a decision was made to [consistently use functions over the methods](https://github.com/swcarpentry/python-novice-inflammation/pull/244). The image has not been updated to match since then. In the array indexing, I am not sure about why the image uses the `data[0][0]` syntax instead of `data[0, 0]`, which is the syntax given in the episode text and `numpy` documentation, so it would be good to change to prevent confusion. Would you be able to submit a pull request to update the images to remedy the inconsistencies? Thank you!

ldko:
Hi James, just want to check and see if you think you will have time to work on this in the near future. Thank you!

PR #826, opened 2020-05-27
[FIX] 02-numpy: change method call to function call in images

jdkent:
partial fix of #823 

preview of image:

current image used in episode:


maxim-belkin:
Thank you for updating the image, @jdkent! Please update the episode as well (02-numpy.md) so that the image appears in the lesson.

maxim-belkin:
Oops, my bad -- you updated both files! Thank you!

PR #827, opened 2020-05-27
[FIX] 02-numpy: update zero index image

jdkent:
partial fix of #823


ldko:
Thanks @jdkent  for making the change to use the `data[0, 0]` syntax. It is nice to have it match the episode text now.

I noticed that the size of the file went from 51.4 KB to 184 KB, and the dimensions of the image went from 2000px x 326px to 3000px x 1000px. I think the change in height that seems to add a noticeable amount of whitespace in the episode is the difference that stands out the most to me. Also it seems like the png doesn't display Helvetica Bold, which to me isn't a big deal, but I'm not sure if anyone else feels strongly about the text no longer being bold. Any thoughts about these points @jdkent @maxim-belkin or anyone else?

jdkent:
Thanks! I didn't check the original size of the image, I can specify that in the `svg`, and I can work on getting the font to be Helvetica Bold (I'm not using adobe illustrator, which appears to be how the original png was made).

jdkent:
switched to the openly available font sans-serif and made the png file 2000px by 326px.

maxim-belkin:
Before we discuss fonts, etc, I suggest to discuss the positioning of those `data[x][y]` in the right-hand-side of the figure first. It seems that `data[1, 0]` should be under `data[0, 0]` and not where it is currently located. What do you guys think?

maxim-belkin:
Link to the updated figure:
https://raw.githubusercontent.com/jdkent/python-novice-inflammation/update_zero_index/fig/python-zero-index.svg

Link to the current figure:
https://raw.githubusercontent.com/swcarpentry/python-novice-inflammation/gh-pages/fig/python-zero-index.svg

ldko:
@maxim-belkin I had never thought about that before, since I have always looked at the colors and letters matching up correctly. I hadn't thought about the positioning of the code references aligning with the placement in the ndarray. What you suggest makes sense to me if @jdkent would be willing to change the placement.

jdkent:
@maxim-belkin that would make more intuitive sense to me as well.

fig/python-zero-index.svg
```diff
@@ -1,14 +1,15 @@
...
+	 viewBox="0 0 460 74.6" style="enable-background:new 0 0 460 74.6;" xml:space="preserve"
+	 width="2000px" height="326px">
 <style type="text/css">
 	.st0{opacity:0.75;fill:#FFF200;}
 	.st1{opacity:0.5;fill:#F7941D;}
 	.st2{opacity:0.25;fill:#EC008C;}
 	.st3{fill:none;}
 	.st4{fill:#00A651;}
-	.st5{font-family:'Helvetica-Bold';}
+	.st5{font-family:'sans-serif';}
```
ldko, line 12: Perhaps the `font-weight="bold"` should be in here?

fig/python-zero-index.svg
```diff
@@ -21,55 +22,78 @@
...
+<text transform="matrix(1 0 0 1 75.754 34.6281)" font-weight="bold"><tspan x="0" y="0" class="st4 st5 st6">B</tspan><tspan x="0.3" y="14.4" class="st7 st5 st6">E</tspan><tspan x="0" y="28.8" class="st8 st5 st6">H</tspan></text>
 <rect x="89.4" y="26" class="st3" width="17.4" height="38.6"/>
-<text transform="matrix(1 0 0 1 93.754 34.6281)"><tspan x="0" y="0" class="st4 st5 st6">C</tspan><tspan x="0.7" y="14.4" class="st7 st5 st6">F</tspan><tspan x="2.7" y="28.8" class="st8 st5 st6">I</tspan></text>
+<text transform="matrix(1 0 0 1 93.754 34.6281)" font-weight="bold"><tspan x="0" y="0" class="st4 st5 st6">C</tspan><tspan x="0.7" y="14.4" class="st7 st5 st6">F</tspan><tspan x="2.7" y="28.8" class="st8 st5 st6">I</tspan></text>
 <rect x="0" y="26" class="st3" width="53.4" height="45.7"/>
-<text transform="matrix(1 0 0 1 7.0493 34.6281)"><tspan x="0" y="0" class="st5 st6">data = [</tspan><tspan x="42.3" y="0" class="st4 st5 st6">[</tspan><tspan x="42.3" y="14.4" class="st7 st5 st6">[</tspan><tspan x="42.3" y="28.8" class="st8 st5 st6">[</tspan></text>
```
ldko, line 30: I see the spaces were moved from around the `=` which goes against the Python convention, though the way this is rendering as a png, I don't think you really see a difference between this and where there are spaces are still in place in `data[0, 0] = A`.

fig/python-zero-index.svg
jdkent: I tried to place it there (e.g., `.st5{font-family:'sans-serif';font-weight:'bold';}`) and remove all the calls I was making to `font-weight=bold`, but none of the text was becoming bold...I'm not sure what I'm missing.

fig/python-zero-index.svg
jdkent: thanks for pointing this out!
It may take a little while to get to this since I will need to nudge each element to fit.

Honestly, I would need to figure out how to use some proper svg editing software instead of changing the svg file directly.

If you're alright with the pull request sitting for a while, I'll get to it.

maxim-belkin: > Honestly, I would need to figure out how to use some proper svg editing software instead of changing the svg file directly.

Have you tried using `svgcleaner`?
We have some notes about SVGs here:
https://github.com/swcarpentry/python-novice-inflammation/blob/gh-pages/fig/README.md

fig/python-zero-index.svg
ldko: When I try it without the quotes around bold it works for me. Does `.st5{font-family:'sans-serif';font-weight:bold;}` work for you?

maxim-belkin: you can also use numeric values: 400 for normal, 700 for bold.

ldko:
Hey @maxim-belkin , if @jdkent is unavailable to adjust the spacing on this, do you think it looks ok to merge as-is? or would you be willing to adjust the spacing?

maxim-belkin:
> Hey @maxim-belkin , if @jdkent is unavailable to adjust the spacing on this, do you think it looks ok to merge as-is? or would you be willing to adjust the spacing?

Would something like this work? I changed font to Ubuntu-mono (which all lessons might end up using):
image

ldko:
That looks good to me, Maxim!

maxim-belkin:
updated!

ldko:
Thanks for updating the svg, Maxim. I am thinking based on fig/python-zero-index.png being a file in @jdkent 's PR, we need to update that too. I don't see anything about doing that here: https://github.com/swcarpentry/python-novice-inflammation/blob/gh-pages/fig/README.md. Should that info be included there if it is necessary (I have little knowledge of the svg figure process here)?

maxim-belkin:
I think we can yank that .png file and switch to using .svg in the lesson -- what do you think?

ldko:
If we don't have to have the png, I think it makes sense to use the svg directly.

maxim-belkin:
done!

PR #828, opened 2020-05-28
Update 12-cmdline.md: clarifying goals and some text

jennynakai:
I added text at L24-28 clarifying the goals in the lesson, L85-88 adding detail on the sys library, and L127-128 adding detail on argv. I also suggest explaining the use of 'cat' at L165 to view files vs. editing them. These suggestions are based on points of confusion when I was going through the lesson for the live coding demo.

### DIFF
--- a/_episodes/12-cmdline.md
+++ b/_episodes/12-cmdline.md
@@ -22,7 +22,10 @@ In order to do that,
 we need to make our programs work like other Unix command-line tools.
 For example,
 we may want a program that reads a dataset
-and prints the average inflammation per patient.
+and prints the average inflammation per patient. One of our goals is to produce a program which prints
+the average inflammation per patient in a given file using command line arguments and flags. 
+The second goal is to use shell commands and flags to look at a minimum of the first four lines. 
+A third goal is to find the maximum inflammation in different inflammation files.
 
 > ## Switching to Shell Commands
 >
@@ -79,8 +82,11 @@ We'll tackle these questions in turn below.
 
 ## Command-Line Arguments
 
-Using the text editor of your choice,
-save the following in a text file called `sys_version.py`:
+Using the text editor of your choice, we can
+write a simple script that will tell us what version of
+python we are using. The sys library contains variables and functions
+associated with the interpreter.
+Save the following in a text file called `sys_version.py`:
 
 ~~~
 import sys
@@ -118,6 +124,8 @@ Whenever Python runs a program,
 it takes all of the values given on the command line
 and puts them in the list `sys.argv`
 so that the program can determine what they were.
+Our script name is the first argument after `python` on the
+command line, therefore it is assigned to the variable argv[0].
 If we run this program with no arguments:
 
 ~~~


_episodes/12-cmdline.md
```diff
@@ -22,7 +22,10 @@ In order to do that,
 we need to make our programs work like other Unix command-line tools.
 For example,
 we may want a program that reads a dataset
-and prints the average inflammation per patient.
+and prints the average inflammation per patient. One of our goals is to produce a program which prints
+the average inflammation per patient in a given file using command line arguments and flags. 
+The second goal is to use shell commands and flags to look at a minimum of the first four lines. 
+A third goal is to find the maximum inflammation in different inflammation files.
```
ldko, line 28: I tend to think of these programs/commands listed in the text you provide here as examples that lead us to our goals (rather than the goals themselves) which I consider to be the Objectives listed in the episode Overview. Between what is given in the Objectives section and the text given before each of the examples (i.e. "prints the average inflammation per patient for a given file", "might also want to look at the minimum of the first four lines", "or the maximum inflammations in several files one after another"), how do you see this change expanding on what was there previously?

_episodes/12-cmdline.md
```diff
@@ -79,8 +82,11 @@ We'll tackle these questions in turn below.
 
 ## Command-Line Arguments
 
-Using the text editor of your choice,
-save the following in a text file called `sys_version.py`:
+Using the text editor of your choice, we can
+write a simple script that will tell us what version of
+python we are using. The sys library contains variables and functions
+associated with the interpreter.
+Save the following in a text file called `sys_version.py`:
```
ldko, line 89: I think it makes sense to give a quick explanation of what the sys_version.py script will do as you have added here. I suggest a few edits: removing the word "simple" in case it is demotivating for anyone who does not consider this is simple,  capitalize Python, move the description of `sys` library to where it is talked about a little later and tweak it slightly. What do you think about something like:
```
Using the text editor of your choice, let's write a
short script that tells us what version of Python we are using.
Save the following in a text file called `sys_version.py`:
```
Then after the code:
```
The first line imports a library called `sys`, which is short for “system”.
`sys` provides access to variables and functions associated with the interpreter.
It defines values...
```
```

File: _episodes/12-cmdline.md [hunk 3]
```diff
@@ -118,6 +124,8 @@ Whenever Python runs a program,
 it takes all of the values given on the command line
 and puts them in the list `sys.argv`
 so that the program can determine what they were.
+Our script name is the first argument after `python` on the
+command line, therefore it is assigned to the variable argv[0].
```
ldko, line 128: I am not sure if this adds to the `sys.argv[0]` explanation below. It may be confusing that here it is referring to the script name as an argument, but [`sys.argv`](https://docs.python.org/3/library/sys.html#sys.argv) is explained as the arguments passed to the script (implying that though the path or name is in `sys.argv[0]` it is not considered an argument in the context of arguments we are passing to the script we are calling). Does that make sense?

_episodes/12-cmdline.md
jennynakai: I think my initial confusion came from jumping right into using the command line with flags when the lesson hadn't started yet. It was abrupt and didn't flow, there was no introduction. There may be a better way to address this without restating the goals, but I can't tell you what that is.

_episodes/12-cmdline.md
jennynakai: I agree.

_episodes/12-cmdline.md
jennynakai: You know, I don't really know, but thanks for addressing the comments!

_episodes/12-cmdline.md
ldko: I see what you mean. When taught in a workshop, this lesson usually comes after the Unix Shell lesson where learners see commands with arguments and flags, which maybe makes this less jarring for some? Perhaps to help ease into the appearance of arguments and flags in the episode, we could modify the first paragraph to include something about them by adding to this sentence:

> In order to do that, we need to make our programs work like other Unix command-line tools, allowing us to use command-line arguments and flags to supply values used in the program execution.

or maybe for another possibility, add the second sentence here:
> In order to do that, we need to make our programs work like other Unix command-line tools. Let's look at how we can use command-line arguments and flags to supply values to our programs.

Do you think this would help clarify things somewhat?

PR #829, opened 2020-06-05
Update 10-defensive.md - Checkout

noise42:
Added a more thorough explanation of the AssertionErrors thrown with the last implementation of range_overlap() and the right corrections that should be made (which covers three points: returning None, initializing from data, managing empty inputs)

---

---

### DIFF
--- a/_episodes/10-defensive.md
+++ b/_episodes/10-defensive.md
@@ -462,11 +462,32 @@ We *don't* know whether the other tests passed or failed
 because Python halted the program as soon as it spotted the first error.
 Still,
 some information is better than none,
-and if we trace the behavior of the function with that input,
-we realize that we're initializing `max_left` and `min_right` to 0.0 and 1.0 respectively,
+and if we trace the behavior of the function with that input, we realize that there is nothing in the code that could return None (which is our desired behaviour).
+Moreover, not only the first assertion fails. If we try the other assertions we can see that the fourth line produces another AssertionError, and we must realize that we're initializing `max_left` and `min_right` to 0.0 and 1.0 respectively,
 regardless of the input values.
 This violates another important rule of programming:
 *always initialize from data*.
+The last assertion would be thrown too, because we did not specify any behaviour on empty inputs.
+
+The correct *range_overlap* function shall be therefore:
+~~~
+def range_overlap(ranges):
+    '''Return common overlap among a set of [left, right] ranges.'''
+    #set behaviour for empty inputs
+    if not ranges: return None 
+    
+    #initialize from data - we can safely assume that at this point there is at least one element
+    max_left = ranges[0][0]
+    min_right = ranges[0][1]
+    
+    for (left, right) in ranges:
+        max_left = max(max_left, left)
+        min_right = min(min_right, right)
+    #the returned values should change according to the final results: if the left end is greater than the right end then there is no overlap, thus returning None
+    return (max_left, min_right) if max_left < min_right else None
+~~~
+{: .language-python}
+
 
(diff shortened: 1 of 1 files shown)


_episodes/10-defensive.md
```diff
@@ -462,11 +462,32 @@ We *don't* know whether the other tests passed or failed
 because Python halted the program as soon as it spotted the first error.
 Still,
 some information is better than none,
-and if we trace the behavior of the function with that input,
-we realize that we're initializing `max_left` and `min_right` to 0.0 and 1.0 respectively,
+and if we trace the behavior of the function with that input, we realize that there is nothing in the code that could return None (which is our desired behaviour).
+Moreover, not only the first assertion fails. If we try the other assertions we can see that the fourth line produces another AssertionError, and we must realize that we're initializing `max_left` and `min_right` to 0.0 and 1.0 respectively,
```
ldko, line 466: I suggest tweaking this to:
```
and if we trace the behavior of the function with that input, we realize there is nothing in the
code that could return `None` (which is our desired behavior).
Moreover, not only the first assertion fails. If we try the other assertions, we see that the
fourth line produces another `AssertionError`, leading us to realize that we're initializing
`max_left` and `min_right` to 0.0 and 1.0 respectively,
```

_episodes/10-defensive.md
```diff
@@ -462,11 +462,32 @@ We *don't* know whether the other tests passed or failed
...
 regardless of the input values.
 This violates another important rule of programming:
 *always initialize from data*.
+The last assertion would be thrown too, because we did not specify any behaviour on empty inputs.
```
ldko, line 470: I suggest tweaking this to:
```
The last assertion also fails because we did not specify any behavior in the case of `ranges`
being empty.
```

_episodes/10-defensive.md
```diff
@@ -462,11 +462,32 @@ We *don't* know whether the other tests passed or failed
...
+
+The correct *range_overlap* function shall be therefore:
```
ldko, line 472: Perhaps:
```
Let's correct `range_overlap`:
```

_episodes/10-defensive.md
```diff
@@ -462,11 +462,32 @@ We *don't* know whether the other tests passed or failed
...
+    #initialize from data - we can safely assume that at this point there is at least one element
+    max_left = ranges[0][0]
+    min_right = ranges[0][1]
+    
+    for (left, right) in ranges:
+        max_left = max(max_left, left)
+        min_right = min(min_right, right)
+    #the returned values should change according to the final results: if the left end is greater than the right end then there is no overlap, thus returning None
+    return (max_left, min_right) if max_left < min_right else None
+~~~
```
ldko, line 488: I like that you are providing corrected code and that you added comments to help explain what is happening in the function. I suggest shortening the comments and keeping all of the statements on individual lines to avoid adding anything new that might confuse new learners:
```
def range_overlap(ranges):
    '''Return common overlap among a set of [left, right] ranges.'''
    # Set behavior for empty inputs.
    if not ranges:
        return None

    # Initialize our max and min using the first range.
    max_left = ranges[0][0]
    min_right = ranges[0][1]

    for (left, right) in ranges:
        max_left = max(max_left, left)
        min_right = min(min_right, right)

    if max_left < min_right:
        return (max_left, min_right)
    else:
        # No overlap was found.
        return None
```
I also wonder what you think about adding a re-running of the tests after this new code is given to let learners know to expect no output on successful tests:
```
Now running our test, we expect no output, indicating all of our assertions passed:
~~
test_range_overlap()
~~
{: .language-python}
```

PR #831, opened 2020-06-16
Added missing explanation bit.

Fnyasimi:
I have come across missing information in this line
`Use the step size argument to create a new string that contains only every other character in the string “In an octopus’s garden in the shade”`

In the new correction I have added an the explanation to guide the learner in on what to do to get the desired output.

**New sentence**
Use the step size argument to create a new string that contains only every other `second` character in the string “In an octopus’s garden in the shade”

### DIFF
--- a/_episodes/05-lists.md
+++ b/_episodes/05-lists.md
@@ -441,7 +441,7 @@ last: 4
 > {: .output}
 >
 > Use the step size argument to create a new string
-> that contains only every other character in the string
+> that contains only every other second character in the string
 > "In an octopus's garden in the shade"
 >
 > ~~~
--- a/_episodes/05-lists.md
+++ b/_episodes/05-lists.md
@@ -441,14 +441,19 @@ last: 4
 > {: .output}
 >
 > Use the step size argument to create a new string
-> that contains only every other second character in the string
-> "In an octopus's garden in the shade"
+> that contains only every other character in the string
+> "In an octopus's garden in the shade". Start with
+> creating a variable to hold the string:
 >
 > ~~~
 > beatles = "In an octopus's garden in the shade"
 > ~~~
 > {: .language-python}
 >
+> What slice of `beatles` will produce the
+> following output (i.e., the first character, third
+> character, and every other character through the end
+> of the string)?
 > ~~~
 > I notpssgre ntesae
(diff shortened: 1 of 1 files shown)


_episodes/05-lists.md
```diff
@@ -441,7 +441,7 @@ last: 4
 > {: .output}
 >
 > Use the step size argument to create a new string
-> that contains only every other character in the string
+> that contains only every other second character in the string
```
ldko, line 444: For me, the original explanation is more clear than with the addition of the word `second` because when I read `every other second character` I think perhaps it would be asking for the second character, fourth character, sixth character, etc. instead of the first, third, fifth, etc. that the expected output shows with `I notpssgre ntesae`. Perhaps if you and others find the exercise explanation "every other character in the string" to be unclear, we could work on making it more clear that the output should look like `I notpssgre ntesae` by retaining the first character, third character, fifth character...in some further way.

Fnyasimi: Thanks @ldko for the review feedback I think the second explanation of making it clear how the output should look like will be more understandable to me.

ldko: @Fnyasimi, would you like to modify this PR to do that or would you like me to commit a suggestion on your branch and you can see what you think?

I am thinking the text could change to something like:
```
> Similarly, we can take non-continuous slices of strings.
> Use the step size argument to create a new string
> that contains only every other character in the string
> "In an octopus's garden in the shade". Start with
> creating a variable to hold the string:
>
> ~~~
> beatles = "In an octopus's garden in the shade"
> ~~~
> {: .language-python}
>
> What slice of `beatles` will produce the
> following output (i.e., the first character, third
> character, and every other character through the end
> of the string)?
> ~~~
> I notpssgre ntesae
> ~~~
> {: .output}
```
Do you think this text makes the exercise more clear?

Fnyasimi: @ldko This is more clear and understandable of what is expected to be done

ldko: Would you like to make these changes to the explanation in another commit @Fnyasimi ?

Fnyasimi:
@ldko Kindly review the new changes I have made.

PR #841, opened 2020-07-01
07-cond.md: Introduce `abs`, linking to documentation

### LINKED ISSUE #840: 07-cond.md: `abs` function used without introduction
In the "Close Enough" challenge of the conditionals episode, the solution uses `abs()`, but we have not taught that there is a built-in function `abs` that returns absolute value of a number. I propose to add a hint in the challenge and link to the documentation for the function.

ldko:
This will close #840 by introducing learners to `abs` and hinting they should use it in the solution to the "Close Enough" challenge.


ldko:
Let me know if the latest commit interpreted your suggestions correctly @maxim-belkin , thanks!

maxim-belkin:
Yes, thank you!

PR #842, opened 2020-07-02
Arithmetic on the Command Line for 12-cmdline.md suggestions

rshpeley:
I think maybe a few clarifications could help this section.

1. The operators `add` and `subtract` maybe should have dashes like the actions in the previous example ==> `--add` and `--subtract`. This is just a matter of form, but may be more intuitive for some students. If so the operators would change in the problem and solution sections.
2. In the problem section the program output shows an expected integer result rather than the float cast used in the solution. The problem expected outputs should show float values to agree with the solution.
3. Giving a hint to the student that the solution may also include multiplication and division could be an incentive to try these operators as well.

Of these, I think (2.) should be changed with the others being optional.

<details>

\- The Carpentries
</details>

### DIFF
--- a/_episodes/12-cmdline.md
+++ b/_episodes/12-cmdline.md
@@ -666,25 +666,25 @@ the program now does everything we set out to do.
 
 > ## Arithmetic on the Command Line
 >
-> Write a command-line program that does addition and subtraction:
+> Write a command-line program that does addition and subtraction (you could also try multiplication and division operators):
 >
 > ~~~
-> $ python arith.py add 1 2
+> $ python arith.py --add 1 2
 > ~~~
 > {: .language-bash}
 >
 > ~~~
-> 3
+> 3.0
 > ~~~
 > {: .output}
 >
 > ~~~
-> $ python arith.py subtract 3 4
+> $ python arith.py --subtract 3 4
 > ~~~
 > {: .language-bash}
 >
 > ~~~
-> -1
+> -1.0
 > ~~~
 > {: .output}
 >
@@ -696,7 +696,7 @@ the program now does everything we set out to do.
 > >     assert len(sys.argv) == 4, 'Need exactly 3 arguments'
 > >
 > >     operator = sys.argv[1]
-> >     assert operator in ['add', 'subtract', 'multiply', 'divide'], \
+> >     assert operator in ['--add', '--subtract', '--multiply', '--divide'], \
 > >         'Operator is not one of add, subtract, multiply, or divide: bailing out'
 > >     try:
 > >         operand1, operand2 = float(sys.argv[2]), float(sys.argv[3])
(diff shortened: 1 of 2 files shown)


rshpeley:
Ok. I missed this earlier. The assert operator would need its error text changed as well.

`assert operator in ['--add', '--subtract', '--multiply', '--divide'], \
        'Operator is not one of --add, --subtract, --multiply, or --divide: bailing out'`

_episodes/12-cmdline.md
```diff
@@ -666,25 +666,25 @@ the program now does everything we set out to do.
 
 > ## Arithmetic on the Command Line
 >
-> Write a command-line program that does addition and subtraction:
+> Write a command-line program that does addition and subtraction (you could also try multiplication and division operators):
```
ldko, line 669: Since the solution contains multiplication and division, I suggest we include it along with addition and subtraction in the main challenge description. Also, since the arithmetic only works with 2 numbers, we should probably specify that as well.
```suggestion
> Write a command-line program that performs addition, subtraction,
 multiplication, or division with two numbers:
```

_episodes/12-cmdline.md
```diff
@@ -696,7 +696,7 @@ the program now does everything we set out to do.
 > >     assert len(sys.argv) == 4, 'Need exactly 3 arguments'
 > >
 > >     operator = sys.argv[1]
-> >     assert operator in ['add', 'subtract', 'multiply', 'divide'], \
+> >     assert operator in ['--add', '--subtract', '--multiply', '--divide'], \
 > >         'Operator is not one of add, subtract, multiply, or divide: bailing out'
```
ldko, line 700: To prevent confusion, do you think we should have the dashes in the error message as well?
```suggestion
> >         'Operator is not one of --add, --subtract, --multiply, or --divide: bailing out'
```

_episodes/12-cmdline.md
maxim-belkin: I agree with Lauren's suggestion to improve this change and I'd like to propose further improvement:
```suggestion
> Write a Python program that adds, subtracts, multiplies, or divides two numbers provided on the command line:
```

maxim-belkin:
Hi all,

Thank you, Raymond, for suggesting improvements! I think this part of the lesson rarely gets the same amount of attention as some other parts of the lesson.

Now, regarding the proposed changes. I'm not sure if adding `--` to program arguments would necessarily improve this part of the lesson because someone could argue that we'd better modify the other examples and remove double dashes or. But I'm OK with this change so I'll leave it to @ldko  and @annefou to decide. If the decision is made to proceed with the changes, we would then also need to update `arith.py` file in `code` folder accordingly.

Again, thank you for your contributions, Raymond! 🥇

_episodes/12-cmdline.md
ldko: I like the version by @maxim-belkin .

PR #844, opened 2020-07-03
Detailed alt-text for images in lesson

magsol:
These are some initial attempts at providing detailed alt-text titles and descriptions for some of the images in the lesson. I wanted to stop with three as a check in case I'm doing this incorrectly, but I'm more than happy to add descriptions to all the remaining images from #781 . Thank you! Happy for your feedback.

### DIFF
--- a/_episodes/04-loop.md
+++ b/_episodes/04-loop.md
@@ -18,7 +18,7 @@ keypoints:
 In the last episode, we wrote Python code that plots values of interest from our first
 inflammation dataset (`inflammation-01.csv`), which revealed some suspicious features in it.
 
-![Analysis of inflammation-01.csv](../fig/03-loop_2_0.png)
+![Three panels showing different properties of the data in inflammation-01.csv, as uncovered in the last lesson. First panel is the mean of the data over time, second panel is the maximum of the data over time, and the third panel is the minimum of the data over time.](../fig/03-loop_2_0.png)
 
 We have a dozen data sets right now, though, and more on the way.
 We want to create plots for all of our data sets with a single statement.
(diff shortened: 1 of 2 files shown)


_episodes/04-loop.md
```diff
@@ -18,7 +18,7 @@ keypoints:
 In the last episode, we wrote Python code that plots values of interest from our first
 inflammation dataset (`inflammation-01.csv`), which revealed some suspicious features in it.
 
-![Analysis of inflammation-01.csv](../fig/03-loop_2_0.png)
+![Three panels showing different properties of the data in inflammation-01.csv, as uncovered in the last lesson. First panel is the mean of the data over time, second panel is the maximum of the data over time, and the third panel is the minimum of the data over time.](../fig/03-loop_2_0.png)
```
ldko, line 21: I like how you have pretty succinctly covered with text what is a lot of information shown. I think using "plot" instead of "panel" would be more specific. I don't think we need to mention the "uncovered in the last lesson" part because it is  mentioned in the text above the image. I wonder if we should describe the data briefly, since the text only vaguely mentions "suspicious features". What do you think about:
```suggestion
![Three plots of the inflammation-01.csv data. The first plot, a jagged rise and fall, is the mean inflammation across all patients for each day, the second plot, a smooth and even rise then fall, is the maximum inflammation each day, and the third plot, seeming to show a step function, is the minimum inflammation each day.](../fig/03-loop_2_0.png)
```

ldko:
The alt text was completed in #922. Thanks for submitting this @magsol !

ldko:
Thanks, Shannon! I think we actually got through all of the figures listed on our lesson for the issue. We'd be happy to see other contributions from you in the future though...after the semester ends. :)

Issue #830, opened 2020-06-08
pyplot import


RMeli:
In the episode [Visualizing Tabular Data](https://swcarpentry.github.io/python-novice-inflammation/03-matplotlib/index.html), `pyplot` is imported as follows:

```
import matplotlib.pyplot
```

This import requires to use `matplotlib.pyplot` everywhere in the code, which is much longer than the usual convention `plt`. I suggest to change the import to

```
import matplotlib.pyplot as plt
```
or (see Issue #665)
```
from matplotlib import pyplot as plt
```

and to update the code accordingly. For example the first cell would become
```
from matplotlib import pyplot as plt
image = plt.imshow(data)
plt.show()
```

which is much easier to read (the emphasis is on the function being called and not `matplotlib.pyplot`) and introduces the `plt` name which is pretty standard.

ldko:
Hi @RMeli , thank you for taking time to contribute this suggestion. You are right that using `as plt` would be less to type and follows a common practice.

Using `as` to supply a name for an imported module is something that has come up in discussion before, enough that it has found a place on our short list of ["Frequently Argued Issues"](https://swcarpentry.github.io/python-novice-inflammation/guide/). This links to a discussion that happened on this topic in pull request #61. The decision was to keep the import as `import matplotlib.pyplot` to keep from overloading learners new to programming with an extra concept. It is left up to an instructor teaching the lesson to decide if they feel teaching the `as` syntax would be right for their audience. There is a callout, entitled "Importing libraries with shortcuts",  in the [Analyzing Patient Data episode](https://swcarpentry.github.io/python-novice-inflammation/02-numpy/index.html) that mentions this style of import with `import numpy as np` for those who want to apply it.

RMeli:
@ldko Thanks for pointing out #61 and the "Frequently Argued Issues" , which I obviously missed. 

I think the cognitive load of "you can use `as` to rename your import, which is useful for shortcuts" is very low compared to the benefits; my vote would have contribute to a 50/50 split on the matter! ;)

I'm happy to close the issue following #61.

ldko:
> I think the cognitive load of "you can use as to rename your import, which is useful for shortcuts" is very low compared to the benefits; my vote would have contribute to a 50/50 split on the matter! ;)

[...]

PR #845, opened 2020-07-09
Replaced cumbersome matplotlib.pyplot with plt

jakekara:
I think that it's easier to write and read the sample code in this lesson if we use the common "plt" nickname for matplotlib.pyplot. The previous lesson shows how to use "import ... as" syntax.

### DIFF
--- a/_episodes/03-matplotlib.md
+++ b/_episodes/03-matplotlib.md
@@ -21,9 +21,9 @@ import the `pyplot` module from `matplotlib` and use two of its functions to cre
 heat map of our data:
 
 ~~~
-import matplotlib.pyplot
-image = matplotlib.pyplot.imshow(data)
-matplotlib.pyplot.show()
+from matplotlib import pyplot as plt
+image = plt.imshow(data)
+plt.show()
 ~~~
 {: .language-python}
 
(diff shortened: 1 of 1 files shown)


ldko:
Hi @jakekara ,
Thank you for taking the time to contribute to the python-novice-inflammation lesson.

`matplotlib.pylot` vs. `plt` is not new a new issue to us. I wonder if you have taken a look at the discussion and links from #830 about this issue of "import...as" and might like to comment there. At one point the lesson did use "import...as" syntax, but after a debate on PR #61, it was removed from the body of the lesson and put into a callout box that you reference. This was partly due to wanting to keep new concepts minimal for the workshop learners who are new to programming and "import...as" not being as common outside of the `numpy` and `matplotlib.pyplot` world.

maxim-belkin:
Hi Jake,

Thanks for the suggestion. As Lauren has already explained, this is one of those simplifications that might be, in fact, confusing to novices. Both `import as` and `from ... import ... as ...` notations might seem easy to grasp to you, me, Lauren, but they also can be confusing... to the point that you kept `import plt` in a few places in this PR. Imagine what it can do to those who interact with Python and programming for the first time. Provided that our goal is to teach programming to novices, we need to keep the amount of new information to an absolute minimum. Keeping `import matplotlib.pyplot` and using methods as `matplotlib.pyplot.method-name` should make a clear connection between the two lines. I hope this explains why we don't practice `import ... as ...` and `from ... import ... as ...` in the lesson.

Please let us know if you'd like to discuss this further of if you have suggestions on how to improve the point we make in the callout box in the first episode.

Again, thanks for the suggestion!

jakekara:
Thanks for the feedback. I should have checked the issues first. Was in a hurry to get checkout credit for contributing a suggestion.

I still feel the benefit of shorter lines throughout the code, with less syntax to digest on each line, outweighs the benefit of not having to explain `import as`, but I will join the existing discussion on this topic!

Issue #854, opened 2020-07-16
Consistency in use of single- and double-quote marks


troub1:
In the first episode, "[Python Fundamentals](https://swcarpentry.github.io/python-novice-inflammation/01-intro/index.html)," it's mentioned that 

> And to create a string, we add single or double quotes around some text

and I often take a second there to say that the main thing is to be consistent with which one you choose. And then the workshop proceeds and we're using single-quotes the whole time. _Except_ later, in ["Storing Multiple Values in Lists"](https://swcarpentry.github.io/python-novice-inflammation/05-lists/index.html), we start switching to double-quotes about 2/3 through:  

    binomial_name = "Drosophila melanogaster"
    ... 

for the rest of the episode, and then back and forth a bit in subsequent episodes, like in "Errors and Exceptions":  

```
def some_function()  
    msg = "hello, world!"  
    print(msg)  
    return msg
```

and  

    letters = ['a', 'b', 'c']  
    print("Letter #1 is", letters[0])

The latter one seems to follow the style of literals being double-quoted and lists/dicts/etc being single-quoted, but it's different from all previous instances I can see.

It's a small thing, especially since in the very beginning we do say that either can be used, but I would say it increases cognitive load on both the learners as well as less experienced instructors because the sudden change _is_ noticed by the learners if the instructor types it as written, and otherwise the instructor has to substitute on the fly when live-coding the example or displaying the exercise. I think it often prompts a question "Why are we doing double-quotes here? Is it significant?" to which I don't really have a good answer except that the lessons are made up from contributions of a large community, and you yourself should try to be consistent 😄 

So I'd be thrilled to get this fixed, or to fix it. I assume the SWC style, if there is one, is to single-quote everything since that's how the vast majority of everything is written? I'm happy to do a pull request and fix these up, if it should be done.

Thanks!
Steven

ldko:
Hi Steven,
Yes, we should be consistent with single vs. double quotes, and in the lesson, where we are inconsistent is not intentional. Like you suggest, it likely came from several people contributing small chunks of episodes over a long period of time. I think it makes sense to consistently use single quotes, and it would be great if you want to do pull requests to fix this issue. I suggest one pull request per episode, for easier review by maintainers. The only place I think we should be using double quotes is for triple double quotes,such as in a docstring, or in the case where the string contains a single quote (not sure this comes up in the lesson). Thank you very much for bringing up the issue and offering to fix it! --Lauren

[...]

PR #855, opened 2020-07-17
Update episode 5 double-quotes to single-quotes

troub1:
to improve consistency in style. There are two instances '"-1"' and "octopus's garden" left to preserve the quote-in-quote.

### DIFF
--- a/_episodes/05-lists.md
+++ b/_episodes/05-lists.md
@@ -377,7 +377,7 @@ last: 4
 > {: .language-python}
 >
 > ~~~
-> 2013
+> '2013'
 > [['chlorine', 'Cl'], ['bromine', 'Br'], ['iodine', 'I'], ['astatine', 'At']]
 > ~~~
 > {: .output}
--- a/_episodes/05-lists.md
+++ b/_episodes/05-lists.md
@@ -311,7 +311,7 @@ odds: [1, 3, 5, 7]
 > Use a for-loop to convert the string "hello" into a list of letters:
 >
 > ~~~
-> ["h", "e", "l", "l", "o"]
+> ['h', 'e', 'l', 'l', 'o']
 > ~~~
 > {: .language-python}
 >
@@ -325,7 +325,7 @@ odds: [1, 3, 5, 7]
 > > ## Solution
 > > ~~~
 > > my_list = []
-> > for char in "hello":
+> > for char in 'hello':
 > >     my_list.append(char)
 > > print(my_list)
 > > ~~~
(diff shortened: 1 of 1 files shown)


troub1:
#854

_episodes/05-lists.md
```diff
@@ -367,18 +367,18 @@ last: 4
...
+>                     ['bromine', 'Br'],
+>                     ['iodine', 'I'],
+>                     ['astatine', 'At']]
 > ~~~
 > {: .language-python}
 >
 > ~~~
-> "2013"
-> [["chlorine", "Cl"], ["bromine", "Br"], ["iodine", "I"], ["astatine", "At"]]
+> 2013
```
ldko, line 380: I believe this should have single quotes around it:
```suggestion
> '2013'
```

_episodes/05-lists.md
```diff
@@ -501,20 +501,20 @@ And similarly, you can omit the ending index in the range to take a slice to the
...
-print("Omitting ending index:", sond)
+print('Omitting ending index:', sond)
 ~~~
 {: .language-python}
 
 ~~~
-With known last position: ["sep", "oct", "nov", "dec"]
-Using len() to get last entry: ["sep", "oct", "nov", "dec"]
-Omitting ending index: ["sep", "oct", "nov", "dec"]
+With known last position: ['sep", 'oct', 'nov', 'dec']
```
ldko, line 515: Missed one double quote:
```suggestion
With known last position: ['sep', 'oct', 'nov', 'dec']
```

_episodes/05-lists.md
troub1: I believe this line is the representation of the command output...when I tested this in my particular Jupyter notebook at least, this output from print didn't have quotes around it (it usually wouldn't, would it?)

maxim-belkin: Unless we change the solution section below, output should have single quotes around it because we're not printing here. 
image

troub1: Ah, that's right, I see. I also remember now when I was running through this I noticed something interesting. It didn't even work for me as given. It might be a Jupyter-related particular, and might work differently from another interpreter. But mine looks like this when I try to output both lists as given in the example:
image
If each is run separately they'll both output, but called in the same cell only the last one seems to show up (unless I changed it to use print).

maxim-belkin: > If each is run separately they'll both output, but called in the same cell only the last one seems to show up (unless I changed it to use print).

Having taught our lesson last week, I'm very tempted to propose to create a "Jupyter Gotchas" section in our Instructor Guide to warn instructors of all the gotchas they might face when they teach using Jupyter Notebooks.

PR #865, opened 2020-09-01
Add int, float and string functions to intro

robelgeda:
The motivation for this change is to provide a method to convert between different types of python objects. in this PR I add the functions for each basic alpha-numeric type and give an example of how to use those functions to convert between float -> int.

### DIFF
--- a/_episodes/01-intro.md
+++ b/_episodes/01-intro.md
@@ -73,6 +73,13 @@ weight_kg_text = 'weight in kilograms:'
 ~~~
 {: .language-python}
 
+To convert the weight from float to an integer we use the int function, the same way we use the print statement, as follows:
+
+~~~
+weight_kg_int = int(weight_kg)
+~~~
+{: .language-python}
+
 ## Using Variables in Python
 To display the value of a variable to the screen in Python, we can use the `print` function:
 
--- a/_episodes/01-intro.md
+++ b/_episodes/01-intro.md
@@ -54,9 +54,9 @@ This means that, for example:
 ## Types of data
 Python knows various types of data. Three common ones are:
 
-* integer numbers
-* floating point numbers, and
-* strings.
+* integer numbers (`int`)
+* floating point numbers (`float`), and
+* strings (`str`).
 
 In the example above, variable `weight_kg` has an integer value of `60`.
 To create a variable with a floating point value, we can execute:


_episodes/01-intro.md
```diff
@@ -54,9 +54,9 @@ This means that, for example:
 ## Types of data
 Python knows various types of data. Three common ones are:
 
-* integer numbers
-* floating point numbers, and
-* strings.
+* integer numbers (`int`)
+* floating point numbers (`float`), and
+* strings (`str`).
```
ldko, line 59: I like this addition of the Python names for the types.

_episodes/01-intro.md
```diff
@@ -73,6 +73,13 @@ weight_kg_text = 'weight in kilograms:'
 ~~~
 {: .language-python}
 
+To convert the weight from float to an integer we use the int function, the same way we use the print statement, as follows:
+
```
ldko, line 77: @maxim-belkin what do you think about this proposed introduction of type conversion here? I know we are trying to keep from introducing too much new content that does not directly affect the continuation of the lesson due to time constraints and too much info for a workshop. This is a rather small addition though.

If we do want to include it, we should have backticks around `float` and `int`. Also, this is just before we mention `print`, so I think it would make sense to leave that reference out. Since this would  be the first mention of "function", we could link to the reference at https://swcarpentry.github.io/python-novice-inflammation/reference/#function.

ldko (commented):
Thank you for proposing this addition @robelgeda ! Lets see if other maintainers @maxim-belkin @annefou or anyone else has any comments on this change to the introductory episode.

_episodes/01-intro.md
annefou: Yes. Actually it is a very good idea. It is often not very obvious for novices to find out that string is `str`.

_episodes/01-intro.md
annefou: I would not introduce type conversion at this stage. In this episode, we mostly focus on learning about variables and how to define/print them. I don't find it necessary to discuss about type conversion.

_episodes/01-intro.md
maxim-belkin: I don't mind introducing `int`, `float`, and `str`, but we should provide some narrative that instructors could use when talking about them -- putting them in parentheses leaves this explanation up to an individual instructor. I think that it'll be easier to talk about them right after this list ... where we could also introduce the `type` function but this might be out of scope of this PR.

_episodes/01-intro.md
maxim-belkin: I concur with Anne -- I don't see how we can plug type conversion into this episode (it's too early and there is no real need for that here). We could do that in the episode where we deal with command line arguments if we needed to convert user input to `int` or `float`.

robelgeda:
It looks like this stage of the introduction is too early to mention python type casting. Please let me know if there is a better place for this or if by then people would have found out, this PR may be closed as it presents redundant info.

Issue #867, opened 2020-09-18
First introduction of comments


zerpiko:
Hi. 

I noticed that in lesson 2 (**Analyzing Patient Data**) the use of **#** to comment out text is explained with:

>Everything in a line of code following the ‘#’ symbol is a comment that is ignored by Python. Comments allow programmers to leave explanatory notes for other programmers or their future selves.

after using a comment in an example. However, the first use of comments is in lesson 1 (**Python Fundamentals**)
```
# There are 2.2 pounds per kilogram
weight_lb = 2.2 * weight_kg
print(weight_kg_text, weight_kg, 'and in pounds:', weight_lb)
```

Wouldn't be better to move the explanation (and associated key point) to the first lesson?

maxim-belkin:
Good catch, @zerpiko! I think we should do that.
Do you agree, @ldko and @annefou?

ldko:
Yes, I think @zerpiko makes a good point. If Jose wants to open a PR to move the explanation, I think that would be great!

annefou:
Yes I agree,

zerpiko:
Hi, thanks for the feedback. I created a PR for this issue.

PR #868, opened 2020-09-21
moved comment description to first lesson

josemunozc:
In relation to issue #867 

I just moved the description verbatim (and associated key point) to the first lesson and put it after the python example. Another option could be to put the description in brackets right before the first use of **#**:

> This means that assigning a value to one variable does not change values of other variables. For example, let’s store the subject’s weight in pounds in its own variable (everything in a line of code following the ‘#’ symbol is a comment that is ignored by Python, comments allow programmers to leave explanatory notes for other programmers or their future selves):

But might be too lengthy?


ldko (commented):
Hi Jose,
This takes care of the issue you brought up (thank you!), and I think it could be merged as-is if @maxim-belkin or @annefou approve.

Looking at the change in the full context of the lesson now, since we bring up first introductions in the issue, I wonder if it would make sense to move the comment "# There are 2.2 pounds per kilogram" to the first time we do the math and put it higher in the episode in line after `print('weight in pounds:', 2.2 * weight_kg)`:
```
print('weight in pounds:', 2.2 * weight_kg)  # There are 2.2 pounds per kilogram
```
and move the explanation of comments under that. What do you think @zerpiko @maxim-belkin @annefou

PR #869, opened 2020-09-21
Update 07-cond.md

gerdos:
Added two challenge exercises that my students usually find fun to think about.

<details>

\- The Carpentries
</details>

### DIFF
--- a/_episodes/07-cond.md
+++ b/_episodes/07-cond.md
@@ -490,6 +490,36 @@ freeing us from having to manually examine every plot for features we've seen be
 > {: .solution}
 {: .challenge}
 
+> ## Math vs Programming
+>
+> We all know that 3>2>1 is True. Lets try (3>2)>1. What happens here?
+>
+> > ## Solution
+> > ~~~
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Very interesting challenges, @gerdos!

What do you think, @ldko and @annefou? I see a couple of grammatical and stylistic issues... but other than that -- I like both challenges.

annefou:
Ineed very interesting. I like the first challenge. For the second part, I wonder if it is worth to have is as a challenge (I don't think this is something a novice can easily find the solution); maybe rather have it as a callout.

ldko:
I would be ok with including either or both challenges. Regarding what @annefou says about the second challenge, I wonder if modifying it to be a multiple choice type of question that has multiple correct answers and explanations in the solution would make it more suitable.

annefou:
> I would be ok with including either or both challenges. Regarding what @annefou says about the second challenge, I wonder if modifying it to be a multiple choice type of question that has multiple correct answers and explanations in the solution would make it more suitable.

@ldko I like your suggestion!

PR #870, opened 2020-09-23
Fix exercise with f2k

unode:
Is missing a `print()` so output was not shown

### DIFF
--- a/_episodes/08-func.md
+++ b/_episodes/08-func.md
@@ -814,9 +814,9 @@ readable code!
 >     k = ((f-32)*(5.0/9.0)) + 273.15
 >     return k
 >
-> f2k(8)
-> f2k(41)
-> f2k(32)
+> print(f2k(8))
+> print(f2k(41))
+> print(f2k(32))
 >
 > print(k)
 > ~~~


ldko (approved):
Hi Renato,
Thanks for fixing this inconsistency. Your change aligns the code and the solution.

maxim-belkin:
Thanks for fixing this inconsistency, Renato, and thanks for merging, Lauren!

I'm not sure that wrapping `f2k` calls in `print()` here instead of adjusting the output is the best solution here. From the standpoint of the goal of the exercise, `print()` functions add an additional layer of cognitive load. I also wonder if a single call to `f2k` is sufficient to demonstrate the point. What do you think?

Apologies for the late comment.

ldko:
I don't think the `print` calls add much cognitive load as by episode 8, learners have seen/used `print` a lot and the `print` code is given to them. I initially also wondered if the output was meant to be only 0, but I think since this is the functions lesson, seeing the multiple function calls and different outputs is a reassurance of what is happening when we call the function and its non-affect on the `k` outside the function.

maxim-belkin:
> I don't think the `print` calls add much cognitive load as by episode 8, learners have seen/used `print` a lot and the `print` code is given to them.

The fact that learners have seen the `print()` function doesn't mean it doesn't add an extra load -- it does for two reasons:
1. explanation gains an extra step: "we call f2k function to compute... >>and then display the result using print function<<"
2. Learners have to understand that `print` function doesn't alter / conceal changes to `k` either.

I sort of agree with adding `print`s, but I think that we then need to remove one or two `print(f2k(...))`  lines to reduce the load.

> I initially also wondered if the output was meant to be only 0

The output was from the standard CPython interpreter so it was correct from that point of view. In Jupyter notebook, people would see only 0, though. Adding `print()` calls made output consistent in all interpreters/IDEs.

ldko:
I think removing one or two `print(f2k(...))` is sensible.

PR #872, opened 2020-09-25
08-func.md: correct answer in an exercise

maxim-belkin:
Not sure how this stayed unnoticed for so long :)

```
>>> def f2k(f):
...  k = ((f-32)*(5/9)) + 273.15
...  return k
...
>>> print(f2k(41))
278.15
```


ldko (approved):
Oops! Good catch. Thanks for fixing this, Maxim!

annefou (approved):

PR #879, opened 2020-10-12
whitespace inside f2k function 

wiringa:
Whitespace for readability


maxim-belkin (approved):
Thank you, Peter!
Makes sense to me. This is one of those changes that I keep forgetting about after every workshop.

ldko (approved):
Thanks, @wiringa !

Issue #880, opened 2020-10-13
Broken figures on github.io


chrisroat:
At the following page, the images under "Option A: Jupyter Notebook" are broken links:

https://swcarpentry.github.io/python-novice-inflammation/setup.html

I see the figures in the repo, but I am not familiar enough with gh-pages and github.io to know what went wrong.

maxim-belkin:
Thanks for the report, Chris.
François fixed this in #881.

PR #881, opened 2020-10-13
Fix image paths for setup

fmichonneau:
This should take care of addressing the broken paths for #880


maxim-belkin (approved):
Thanks, François!

Issue #853, opened 2020-07-16
Making choices - Add explanation of all comparison operators


jakekara:
In the SWC > Programming With Python > Making Choices Lesson (https://swcarpentry.github.io/python-novice-inflammation/), there is no explanation of the "<=" operator, which is included in the solution to the challenge where students are asked to determine if a is within 10% of b.

This lesson explains the == operator, but it might also be a good lesson to introduce all of the comparison operators (https://docs.python.org/3/reference/expressions.html#value-comparisons).

We already have this example:

```
num = -3

if num > 0:
    print(num, 'is positive')
elif num == 0:
    print(num, 'is zero')
else:
    print(num, 'is negative')
```

Perhaps something like this would be a good follow-on in the same style.

```
num = 0
if num >= 0:
   print(num, "is not negative")
if num <= 0:
   print(num, "is not positive")
```

ldko:
Hi Jake,
You have made a good point about the lack of explanation of different operators available for value comparisons in the [Making Choices](https://swcarpentry.github.io/python-novice-inflammation/07-cond/index.html) episode.

Perhaps we could add a sentence (or a callout box) after the sentence "Note that to test for equality we use a double equals sign == rather than a single equals sign = which is used to assign values" such as:
In addition to the `>` and `==` operators we have seen so far, other operators available for comparing the values of two objects include `<` (less than), `<=` (less than or equal), `>=` (greater than or equal), and `!=` (not equal).

We could then add your suggested example or perhaps change this example (rather than add another example which would add more time to teaching the episode):
```
if (1 < 0) or (-1 < 0):
    print('at least one test is true')
```
to
```
if (1 < 0) or (1 >= 0):
    print('at least one test is true')
```
What do you think @maxim-belkin ?

Thanks for opening this issue and contributing to the lesson, Jake!

maxim-belkin:
> What do you think @maxim-belkin ?

Thanks for pinging me, Lauren. I think this would be a very fine addition to the lesson, especially for operators made of two symbols (`==`, `!=`, `>=`, and `<=`).

Thank you, Jake, for the suggestion!

stefpiatek:
I'd be happy to create a PR with that if you'd like?

stefpiatek:
Oh this is also very delayed (sorry for bringing it up again @ldko) but I was wondering if you'd be happy to add the `hacktoberfest-accepted` tag for this PR? Making a late attempt to get the 4 PRs

PR #882, opened 2020-10-17
Add explanation of comparison operators

stefpiatek:
Added quick explanation of comparison operators, didn't want to bloat the message so hopefully it's the right level of detail - happy to change anything if you'd like. 

Closes swcarpentry/python-novice-inflammation#853

### DIFF
--- a/_episodes/07-cond.md
+++ b/_episodes/07-cond.md
@@ -98,9 +98,10 @@ else:
 Note that to test for equality we use a double equals sign `==`
 rather than a single equals sign `=` which is used to assign values.
 
-> ## Comparing in python
+> ## Comparing in Python
 >
-> We've already seen a few of these already, but there are a few more options that you should know about:
+> Along with the `>` and `==` operators we have already used for comparing values in our conditionals, 
+> there are a few more options to know about:
 >
 > - `>`: greater than
 > - `<`: less than
--- a/_episodes/07-cond.md
+++ b/_episodes/07-cond.md
@@ -98,6 +98,18 @@ else:
 Note that to test for equality we use a double equals sign `==`
 rather than a single equals sign `=` which is used to assign values.
 
+> ## Comparing in python
+>
+> We've already seen a few of these already, but there are a few more options that you should know about:
+>
+> - `>`: greater than
+> - `<`: less than
+> - `==`: equal to
+> - `!=`: does not equal
+> - `>=`: greater than or equal to
+> - `<=`: less than or equal to
+{: .callout}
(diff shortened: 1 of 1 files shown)


_episodes/07-cond.md
```diff
@@ -98,6 +98,18 @@ else:
 Note that to test for equality we use a double equals sign `==`
 rather than a single equals sign `=` which is used to assign values.
 
+> ## Comparing in python
```
ldko, line 101: For consistency, we should capitalize "Python".

_episodes/07-cond.md
```diff
@@ -98,6 +98,18 @@ else:
...
+>
+> We've already seen a few of these already, but there are a few more options that you should know about:
```
ldko, line 103: The above sentence uses "already" a couple times in the first clause, what do you think about changing the wording slightly to something like:
"Along with the  `>` and `==` operators we have already used for comparing values in our conditionals, there are a few more options to know about:"

ldko (commented):
Thanks for submitting this PR, Stefan! I think it looks good, but have a few minor suggestions. In addition to the inline comments, what do you think about changing the example of:
```
if (1 < 0) or (-1 < 0):
    print('at least one test is true')
```
to
```
if (1 < 0) or (1 >= 0):
    print('at least one test is true')
```
to show use of a two symbol operator @stefpiatek @maxim-belkin ?

_episodes/07-cond.md
maxim-belkin: Should we say what we're comparing here (numbers)? Or should we leave it vague because later they might learn about `"a" < "b"` ?

ldko: I like keeping it general, since the operators can be used with other types than numbers as you point out.

stefpiatek:
Took a while to set aside the time but those should be addressed - let me know if there's anything else you'd like changed

maxim-belkin (approved):
Thank you, @stefpiatek! Looks good to me!

ldko (approved):
This looks good to me. Thank you, Stefan!

stefpiatek:
Oh this is also very delayed (sorry for bringing you back to a closed PR @ldko) but I was wondering if you'd be happy to add the `hacktoberfest-accepted tag` for this PR? Making a late attempt to get the 4 PRs

ldko:
Hi @stefpiatek, I have labeled the PR with  `hacktoberfest-accepted`. Hopefully this gets you closer to your quota! :shirt:

PR #885, opened 2020-10-29
Fix errors in text to code line references.

jsteyn:
<details>
In the text, describing the error that occurs (for the very first section of code in the Episode) reference is made to lines 8 and 6. This should be line 11 and line 9.

**The actual original text: (error shown in bold)**
 1. The first shows code from the cell above, with an arrow pointing to Line **8** (which is ```favorite_ice_cream()```).
 1. The second shows some code in the function favorite_ice_cream, with an arrow pointing to Line **6** (which is ```print(ice_creams[3])```).

**It should read: (correction shown in bold)**
 1. The first shows code from the cell above, with an arrow pointing to Line **11** (which is ```favorite_ice_cream()```).
 1. The second shows some code in the function favorite_ice_cream, with an arrow pointing to Line **9** (which is ```print(ice_creams[3])```).
</details>

### DIFF
--- a/_episodes/09-errors.md
+++ b/_episodes/09-errors.md
@@ -80,10 +80,10 @@ You can determine the number of levels by looking for the number of arrows on th
 In this case:
 
 1.  The first shows code from the cell above,
-    with an arrow pointing to Line 8 (which is `favorite_ice_cream()`).
+    with an arrow pointing to Line 11 (which is `favorite_ice_cream()`).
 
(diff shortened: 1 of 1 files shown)


ldko (approved):
Thank you for fixing this discrepancy, Jannetta. Looks good to me!

maxim-belkin (approved):
Thank you, Jannetta!

PR #888, opened 2020-11-16
Added an image description for linflammation-01-imshow.svg in 03-matplotlib.md

grossular:
Related to issue #781

<details>
Extended description for Heatmap image in 03-matplotlib.md
</details>

### DIFF
--- a/_episodes/03-matplotlib.md
+++ b/_episodes/03-matplotlib.md
@@ -27,7 +27,7 @@ matplotlib.pyplot.show()
 ~~~
 {: .language-python}
 
-![Heatmap representing the 'data' variable matrix. Each cell is coloured by value along a colour gradient from blue to yellow.](../fig/inflammation-01-imshow.svg)
+![Heat map representing the 'data' variable matrix. Each cell is colored by value along a color gradient from blue to yellow.](../fig/inflammation-01-imshow.svg)
 
 Blue pixels in this heat map represent low values, while yellow pixels represent high values.  As we
 can see, inflammation rises and falls over a 40-day period.  Let's take a look at the average inflammation over time:
--- a/_episodes/03-matplotlib.md
+++ b/_episodes/03-matplotlib.md
@@ -27,7 +27,7 @@ matplotlib.pyplot.show()
 ~~~
 {: .language-python}
 
-![Heat map representing the 'data' variable matrix. Each cell is colored by value along a color gradient from blue to yellow.](../fig/inflammation-01-imshow.svg)
+![Heat map representing the `data` variable. Each cell is colored by value along a color gradient from blue to yellow.](../fig/inflammation-01-imshow.svg)
 
 Blue pixels in this heat map represent low values, while yellow pixels represent high values.  As we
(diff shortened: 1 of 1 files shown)


ldko (approved):
Thanks for adding to the description, John. I think this is a good amount of detail in context of the text that precedes and follows the image.

_episodes/03-matplotlib.md
```diff
@@ -27,7 +27,7 @@ matplotlib.pyplot.show()
 ~~~
 {: .language-python}
 
-![Heatmap of the Data](../fig/inflammation-01-imshow.svg)
+![Heatmap representing the 'data' variable matrix. Each cell is coloured by value along a colour gradient from blue to yellow.](../fig/inflammation-01-imshow.svg)
```
maxim-belkin, line 30: Couple of comments:
1. Our lesson uses American English -- please change colour to color.
2. I believe "heatmap" is jargon that we don't explain anywhere and we should. We can address it here or in a separate PR. What do you think?

maxim-belkin (changes requested):
Thanks for your contribution, John! ❤️
Please see my comments below.

_episodes/03-matplotlib.md
grossular: Thank you for the comments. I will make the spelling change.

I don't mind adding some text explaining a heatmap in this PR if you think it is appropriate, but I do think it should be outside the image description. Maybe in the text directly following this image?

Also, I have just noticed that the rest of the text uses 'heat map' rather than 'heatmap'. Should I modify the image description to make this consistent?

maxim-belkin: > I don't mind adding some text explaining a heatmap in this PR if you think it is appropriate, but I do think it should be outside the image description. Maybe in the text directly following this image?

We can add an explanation to the Glossary (reference.md) and link to it in the main text.
I think it makes sense to do it in a separate PR but this is totally up to you.

> Also, I have just noticed that the rest of the text uses 'heat map' rather than 'heatmap'. Should I modify the image description to make this consistent?

That'd be great! I don't have a preference as to what spelling to use but we should use the same spelling throughout the lesson.

grossular: > We can add an explanation to the Glossary (reference.md) and link to it in the main text.
I think it makes sense to do it in a separate PR but this is totally up to you.

I agree a separate PR would make more sense.

maxim-belkin:
Looks great! Final comments/questions: how about changing`'data' variable matrix` to  `data variable` (or ``` `data` variable ```) -- I think 'variable matrix' has a different meaning than what we're dealing with here.

grossular:
> Looks great! Final comments/questions: how about changing` 'data' variable matrix` to `data variable` (or `` `data` variable ``) -- I think 'variable matrix' has a different meaning than what we're dealing with here.

Agreed. I'll make that change now

maxim-belkin (approved):
Thank you, John!

maxim-belkin:
Thank you for your contribution to the lesson, John! 🥇

PR #889, opened 2020-11-16
Add a description for heat map.

grossular:
<details>
Added description for heat map to reference.md and linked in 03-matplotlib.md.

As discussed in #888

</details>

### DIFF
--- a/reference.md
+++ b/reference.md
@@ -111,6 +111,10 @@ function
 function call
 :   A use of a function in another piece of software.
 
+heat map
+:   A graphical representation of data in two-dimensions,
+    where data values within the matrix are transformed into a color scale.
+
 immutable
 :   Unchangeable.
     The value of immutable data cannot be altered after it has been created.
--- a/reference.md
+++ b/reference.md
@@ -112,8 +112,8 @@ function call
 :   A use of a function in another piece of software.
 
 heat map
-:   A graphical representation of data in two-dimensions,
-    where data values within the matrix are transformed into a color scale.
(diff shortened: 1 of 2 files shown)


reference.md
```diff
@@ -111,6 +111,10 @@ function
 function call
 :   A use of a function in another piece of software.
 
+heat map
+:   A graphical representation of data in two-dimensions,
+    where data values within the matrix are transformed into a color scale.
```
maxim-belkin, line 116: I'm not sure if everyone knows what a color scale is. I'd say that heat map is a graphical representation of two-dimensional data, where different data values are represented with different colors. Perhaps, @ldko has a better suggestion.

maxim-belkin (changes requested):
Thank you, @grossular! Your changes look great! 👍 Only one comment about "color scale" and a suggestion on how to avoid using it.

reference.md
ldko: I don't think "represented with different colors" is as accurate as "color scale", as it doesn't indicate/imply the variation in colors providing meaning to the visualization. If wanting to avoid "color scale", another suggestion:  "A graphical representation of two-dimensional data in which colors, ranging on a scale of hue or intensity, represent the data values."

maxim-belkin: Thanks, Lauren. I like your suggestion -- it's a bit more involved but is more accurate, which is probably what we want for the glossary.

grossular: That sounds good to me. I'll make the change now.

maxim-belkin (approved):
Perfect! Thank you, John!

PR #907, opened 2021-01-30
index.md: add lesson-overview alt text

ldko:
This adds a title and alt text for the lesson overview image in index.md. Also, it shortens one line in the file that was over 100 characters in length. ping @maxim-belkin does this alt text work for you?


maxim-belkin (approved):
Looks great! Thanks, @ldko. Below is how I would rephrase it but please feel free to merge the PR as is.

> 3-step flowchart shows inflammation data records for several patients moving to the "Analysis" step where a heat map of provided data is generated moving to the "Conclusion" step that asks the question "How does the medication affect patients?"

PR #910, opened 2021-02-05
Add a note on the convention of importing matplotlib.pyplot as plt

jack89roberts:
As discussed in #830 (and perhaps controversially!) this adds a note to episode 3 to mention that `import matplotlib.pyplot as plt` is a common convention. Note that:

- The same is done in the previous episode for `import numpy as np`. I've used the same heading as that for the pyplot import and kept the content short as it has been seen previously.

- I've put it at the end of the episode, just before the exercises, to try to help avoid cluttering/confusing the flow of the rest of the episode.

Very happy to make any necessary changes based on comments, but I do believe this is a worthwhile addition due to how ubiquitous this syntax is (including in the matplotlib documentation itself).

### DIFF
--- a/_episodes/03-matplotlib.md
+++ b/_episodes/03-matplotlib.md
@@ -147,7 +147,6 @@ formats, including SVG, PDF, and JPEG.
 > `import matplotlib.pyplot as plt` then `matplotlib.pyplot.plot(...)` will not work, and
 > you must use `plt.plot(...)` instead. Because of this, when working with other people it
 > is important you agree on how libraries are imported.
->
 {: .callout}
 
 > ## Plot Scaling
--- a/_episodes/03-matplotlib.md
+++ b/_episodes/03-matplotlib.md
@@ -135,7 +135,7 @@ formats, including SVG, PDF, and JPEG.
 > `import matplotlib.pyplot as plt` are frequently used.
 > Importing `pyplot` this way means that after the initial import, rather than writing
 > `matplotlib.pyplot.plot(...)`, you can now write `plt.plot(...)`.
-> Another common convention to use the shortcut `import numpy as np` when importing the
+> Another common convention is to use the shortcut `import numpy as np` when importing the
 > NumPy library. We then can write `np.loadtxt(...)` instead of `numpy.loadtxt(...)`,
 > for example.
 > 
--- a/_episodes/03-matplotlib.md
+++ b/_episodes/03-matplotlib.md
@@ -130,10 +130,24 @@ formats, including SVG, PDF, and JPEG.
 
 > ## Importing libraries with shortcuts
 >
-> When importing the `pyplot` module of `matplotlib` it is common to use the shortcut
-> `import matplotlib.pyplot as plt`. If we do this we then must write `plt.plot(...)`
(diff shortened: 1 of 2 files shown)


ldko (approved):
Thanks very much @jack89roberts. I think this is a helpful addition to include. It looks good to me. @maxim-belkin , does this work for you?

maxim-belkin:
I'd like to push back on this a little... The reason being, it was previously decided not to use `import ***** as ***` in the lesson (I'll refer to #61 for a number of good points "for" and "against" this decision) and adding one more callout box on this topic seems a bit strange. However, I think that an example of using `plt` instead of `matplotlib.pyplot` is more vivid than the `np`/`numpy` one... , so, perhaps, we should move the callout box that introduces `import ***** as ***` syntax to the episode where we introduce/use matplotlib and then give both examples (numpy and matplotlib). What do you think?

jack89roberts:
That would also have the benefit of it coming after the students have had a bit more experience with importing libraries the normal way. I'm happy to try that (may not get to it until later in the week though). I just think it's important for students to see both shortcuts, at least briefly, since essentially all documentation or help they look for afterwards will use that syntax.

ldko:
I think putting both shortcuts into one callout box in the matplotlib lesson makes sense.

jack89roberts:
I've just pushed a commit removing the callout box from the NumPy episode and adding the content to the Matplotlib episode. Let me know what you think (also about where the box should be placed in the episode - it's at the end currently).

_episodes/03-matplotlib.md
```diff
@@ -130,10 +130,24 @@ formats, including SVG, PDF, and JPEG.
...
-> When importing the `pyplot` module of `matplotlib` it is common to use the shortcut
-> `import matplotlib.pyplot as plt`. If we do this we then must write `plt.plot(...)`
-> instead of `matplotlib.pyplot.plot(...)`, for example. This is popular as it saves
-> a lot of typing!
+> In this lesson we use the `import matplotlib.pyplot` [syntax]({{ page.root }}/reference.html#syntax)
+> to import the `pyplot` module of `matplotlib`. However, shortcuts such as
+> `import matplotlib.pyplot as plt` are frequently used.
+> Importing `pyplot` this way means that after the initial import, rather than writing
+> `matplotlib.pyplot.plot(...)`, you can now write `plt.plot(...)`.
+> Another common convention to use the shortcut `import numpy as np` when importing the
```
ldko, line 138: I think this is missing a word...perhaps it should be "Another common convention **is**"...

ldko (commented):
I am happy with the new placement of the shortcuts callout and the mention of both `np` and `plt`, so learners will recognize the shortcuts if they see them in future work. I think there is one typo to fix.

_episodes/03-matplotlib.md
jack89roberts: Thanks, just pushed a commit to add the "is".

ldko:
@maxim-belkin does this PR now satisfy the changes you wanted?

maxim-belkin:
The changes look great! I think two examples make the idea we're trying to convey in the callout box clearer without putting too much emphasis on it. Thank you, @jack89roberts!

maxim-belkin (approved):

PR #911, opened 2021-02-28
Fix long lines in README.md

gmingas:
Fixes long lines in README.md.

Partially addresses #498.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,28 +28,18 @@ This lesson is also available in [R][R] and [MATLAB][MATLAB].
 
 | # |  Episode | Time | Question(s) |
 |--:|:---------|:----:|:------------|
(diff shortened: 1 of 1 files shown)


ldko (changes requested):
Hi @gmingas ,
Thanks for putting together pull requests to fix the long lines in this lesson. Looking at the rendered version of the [README.md with your changes](https://github.com/gmingas/python-novice-inflammation/tree/readme-fix-long-lines), it looks like breaking up the lines that make up the table in the Github markdown causes the separated out text to appear on separate rows. I'm not sure if there is a way to get those lines under 100 characters without this happening (if anyone knows how, please let us know), so we may need to leave those lines longer than the general preference to have it render correctly.

maxim-belkin (approved):
👍🏻 Thanks again, @gmingas!

PR #914, opened 2021-02-28
Fix long line in episode 11

gmingas:
Fixes long line in episode 11,  making it 100 characters max.

Partially addresses #498


ldko (approved):
This looks like it takes care of episode 11. Thank you very much, @gmingas

Issue #874, opened 2020-10-03
Episode 10 Defensive — Using the assert statement to validate data is bad practice


baileythegreen:
# Relating to [Episode 10 — Defensive Programming](https://github.com/swcarpentry/python-novice-inflammation/blob/gh-pages/_episodes/10-defensive.md)

The first half of this lesson advocates using Python's `assert` statement to do data validation. While this may be *common* (I don't know), it is not correct, and can potentially be very dangerous (explanation to follow).

By default if a Python program is run from the command line, assertions will be evaluated; however, there are two flags, `-O` and `-OO`, which run Python programs in optimisation mode and disable assertions:

```
-O : remove assert and __debug__-dependent statements; add .opt-1 before .pyc extension; also PYTHONOPTIMIZE=x
-OO    : do -O changes and also discard docstrings; add .opt-2 before .pyc extension
```

If `assert` statements are used for data validation, this means that none of that validation will take place when the program is run in one of these modes, which, depending on the program, will result in surprising errors at best, and at worst, could be a security risk.

`assert` statements should be used as debugging aids, not to handle run-time errors.

A more suitable way of validating the data in the first example would be:

```
for num in numbers:
          if num > 0:
              total += 1
          else:
              raise ValueError('Data should only contain positive values')
```

This data validation step will always run.

NB: I would be happy to work on, or contribute to, changing this lesson so that it reflects best practices, but as it is a larger change, though raising it in an issue first was the best approach. — Bailey

ldko:
Hi @baileythegreen,
We have similarly had people bring up the trouble with assertions in issues #622, #773 and propose to bring in exception handling instead. The maintainers agree that it makes sense to modify this part of the episode, and we would welcome a PR from you. Thank you, Bailey!

baileythegreen:
Hi @ldko,
I've had a look through those issue threads and I don't see any current pull requests related to them, so I'll start on a new one, trying to keep the discussion that has gone before in mind.

As it is, I think the main consideration is whether to demonstrate the `try...except(...else...finally)` statements, or stick with simply raising an appropriate error type. I'll look into both to see how they fit with the rest of the lesson(s). Cheers!

DanShort12:
I was also taking a look at this session and wasn't so keen to see the use of `assert` without any clarification that it should only be used in unreleased settings (testing mainly).

[...]

Issue #773, opened 2019-12-21
Defensive Programming: replacing assertions with exceptions 


alexdesiqueira:
Hi everyone,
I was lecturing using the Defensive Programming lesson ([09-defensive.md](https://github.com/swcarpentry/python-novice-inflammation/blob/gh-pages/_episodes/09-defensive.md)) today, and I checked we are using several assertions to verify code. However, this is not a very good programming practice: assertions should provide a security check only when writing and debugging the program.
When the code is released, no assertions should fail, as Bil Lewis explains with more details [in this post](https://www.drdobbs.com/jvm/assertations-vs-exceptions/228701655).
More than that, using assertions in Python could induce to some errors: `assert` statements are removed when the compilation is optimized, as noted in [this Stack Overflow answer](https://stackoverflow.com/questions/944592/best-practice-for-python-assert/1838411#1838411). This way, _all_ tests and checking would be lost if the code is optimized (executed using `python -O`):
```
$ python -h                                                                                                     
usage: python [option] ... [-c cmd | -m mod | file | -] [arg] ...
Options and arguments (and corresponding environment variables):
(...)
-O     : remove assert and __debug__-dependent statements; add .opt-1 before
         .pyc extension; also PYTHONOPTIMIZE=x
```
I propose to rewrite this episode using exceptions instead of assertions. For example, the first piece of code in that lesson would be:

```
numbers = [1.5, 2.3, 0.7, -0.001, 4.4]
total = 0.0
for num in numbers:
    if num <= 0:
        raise ValueError('Data should only contain positive values')
    total += num
print('total is:', total)
```
If we choose different exceptions, remembering they were shown previously in [08-errors.md](https://github.com/swcarpentry/python-novice-inflammation/blob/gh-pages/_episodes/08-errors.md), we show how to present more meaningful errors, rather than always presenting an `AssertionError`.
Thank you very much for considering this. Thanks also to @stefanv and @jarrodmillman for helping me understanding some points in this issue.

ldko:
[...]

PR #917, opened 2021-03-16
Add mention of swap idiom to 'The Old Switcheroo' exercise

### LINKED ISSUE #883: Lesson Contribution - Swapping variables
I'm a member of The Carpentries staff and I'm submitting this issue on behalf of another member of the community. In most cases, I won't be able to follow up or provide more details other than what I'm providing below.

----------------------------------------------
1. python-novice-inflammation -
> Episode 8 - *https://swcarpentry.github.io/python-novice-inflammation/08-func/index.html
> <https://swcarpentry.github.io/python-novice-inflammation/08-func/index.html>;
> - *Creating functions introduces swapping variables under exercise "The
> Old Switcheroo". I would like to suggest mentioning that there is an
> elegant way of swapping variables in python.
> b, a = a, b

svaberg:
This pull request proposes to close #883 by adding a brief note about the switch idiom `a, b = b, a` in  **The Old Switcheroo** exercise and linking to the description of the switch idiom already present in the **Extra Exercises** page.

In the **Extra Exercises** page it also replaces `a, b = [b, a]` with the more idiomatic and simpler `a, b = b, a`.

### DIFF
--- a/_episodes/08-func.md
+++ b/_episodes/08-func.md
@@ -918,6 +918,10 @@ readable code!
 > > `a` and `b` in this case) and trades their values. The function does not
 > > return any values and does not alter `a` or `b` outside of its local copy.
 > > Therefore the original values of `a` and `b` remain unchanged.
+> >
+> > Note that the recommended way of swapping the content of `a` and `b` in Python is
+> > `a, b = b, a`. You can see why this works in the
+> > [Additional Exercises]({{ page.root }}/extra_exercises/index.html).
 > {: .solution}
 {: .challenge}
 
--- a/_episodes/08-func.md
+++ b/_episodes/08-func.md
@@ -832,7 +832,11 @@ readable code!
 > > ~~~
 > > {: .output}
 > > `k` is 0 because the `k` inside the function `f2k` doesn't know
-> > about the `k` defined outside the function.
+> > about the `k` defined outside the function. When the f2k function is called, 
+> > it creates a [local variable]({{ page.root }}/reference.html#local-variable)
+> > `k`. The function does not return any values 
+> > and does not alter `k` outside of its local copy. 
+> > Therefore the original value of `k` remains unchanged.
 > {: .solution}
 {: .challenge}
 
--- a/_episodes/08-func.md
+++ b/_episodes/08-func.md
@@ -832,7 +832,7 @@ readable code!
 > > ~~~
 > > {: .output}
 > > `k` is 0 because the `k` inside the function `f2k` doesn't know
-> > about the `k` defined outside the function. When the f2k function is called, 
+> > about the `k` defined outside the function. When the `f2k` function is called, 
 > > it creates a [local variable]({{ page.root }}/reference.html#local-variable)
 > > `k`. The function does not return any values 
 > > and does not alter `k` outside of its local copy. 
--- a/_episodes/08-func.md
+++ b/_episodes/08-func.md
@@ -889,46 +889,6 @@ readable code!
 > {: .solution}
 {: .challenge}
 
-> ## The Old Switcheroo
->
-> Consider this code:
->
-> ~~~
-> a = 3
-> b = 7
->
-> def swap(a, b):
->     temp = a
->     a = b
->     b = temp
->
-> swap(a, b)
->
-> print(a, b)
-> ~~~
-> {: .language-python}
->
-> Which of the following would be printed if you were to run this code?
-> Why did you pick this answer?
->
-> 1. `7 3`
-> 2. `3 7`
-> 3. `3 3`
-> 4. `7 7`
->
-> > ## Solution
-> > `3 7` is the correct answer. Initially, `a` has a value of 3 and `b` has a value of 7.
-> > When the `swap` function is called, it creates local variables (also called
-> > `a` and `b` in this case) and trades their values. The function does not
-> > return any values and does not alter `a` or `b` outside of its local copy.
-> > Therefore the original values of `a` and `b` remain unchanged.
-> >
-> > Note that the recommended way of swapping the content of `a` and `b` in Python is
-> > `a, b = b, a`. You can see why this works in the
-> > [Additional Exercises]({{ page.root }}/extra_exercises/index.html).
-> {: .solution}
-{: .challenge}
-
 > ## Readable Code
 >
 > Revise a function you wrote for one of the previous exercises to try to make
--- a/_extras/extra_exercises.md
+++ b/_extras/extra_exercises.md
@@ -22,7 +22,7 @@ or not (yet) added to the main lesson.
 > Compare it to:
 >
 > ~~~
-> left, right = [right, left]
+> left, right = right, left
 > ~~~
 > {: .language-python}
 >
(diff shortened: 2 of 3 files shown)


_extras/extra_exercises.md
```diff
@@ -22,7 +22,7 @@ or not (yet) added to the main lesson.
 > Compare it to:
 >
 > ~~~
-> left, right = [right, left]
+> left, right = right, left
```
ldko, line 25: The solution mentions a `list`, so if we change this, we would need to change that. However, we don't teach tuples in the lesson, so we would need to take that into consideration if we make a change here that uses a `tuple`.

_episodes/08-func.md
```diff
@@ -918,6 +918,10 @@ readable code!
 > > `a` and `b` in this case) and trades their values. The function does not
 > > return any values and does not alter `a` or `b` outside of its local copy.
 > > Therefore the original values of `a` and `b` remain unchanged.
+> >
+> > Note that the recommended way of swapping the content of `a` and `b` in Python is
+> > `a, b = b, a`. You can see why this works in the
+> > [Additional Exercises]({{ page.root }}/extra_exercises/index.html).
```
ldko, line 924: Based on the resolution of the issue #460, I am not sure we want to include this idiom. Also, with the challenge ultimately being about variable scope rather than how to swap variable values, it may be distracting.

ldko (commented):
Hi @svaberg ,
Thanks very much for submitting this contribution to the lesson. In looking back at the history of the topic of swapping variables in this lesson, I see #460 ultimately decided not to bring in the swapping variable idiom of `a, b = b, a` (at least not in the `list` episode). I think the reasoning given there still makes sense in the context of the functions lesson, but let's see what @maxim-belkin has to say as well.

_episodes/08-func.md
svaberg: I was thinking about this overnight. From a distractions perspective the whole swapping functionality of the function may be distracting from the challenge itself. Maybe the **The Old Switcheroo** exercise should be replaced by a function that does not switch variables at all.

ldko: I agree with you about the swapping being distracting; this seems something of a trick question. Looking at the other exercises based on your comment, I notice the challenge titled "Variables Inside and Outside Functions" seems to highlight the same concept of variable scope. Perhaps the Old Switcheroo could go away altogether in that case. Additionally, scanning over the episode material, I am not seeing us actually introduce the concept of variable scope or local variable...if that is the case, perhaps it is something we need to add... @maxim-belkin ?

_extras/extra_exercises.md
svaberg: Tuples are mentioned in episode **10 Defensive Programming** before the `normalize_rectangle` function. 

In the extras they are discussed some more in the **Image Grids** of the **Extras/Discussion**.

ldko: Ah thanks for looking into this @svaberg--I tend to skip teaching the `normalize_rectangle` function part because I find the content a bit confusing, and teaching the `assert` stuff is questionable (#622, #773, #874 ). Looks like we could remove the brackets and mention `tuple` instead of `list` and link to the `tuple` reference again.

svaberg: Okay I'll get into that!

_extras/extra_exercises.md
```diff
@@ -44,7 +44,8 @@ or not (yet) added to the main lesson.
 > >
 > > In the first case we used a temporary variable `temp` to keep the value of `left` before we
 > > overwrite it with the value of `right`. In the second case, `right` and `left` are packed into a
-> > list and then unpacked into `left` and `right`.
+> > [tuple]({{ page.root }}/reference.html#tuple)
```
svaberg, line 47: I changed `list` to `tuple` and added a link to the `reference.html` `tuple` entry.

_extras/extra_exercises.md
svaberg: I made this change, see below.

_episodes/08-func.md
svaberg: I agree that the **Variables Inside and Outside Functions** challenge teaches essentially the same content, and in a less cluttered way. As such, maybe **The Old Switcheroo** should indeed be removed. 

Maybe some of the explanatory text from the  **The Old Switcheroo** solution about local variables could be shifted over to the  **Variables Inside and Outside Functions**?

ldko: I think it makes sense to move some of the description, such as adding to the end of what is there:  

When the `f2k` function is called, it creates a local variable `k`. The function does not return any values and does not alter `k` outside of its local copy. Therefore the original value of `k` remains unchanged.

I think it could also be useful to create a glossary entry for local variable and link to it there. The text could also differ somewhat if you think it could be explained better in a different way.

svaberg: Agreed. Does this mean we are moving towards removing **The Old Switcheroo**? I'll try and implement the changes you propose.

ldko: Unless @maxim-belkin or  others in the community speaks up about why we should keep the exercise, it would be great for you to carry out the proposed changes. Thank you, @svaberg!

svaberg: I've carried out all the proposed changes but I have not yet removed **the Old Switcheroo**. I'll do that now in a separate commit.

maxim-belkin:
I usually cover this "Old switcheroo" exercise when I teach our lesson, so I'd like to propose to modify it instead of removing it. We can do something as primitive as multiplying a by 10 and adding 10 to b instead of switching their values. I think it's an important demonstration for those who're new to functions.

ldko:
Hi @maxim-belkin, can you clarify what you mean as being "an important demonstration for those who're new to functions"? Are you referring to something not demonstrated in the "Variables Inside and Outside Functions" that @svaberg expanded upon in this PR?

maxim-belkin:
I'm referring to the same idea/concept as in the `f2k` exercise but I think it's an important one so would prefer our lesson to have more than a single exercise devoted to it. If you think we can drop it -- so be it, but it's worth noting that other SWC/DC Python lessons don't talk about it much.

_episodes/08-func.md
```diff
@@ -832,7 +832,11 @@ readable code!
 > > ~~~
 > > {: .output}
 > > `k` is 0 because the `k` inside the function `f2k` doesn't know
-> > about the `k` defined outside the function.
+> > about the `k` defined outside the function. When the f2k function is called,
```
ldko, line 835:
```suggestion
> > about the `k` defined outside the function. When the `f2k` function is called, 
```

reference.md
```diff
@@ -148,8 +148,12 @@ library
 :   A family of code units (functions, classes, variables) that implement a set of
     related tasks.
 
+local variable
+:   A variable that is defined in a block of code, and that is not accessible outside the
+    block of code in which it is ceated.
+
 loop variable
-:   The variable that keeps track of the progress of the loop.
+:   The variable that keeps track of the progress of the loop.
```
ldko, line 156:
```suggestion
:   The variable that keeps track of the progress of the loop.
```

ldko (commented):
Thanks for addressing all my comments, @svaberg . I have marked a couple very minor things.

maxim-belkin:
> What do you think about adding something like this after the `fahr_to_kelvin` example:

I think this is a great suggestion. In general, it might be worth converting existing material to a (small) section devoted to variable scopes.

ldko:
Ok, @maxim-belkin I can open a separate PR to add a section about variable scope.

maxim-belkin (approved):
I think it looks good! Thank you for your contribution, Dag!

PR #922, opened 2021-03-22
Episodes 2, 4, 6, 10: Improving image descriptions

vinisalazar:
I've added suggested image descriptions for the remaining figures listed in Issue #781.

This includes changes to Episodes 2, 4, 6, and 10.

Thank you,

V

### DIFF
--- a/_episodes/02-numpy.md
+++ b/_episodes/02-numpy.md
@@ -233,7 +233,11 @@ It takes a bit of getting used to,
 but one way to remember the rule is that
 the index is how many steps we have to take from the start to get the item we want.
 
-![Python indices are zero-based. In array-like objects, this can be observed for both the horizontal and vertical axes.](../fig/python-zero-index.svg)
+!["data" is a 3 by 3 numpy array containing row 0: ['A', 'B', 'C'], row 1: ['D', 'E', 'F'], and
+row 2: ['G', 'H', 'I']. Starting in the upper left hand corner, data[0, 0] = 'A', data[0, 1] = 'B',
+data[0, 2] = 'C', data[1, 0] = 'D', data[1, 1] = 'E', data[1, 2] = 'F', data[2, 0] = 'G',
+data[2, 1] = 'H', and data[2, 2] = 'I',
+in the bottom right hand corner.](../fig/python-zero-index.svg)
 
 > ## In the Corner
 >
--- a/_episodes/02-numpy.md
+++ b/_episodes/02-numpy.md
@@ -233,7 +233,7 @@ It takes a bit of getting used to,
 but one way to remember the rule is that
 the index is how many steps we have to take from the start to get the item we want.
 
-![Zero Index](../fig/python-zero-index.svg)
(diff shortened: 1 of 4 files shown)


_episodes/02-numpy.md
```diff
@@ -233,7 +233,7 @@ It takes a bit of getting used to,
 but one way to remember the rule is that
 the index is how many steps we have to take from the start to get the item we want.
 
-![Zero Index](../fig/python-zero-index.svg)
+![Python indices are zero-based. In array-like objects, this can be observed for both the horizontal and vertical axes.](../fig/python-zero-index.svg)
```
ldko, line 236: For me this doesn't really describe what the image itself shows. We have worked on text for this in a couple other PRs--it is a bit difficult because the image includes so much text. I was awaiting a response on [this comment](https://github.com/swcarpentry/python-novice-inflammation/pull/844/files#r450535168) that suggests using:

> "data" is a 3 by 3 numpy array containing row 0: \['A', 'B', 'C'\], row 1: \['D', 'E', 'F'\], and row 2: \['G', 'H', 'I'\]. Starting in the upper left hand corner, data\[0, 0\] = 'A', data\[0, 1\] = 'B', data\[0, 2\] = 'C', data\[1, 0\] = 'D', data\[1, 1\] = 'E', data\[1, 2\] = 'F',  data\[2, 0\] = 'G', data\[2, 1\] = 'H', and data\[2, 2\] = 'I', in the bottom right hand corner.

What do you think about this text @vinisalazar ?

ldko (commented):
I think the alt text you offer for episodes 4, 6, and 10 would be fine. I suggest more description for the python-zero-index.svg figure. Thanks for putting this together, @vinisalazar !

_episodes/02-numpy.md
vinisalazar: Thanks for pointing that out. I agree that it doesn't literally describe what the image shows. Sorry, I missed that point.

I tried thinking about alternatives for this one but I think this suggestion you posted is the most literal description of the image. Adding it would be better than the current alt-text, but I guess that only a user that depends on these alt texts could provide more accurate feedback.

I'm not sure on the best way to proceed here, *i.e.* should I undo my change or should I leave it as is and it will be replaced by the text on #844?

Best,
V

ldko: Unfortunately I didn't follow-up further to get either of the 2 previous PRs about alt text for this image resolved (sorry @blankdots @magsol). If you are ok with the suggested text @vinisalazar  , I think the easiest thing would be if you add this text in your PR. If someone feels this needs to be changed down the line, we will welcome a PR to improve it when it comes. If you prefer not to modify the text as suggested, it is fine if you want to remove the change for 02-numpy.md. Thank you!

vinisalazar: Yep, no problem. Done with bc1dec7.

ldko (approved):
Thanks, @vinisalazar ! I think these look good.

annefou:
@vinisalazar These are great changes! Very helpful. Thanks!

annefou (approved):

PR #923, opened 2021-03-25
Episode 7: Add `None` to "What is Truth?" challenge

vinisalazar:
Related to issue #358.

### DIFF
--- a/_episodes/07-cond.md
+++ b/_episodes/07-cond.md
@@ -287,6 +287,8 @@ freeing us from having to manually examine every plot for features we've seen be
 >     print('empty list is true')
 > if [1, 2, 3]:
 >     print('non-empty list is true')
+> if None:
+>     print('none is true')
 > if 0:
 >     print('zero is true')
 > if 1:


ldko:
Note: up through episode 7, `None` has not been introduced in the lesson, so including it in the challenge is problematic.

vinisalazar:
Hi, thank you for the comments.

@maxim-belkin I am aware that Python is case-sensitive. All other statements in the challenge are in lower case, I simply wanted to maintain consistency.

@ldko that is indeed a problem. I see now that `None` is only introduced in Episode 8.

My apologies, I will close this PR and refrain from making such changes without proper justification.

Best,
V

ldko:
@vinisalazar thanks for taking the time to submit this suggestion. `None` evaluating to `False` is one of those things that will come in handy for people who continue using Python to know at some point. Unfortunately, there are so many useful things like this, we have to leave a lot of them out keeping in mind that these materials are designed to be used in a workshop format where the lesson is generally covered in 6 hours. Some of the things we leave out are mildly painful, but in the context of the materials, we do it. :) Thanks for all your contributions this past week!

vinisalazar:
@ldko I understand and agree with you. As for the contributions, I'm happy to help! :)

Best,
V

PR #924, opened 2021-03-31
UPD: correct python snippets

vkhodygo:
I noticed that almost all examples of the code are slightly outdated and could be improved.
I've replaced certain strings with old formatting with f-strings as it is the recommended way at the moment. 
In addition, I added shebangs to all files that can be run as individual scripts;
I also made code style more consistent, i.e., all function now have a return statement even if they return None. 
Finally, some code has been modified to catch minor exceptions such as division by zero or to read files with various separators, not CSVs only. 

This is a draft version as a) some changes are subject to discussion  and b) all corresponding code in the lessons has to be updated respectively after all changes have been made.

### DIFF
--- a/code/argv_list.py
+++ b/code/argv_list.py
@@ -1,2 +1,4 @@
+#!/usr/bin/env python
 import sys
-print('sys.argv is', sys.argv)
+
+print(f'''sys.argv is {sys.argv}''')
--- a/code/arith.py
+++ b/code/arith.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python
 import sys
 
 
@@ -6,19 +7,19 @@ def main():
 
     operator = sys.argv[1]
     assert operator in ['add', 'subtract', 'multiply', 'divide'], (
-        'Operator is not one of add, subtract, multiply, or divide: '
-        'bailing out')
+        '''Operator is not one of add, subtract, multiply, or divide:
+        bailing out''')
     try:
         operand1, operand2 = float(sys.argv[2]), float(sys.argv[3])
     except ValueError:
         print('Cannot convert input to a number: bailing out')
-        return
+        return None
 
     do_arithmetic(operand1, operator, operand2)
+    return None
 
 
 def do_arithmetic(operand1, operator, operand2):
-
     if operator == 'add':
         value = operand1 + operand2
     elif operator == 'subtract':
@@ -26,8 +27,15 @@ def do_arithmetic(operand1, operator, operand2):
     elif operator == 'multiply':
         value = operand1 * operand2
     elif operator == 'divide':
-        value = operand1 / operand2
+        try:
+            value = operand1 / operand2
+        except ZeroDivisionError:
+            print('Cannot divide by zero: bailing out')
+            return None
+    else:
+        value = None
     print(value)
+    return None
 
 
 if __name__ == '__main__':
--- a/code/check.py
+++ b/code/check.py
@@ -1,32 +1,36 @@
+#!/usr/bin/env python
 import sys
 import numpy
 
 
 def main():
-    script = sys.argv[0]
     filenames = sys.argv[1:]
-    if len(filenames) <= 1:  # nothing to check
-        print('Only 1 file specified on input')
+    total_files = len(filenames)
+    if total_files <= 1:  # nothing to check
+        print(f'''{'Only one file' if total_files else 'None files'} specified
+        on input''')
     else:
         nrow0, ncol0 = row_col_count(filenames[0])
-        print('First file %s: %d rows and %d columns' % (
-            filenames[0], nrow0, ncol0))
+        print(f'''First file {filenames[0]}: {nrow0} rows and {ncol0}
+        columns''')
         for filename in filenames[1:]:
             nrow, ncol = row_col_count(filename)
             if nrow != nrow0 or ncol != ncol0:
-                print('File %s does not check: %d rows and %d columns'
-                      % (filename, nrow, ncol))
+                print(f'''File {filename} does not check: {nrow} rows
+                and {ncol} columns''')
             else:
-                print('File %s checks' % filename)
-        return
+                print(f'File {filename} checks')
+
+    return None
 
 
 def row_col_count(filename):
     try:
-        nrow, ncol = numpy.loadtxt(filename, delimiter=',').shape
-    except ValueError:
-        # This occurs if the file doesn't have same number of rows and columns,
-        # or if it has non-numeric content
+        nrow, ncol = numpy.loadtxt(filename).shape
+    except ValueError as val:
+        # This occurs if the file does contain rows
+        # with inconsistent number of elements or if it has non-numeric content
+        print(val)
         nrow, ncol = (0, 0)
     return nrow, ncol
 
--- a/code/count_stdin.py
+++ b/code/count_stdin.py
@@ -1,7 +1,8 @@
+#!/usr/bin/env python
 import sys
 
 count = 0
 for line in sys.stdin:
     count += 1
 
-print(count, 'lines in standard input')
+print(f'{count} lines in standard input')
--- a/code/gen_inflammation.py
+++ b/code/gen_inflammation.py
@@ -6,6 +6,8 @@ Generate pseudo-random patient inflammation data for use in Python lessons.
 
 import random
 
+random.seed(1)
+
 n_patients = 60
 n_days = 40
 n_range = 20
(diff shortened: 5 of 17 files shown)


code/argv_list.py
```diff
@@ -1,2 +1,4 @@
+#!/usr/bin/env python
```
ldko, line 1: Including the shebang in files would require us taking time to explain what this is to learners. Since it is not already in the lesson material, we would need to justify the time it would require to add to a workshop's limited time frame. I see it as being difficult to justify including the shebang in the context of the workshop's limited time and scope, since it is not required to include in a script.

code/argv_list.py
```diff
@@ -1,2 +1,4 @@
...
 import sys
-print('sys.argv is', sys.argv)
+
+print(f'''sys.argv is {sys.argv}''')
```
ldko, line 4: I wasn't familiar with f-strings before this PR, so I appreciate you bringing it up as something to consider using. I found the [PEP](https://www.python.org/dev/peps/pep-0498/) for it, but can you point me to where you saw this as being the recommended way of string formatting right now? I didn't see it mentioned as such in the [official docs](https://docs.python.org/3.9/library/string.html#format-string-syntax).

code/arith.py
```diff
@@ -6,28 +7,35 @@ def main():
 
     operator = sys.argv[1]
     assert operator in ['add', 'subtract', 'multiply', 'divide'], (
-        'Operator is not one of add, subtract, multiply, or divide: '
-        'bailing out')
+        '''Operator is not one of add, subtract, multiply, or divide:
+        bailing out''')
```
ldko, line 11: I think the reason this was done with the adjacent single-quoted strings inside the parentheses is because with using the triple-quoted, you get that whitespace before "bailing out" as literal whitespace.

code/arith.py
```diff
@@ -6,28 +7,35 @@ def main():
...
     try:
         operand1, operand2 = float(sys.argv[2]), float(sys.argv[3])
     except ValueError:
         print('Cannot convert input to a number: bailing out')
-        return
+        return None
```
ldko, line 16: I think using `return None` instead of `return` could be a helpful change to point out what is happening to learners.

code/arith.py
```diff
@@ -6,28 +7,35 @@ def main():
...
 
     do_arithmetic(operand1, operator, operand2)
+    return None
```
ldko, line 19: Since we explain in the Functions episode that "functions are not required to include a return statement", and I don't think it is common practice to always put one explicitly at the end of a function, my preference would be not to add these to all of the functions.

code/arith.py
```diff
@@ -6,28 +7,35 @@ def main():
...
 
 
 def do_arithmetic(operand1, operator, operand2):
-
```
ldko, line 21: You're right, we don't need that line :).

code/arith.py
```diff
@@ -6,28 +7,35 @@ def main():
...
         value = operand1 - operand2
     elif operator == 'multiply':
         value = operand1 * operand2
     elif operator == 'divide':
-        value = operand1 / operand2
+        try:
+            value = operand1 / operand2
+        except ZeroDivisionError:
+            print('Cannot divide by zero: bailing out')
+            return None
```
ldko, line 34: Looking at this challenge in the lesson, it says "Write a command-line program that does addition and subtraction:" I am not sure why multiply and divide are included in the solution :/.

code/check.py
```diff
@@ -1,32 +1,36 @@
...
 
 def main():
-    script = sys.argv[0]
     filenames = sys.argv[1:]
-    if len(filenames) <= 1:  # nothing to check
-        print('Only 1 file specified on input')
+    total_files = len(filenames)
+    if total_files <= 1:  # nothing to check
+        print(f'''{'Only one file' if total_files else 'None files'} specified
+        on input''')
```
ldko, line 11: I think this syntax could be confusing to the novice learners we are targeting with this lesson. I don't think we cover it exactly.

code/check.py
```diff
@@ -1,32 +1,36 @@
...
+
+    return None
 
 
 def row_col_count(filename):
     try:
-        nrow, ncol = numpy.loadtxt(filename, delimiter=',').shape
-    except ValueError:
-        # This occurs if the file doesn't have same number of rows and columns,
-        # or if it has non-numeric content
```
ldko, line 29: I am wondering if you have tried to run this code with your change (I haven't). The default value for the [`delimiter` parameter](https://numpy.org/doc/stable/reference/generated/numpy.loadtxt.html) is whitespace, so when trying to load our inflammation datasets (which is specified in the challenge this code aligns with) without setting `delimiter=','`, our data isn't parsed correctly. I think we would get a `ValueError` because of the commas, and the outcome of the code would not be as expected. For this reason, I don't think we can remove the `delimiter` parameter from our `loadtxt` calls.

ldko (commented):
Hi @vkhodygo , it looks like you spent some time going through the included scripts. :)  I didn't go through every change made here, but I started with offering some things we need to consider on some of the types of changes made in this PR. Thanks for the effort you put into this!

code/argv_list.py
vkhodygo: Yes, I know, the only reason I decided to include this part is because there was a file like this already.
I think it's either all of them have shebangs or none to make the whole codebase consistent.
If you think that's too much I can remove all of them and update the draft.

code/argv_list.py
vkhodygo: This is actually mentioned in the PEP you provided, the section "Rationale" starts with the following:
 "This PEP is driven by the desire to have a simpler way to format strings in Python. The existing ways of formatting are either error prone, inflexible, or cumbersome."

code/arith.py
vkhodygo: Could you please elaborate on this one?

code/arith.py
vkhodygo: Probably, I can't recall if there is any mention of the difference between `return`, `return None`, and an empty string in the lessons. Again, it's a matter of consistence rather than my personal preference. Should I adjust all of the functions?

code/arith.py
vkhodygo: I've already provided my thoughts regarding this in the comment above. I'm not sure how to proceed further.

code/arith.py
vkhodygo: That is probably a copy-paste issue, your suggestion is to remove multiplication and division?

code/check.py
vkhodygo: Fair point, I should remove this, it's a shame since loop comprehension is so nice.

code/check.py
vkhodygo: My bad, I need to check it again since I use self-made files for testing.

vkhodygo:
Sorry, @ldko, it took me some time to get back to you since I was occupied with some real life commitments. Could you please provide your feedback?

code/argv_list.py
ldko: @maxim-belkin do you have a preference on including or removing all shebangs when making these consistent?

code/argv_list.py
ldko: Looking at the lesson, we don't teach `format()` or formatting strings with `%` , though we do have a few challenges that use the latter between episodes 11 and 12. Since we don't teach these older ways or the f-string formatted string literals, I would suggest consistently using `print()` with multiple arguments as we are now. While knowing a particular way/syntax of formatting strings in Python can be of use to those continuing to code with Python past the workshop, it is not something we necessarily need to go into detail with in trying to cover general programming concepts in the workshop content. If others think we should teach string formatting more in depth we can look at how we want to go about it. @annefou @maxim-belkin any thoughts on string formatting?

maxim-belkin:
Thanks for pinging me, @ldko. Thanks for the PR, @vkhodygo. 

While the changes proposed here are good from Python perspective, unfortunately, I don't think I can approve any of them for the following reasons.

1. Shebangs break things on some platforms -- we had to remove them in the carpentries/styles repo -- see carpentries/styles@a2d1a1f28184982d145297fb0183b9571e75352a
2. Scripts modified in this PR are tied to the 12th episode of the lesson that teaches participants how to write scripts and changes to these scripts must be accompanied by the changes to the said episode. And changes to the episodes have to be discussed first.
3. Currently, we don't teach exception, so we can't use try/except block(s). This, however, might change in the future, at which point we'll likely change the aforementioned episode.
4. We don't teach list comprehensions, so while we don't use `gen_inflammation.py` in the lesson, I think it's imperative that we use the techniques we teach in the lesson.

So, unless both @ldko and @annefou feel that some of the changes here will benefit the lesson, unfortunately I have to turn this PR down. My apologies for not approving the changes. If you're interested, I'm sure you can find an issue in our lesson you can work on. If this PR is part of your instructor checkout -- it counts as a contribution regardless of whether it's merged into the lesson or not.

Thanks for the PR anyway.

ldko:
Based on my earlier comments and Maxim's as well, I am going to close this pull request. We appreciate that you took time to offer suggestions on improving this lesson. @vkhodygo taking into account the changes you would like to see in this lesson, and our comments on this PR, where you see something still needs action, please open an issue for discussion, such as about string formatting or open a PR addressing a single concern such as to remove any shebang's from scripts for consistency in the lesson. Thanks again.

PR #936, opened 2021-06-10
Sync up with carpentries/styles: June 2021

maxim-belkin:
- `bin/util.py`: Factor out reporter class. Define `__all__`
- `lesson_check.py`: add a comment about importing * from a package
- `lesson.scss`: define 'inline' class for images
- Fix Reporter class imports
- update R dependency search; Allow Bioconductor packages (#600)
- Automatically add deep anchor links using AnchorJS


ldko:
Thanks @maxim-belkin for syncing this again!

Issue #937, opened 2021-06-21
Analyzing data from multiple files: Wording regarding glob


codingS3b:
[This episode](https://swcarpentry.github.io/python-novice-inflammation/06-files/index.html) introduces the `glob` package using the words "a library with a rather unpleasant name". 
I see two problems with that: 
1) From my point of view this is not a good choice of words since students might be led towards feeling negatively about this very useful set of functions.
2) I think we should use "package" rather than "library" in the python world since otherwise students new to python might get confused when switching between those concepts.

To resolve both issues, the sentence introducing `glob` could be rephrased with a focus more strongly based on what it can be used for, e.g. "a package which provides functionality for filtering files based on their naming scheme".

ldko:
Hi @codingS3b ,
I also don't like introducing `glob` by saying it has a "rather unpleasant name" for the reasoning you give and would be glad to see that wording go.

Regarding "package" vs. "library", I believe `glob` is technically a module in the Python standard library, not a package. Library is a vaguer term, that I think works ok in this context because in this lesson we don't get into the detail of what is a module vs. a package vs. a library, and I think that is ok, since we aren't trying to cover every nuance of Python for novice learners. Since we use library to introduce `numpy` and `matplotlib`, I think it may be less confusing to stick with "library" here.

Would you like to open a pull request to remove the "unpleasant name" description?

maxim-belkin:
Hi Sebastian,  thanks for brining this up! 

I concur with you and Lauren on the first issue -- let's fix that unnecessary wording about `glob`. 

Regarding the choice of the right expression for `glob` (package/module/library): as Lauren explained above, `glob` is, technically, a module (it's a single file) but within the scope of our lesson we don't have an opportunity to explain this part of Python nomenclature (that is, there is no need to know it to achieve our goal(s)). In the end, even if modules were called bagels and packages &mdash; donuts, we still would be importing them with `import bagel` and `import donut`. I guess, what I'm really trying to say is that it's more important for a new learner to remember the `import <name>` syntax than the difference between modules and packages which they can learn more about later.

codingS3b:
Hi @maxim-belkin and @ldko, thanks for your quck response and feedback towards that issue and thanks for bringing the module/package issue up and that it might also cause confusion if introduced too early. I was not aware that `glob` is indeed just a single file!
So I will open a PR and change only the part about glob being unpleasant!

maxim-belkin:
[...]

PR #939, opened 2021-06-22
Change wording about glob being an unpleasant name

codingS3b:
As discussed in #937, introducing `glob` by saying it has an unpleasant name is not optimal. Here we change the introductory statement in a way to rather highlight its functionality.

### DIFF
--- a/_episodes/06-files.md
+++ b/_episodes/06-files.md
@@ -12,7 +12,8 @@ keypoints:
 - "Use `*` in a pattern to match zero or more characters, and `?` to match any single character."
 ---
 
-As a final piece to processing our inflammation data, we need a way to get a list of all the files in our data directory whose names start with inflammation- and end with .csv.
+As a final piece to processing our inflammation data, we need a way to get a list of all the files
+in our `data` directory whose names start with `inflammation-` and end with `.csv`.
 The following library will help us to achieve this:
 ~~~
 import glob
--- a/_episodes/06-files.md
+++ b/_episodes/06-files.md
@@ -13,7 +13,7 @@ keypoints:
 ---
 
 We now have almost everything we need to process all our data files.
-The only thing that's missing is a library with a rather unpleasant name:
+The only thing that's missing is a library which provides functionality for filtering files based on their naming scheme:
 
 ~~~
 import glob
--- a/_episodes/06-files.md
+++ b/_episodes/06-files.md
@@ -12,9 +12,8 @@ keypoints:
 - "Use `*` in a pattern to match zero or more characters, and `?` to match any single character."
 ---
 
-We now have almost everything we need to process all our data files.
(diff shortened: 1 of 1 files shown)


_episodes/06-files.md
```diff
@@ -13,7 +13,7 @@ keypoints:
 ---
 
 We now have almost everything we need to process all our data files.
-The only thing that's missing is a library with a rather unpleasant name:
+The only thing that's missing is a library which provides functionality for filtering files based on their naming scheme:
```
ldko, line 17: I think starting the clause with "that" instead of "which" would be proper. Otherwise, I like the wording!

ldko (commented):
I think this is a good change. Thanks a lot for opening this @codingS3b !

_episodes/06-files.md
maxim-belkin: This explanation seems a tiny bit too complex, plus not all files have a naming scheme and `glob` can also return directories.

I suggest to rewrite this sentence and don't talk about a library that is missing but functionality that we need. And what we need is a way to get a list of files from the `data` directory whose names begin with `inflammation-` and end with `.csv`. What do you think?

ldko: I think that works too @maxim-belkin  and would be more directly to the point. Perhaps something along the lines of replacing this sentence and the one above it with:

_As a final piece to processing our inflammation data, we need  a way to get a list of all the files in our `data` directory whose names start with `inflammation-` and end with `.csv`._

codingS3b: That indeed sounds good! I just feel that now a small sentence is missing to actual make the link to the import statement that is following. Could we continue above sentence with "The following library will help us to achieve this"?

ldko: That's a good point @codingS3b.  I think that language makes sense.

_episodes/06-files.md
```diff
@@ -12,9 +12,8 @@ keypoints:
 - "Use `*` in a pattern to match zero or more characters, and `?` to match any single character."
 ---
 
-We now have almost everything we need to process all our data files.
-The only thing that's missing is a library with a rather unpleasant name:
-
+As a final piece to processing our inflammation data, we need a way to get a list of all the files in our data directory whose names start with inflammation- and end with .csv.
```
ldko, line 15: Thanks @codingS3b , I think this text works. I am including a formatting suggestion here that should make it more consistent with our style trends (Keeping lines under 100 characters and using back ticks to set off certain text).
```suggestion
As a final piece to processing our inflammation data, we need a way to get a list of all the files
in our `data` directory whose names start with `inflammation-` and end with `.csv`.
```

ldko (approved):
This looks good to me. Thanks @codingS3b ! :+1:

PR #941, opened 2021-06-24
provide brief inline variable type descriptions

duhaime:
Hello, I thought it could be helpful to add brief inline variable type descriptions when first introducing attendees to different data types. 🎉

### DIFF
--- a/_episodes/01-intro.md
+++ b/_episodes/01-intro.md
@@ -57,9 +57,9 @@ This means that, for example:
 ## Types of data
 Python knows various types of data. Three common ones are:
 
-* integer numbers
-* floating point numbers, and
-* strings.
+* integer numbers (numbers without decimals)
+* floating point numbers (numbers with decimals), and
+* strings (text data).
 
 In the example above, variable `weight_kg` has an integer value of `60`.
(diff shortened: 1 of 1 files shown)


_episodes/01-intro.md
```diff
@@ -57,9 +57,9 @@ This means that, for example:
 ## Types of data
 Python knows various types of data. Three common ones are:
 
-* integer numbers
-* floating point numbers, and
-* strings.
+* integer numbers (numbers without decimals)
+* floating point numbers (numbers with decimals), and
+* strings (text data).
```
ldko, line 62: I think this kind of addition could be helpful. I suggest changing the integer description to "(whole numbers)".  Also, we could link the text for each type to the entries in the glossary which are:

https://swcarpentry.github.io/python-novice-inflammation/reference.html#integer
https://swcarpentry.github.io/python-novice-inflammation/reference.html#floating-point-number
https://swcarpentry.github.io/python-novice-inflammation/reference.html#string

What do you think about adding the descriptions @maxim-belkin ?

maxim-belkin: > What do you think about adding the descriptions @maxim-belkin ?

I like it!

Issue #934, opened 2021-05-18
Episode 8 Creating Functions - About variable scope


chavid:
<details>
<summary><strong>Unexplained key point about reading an outside variable</strong></summary>

The final list of key points include : "If a variable is not defined within the function it is used, Python looks for a definition before the function call".

I am afraid a Python beginner will not understand such a statement, given what was detailed (or not detailed) in the episode, and I find that this information comes late.

Key points are a kind of summary of what was explained during the episode. Am I wrong ?

I strongly support the presence of the exercise called "Variables Inside and Outside Functions", because most python beginners will notice that reading an outside variable works, and wrongly infer that they can also simply modify it from within the function. Yet, showing the modification failure, and not the successful reading of an outside variable, seem to me incomplete, half-way. The final added key point does not really cover the point.

I would review "Variables Inside and Outside Functions" exercise, so to show both the successfull reading of an outside variable, and the failed writing of it. If the lesson is already too long, whatsoever, I would remove the final key point, which comes too late if the concept is not explained before. Don' you think so ?
</details>

ldko:
Thanks for highlighting this @chavid.

Yes, as you say, the key points are meant to outline the concepts introduced in the episode. I think modifying the exercise to make the point of reading an outside variable would be ok to justify the key point's presence. Alternatively, it might be worth extending the variable scope section of the episode to include an example/explanation about this.  Would you like to submit a pull request?

chavid:
Ok. If you think it is ok for the time budget, I think it is really worth to say a word about it in the "variable scope" section. The subsequent question being : to be complete, do we say a word about the keyword "global" ? For the time being, I will make the hypothesis the answer is "no". It is not that bad to let the beginner think he cannot change a global variable, which is a very bad practice most of the cases.

maxim-belkin:
> The subsequent question being : to be complete, do we say a word about the keyword "global" ? For the time being, I will make the hypothesis the answer is "no".

Yes, let's not mention `global` at the moment.

ldko:
This was closed by #942. Thank you @chavid !

PR #942, opened 2021-07-05
Episode 8, functions : about reading a global variable inside a function

chavid:
<details>
<summary><strong>Reading a global variable inside a function</strong></summary>
As discussed in #934, this is a proposal to say a few words abour reading a global variable inside a function, because this is used later on in a challenge of this episode, and finally listed in the key points.
</details>

### DIFF
--- a/_episodes/08-func.md
+++ b/_episodes/08-func.md
@@ -177,6 +177,31 @@ temperature in Kelvin was: 373.15
 ~~~
 {: .output}
 
+On the contrary, inside a function, one can read the value of a variable which has been defined outside the function, before the function is called. We refer to these variables as [global variables]({{ page.root }}/reference.html#global-variable).
+
+In the modified example below, we check the value of the global variable `temp_kelvin` inside a function.
+
+~~~
+def check_temp_kelvin():
+    print('temperature in Kelvin was:', temp_kelvin)
+
+check_temp_kelvin()
+temp_kelvin = fahr_to_kelvin(32.0)
+check_temp_kelvin()
+~~~
+{: .language-python}
+
+~~~
+temperature in Kelvin was: 373.15
+temperature in Kelvin was: 273.15
+~~~
+{: .output}
+
+> ## Avoid global variables
+> 
+> The use of global variables is generally considered as a bad practice, unless for some special cases such as constant values. Especially, never assign a new value to a global variable *inside a function*. This *will not* assign the value to the global variable. Instead, it will assign it to a local variable with the same name, which will disappear once the function is done executing.
+{: .caution}
+
 ## Tidying up
 
 Now that we know how to wrap bits of code up in functions,
--- a/_episodes/08-func.md
+++ b/_episodes/08-func.md
@@ -18,8 +18,8 @@ keypoints:
 - "Call a function using `function_name(value)`."
 - "Numbers are stored as integers or floating-point numbers."
 - "Variables defined within a function can only be seen and used within the body of the function."
-- "If a variable is not defined within the function it is used,
-   Python looks for a definition before the function call"
+- "If a variable is not locally defined within the function it is used,
+   Python looks for a global definition before the function call"
 - "Use `help(thing)` to view help for something."
 - "Put docstrings in functions to provide help for that function."
 - "Specify default values for parameters when defining a function using `name=value`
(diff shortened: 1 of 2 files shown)


_episodes/08-func.md
```diff
@@ -177,6 +177,31 @@ temperature in Kelvin was: 373.15
...
+temp_kelvin = fahr_to_kelvin(32.0)
+check_temp_kelvin()
+~~~
+{: .language-python}
+
+~~~
+temperature in Kelvin was: 373.15
+temperature in Kelvin was: 273.15
+~~~
+{: .output}
```
ldko, line 198: I think this would still make sense if it were shorter. What do you think about taking out the
```
temp_kelvin = fahr_to_kelvin(32.0)
check_temp_kelvin()
```
and 
```
temperature in Kelvin was: 273.15
```

ldko (commented):
Thanks for following up from the issue @chavid ! The review includes my suggestions.

reference.md
```diff
@@ -111,6 +111,12 @@ function
 function call
 :   A use of a function in another piece of software.
 
+global variable
+:   A variable defined outside of a function. It can be used in global statements, and read inside functions.
+
+read , that exists only in the scope of that
+    function, meaning it cannot be accessed by code outside of the function.
+
```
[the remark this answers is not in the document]
chavid: True !

_episodes/08-func.md
```diff
@@ -177,6 +177,31 @@ temperature in Kelvin was: 373.15
 ~~~
 {: .output}
 
+On the contrary, inside a function, one can read the value of a variable which has been defined outside the function, before the function is called. We refer to these variables as [global variables]({{ page.root }}/reference.html#global-variable).
```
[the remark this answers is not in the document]
chavid: I assume this 100 characters limit is about the source mardown ?

_episodes/08-func.md
chavid: I was vaguelay fearing that a beginner may think that the value of `temp_kelvin` is read once for all when the interpreter is going throug the function definition, then reused as is, what new value the variable may get. 

Another misocneption may be that the global variable must exist before the function definition is parsed.

Do you think all those fears are excessive, given all the material that the learner got in previous episodes ?

_episodes/08-func.md
```diff
@@ -177,6 +177,31 @@ temperature in Kelvin was: 373.15
...
+temperature in Kelvin was: 373.15
+temperature in Kelvin was: 273.15
+~~~
+{: .output}
+
+> ## Avoid global variables
+> 
+> The use of global variables is generally considered as a bad practice, unless for some special cases such as constant values. Especially, never assign a new value to a global variable *inside a function*. This *will not* assign the value to the global variable. Instead, it will assign it to a local variable with the same name, which will disappear once the function is done executing.
+{: .caution}
+
```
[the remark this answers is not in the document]
chavid: On one hand, given my original argument about the final keypoint, I agree with you.

On the other hand, it is a pity not to explain one the most frequent trap the python beginners are facing (assign to a freshly created local variable, amtough they think they assign to the global variable).

Well, there is a subsequent challenge that adresses the issue. Probably I should just modify this challenge to insist that the problem is about affecting values, not reading.

Agreed.

_episodes/08-func.md
ldko: Regarding trying to debunk the misconception that the global variable must exist before the function definition is parsed, since there will still already have been a value existing with the name `temp_kelvin` before the function definition, I am not sure how much reassigning it after the definition and first call of `check_temp_kelvin` clarifies that for a beginner. I think to do that, the function could be defined to read a global variable that has not yet been named, then create the variable, then call the function:

```
def print_kelvin():
    print('temperature in Kelvin was:', global_kelvin)

global_kelvin = fahr_to_kelvin(32.0)
print_kelvin()
```

chavid: Yes. And perhaps it would make more clear that the variable is searched for when executing a function call, then fighting also the first misconception. Ok. More to come.

chavid:
New proposal to be reviewed.

_episodes/08-func.md
```diff
@@ -177,6 +177,28 @@ temperature in Kelvin was: 373.15
 ~~~
 {: .output}
 
+The variable `temp_kelvin`, being defined outside any function, is said to be [global]({{ page.root }}/reference.html#global-variable).
```
ldko, line 180: This line of markdown should be split, to keep under the 100 character preference.

ldko (commented):
I think these changes look good! Do you have any further comments @maxim-belkin ?

_episodes/08-func.md
```diff
@@ -18,8 +18,8 @@ keypoints:
 - "Call a function using `function_name(value)`."
 - "Numbers are stored as integers or floating-point numbers."
 - "Variables defined within a function can only be seen and used within the body of the function."
-- "If a variable is not defined within the function it is used,
-   Python looks for a definition before the function call"
+- "If a variable is not locally defined within the function it is used,
+   Python looks for a global definition before the function call"
```
maxim-belkin, line 22: I'm not positive adding "locally" and "global" improves this already rather long keypoint.
Shall we, perhaps, split it into several smaller ones?
1. Variables created outside of any function are called "global" variables.
1. Within a function, we can access global variables.
1. Variables created within a function override global variables if their names match.

chavid: Agreed.

maxim-belkin (approved):

ldko (approved):
This looks good to me. Thank you very much for contributing @chavid ! :+1:

chavid:
Thanks for the reviews.

Issue #944, opened 2021-07-13
How to Contribute to an Open Source Project on GitHub in Contribution Guidelines needs update


leriomaggio:
While looking at the instructions on how to do my first contribution to the Software Carpentries material, I noticed that the link to the `How to Contribute to an Open Source Project on GitHub` is outdated. 

In particular, the link reported in the guidelines is [this](https://egghead.io/series/how-to-contribute-to-an-open-source-project-on-github), whereas the new updated one should be this:[new link](https://app.egghead.io/playlists/how-to-contribute-to-an-open-source-project-on-github)

A PR to fix this issue is coming in a few minutes :)

HTH

leriomaggio:
In the meantime, I've checked all other links in the page and they do all work.

ldko:
This was closed by #945. Thank you @leriomaggio !

PR #945, opened 2021-07-13
Updated link in Contributing instructions

leriomaggio:
This PR updates the link to the "How to Contribute to an Open Source Project on GitHub" article reported in the `CONTRIBUTING.md` document.

The new link is now updated, avoiding the redirection-notice page that was returned instead previously.

This fixes issue #944

Note: I've notice some of the markdown for _italics_ has been changed from asterisk to underscore. I've realised this was done automatically by the Markdown extension in VS Code.
If this is going to cause any issue, I will revert back those changes.


maxim-belkin:
Thank you for the PR, @leriomaggio. Our lesson gets this file from the [lesson template repository](https://github.com/carpentries/styles)when we sync with it. Whitespace and Markdown changes would cause conflicts the next time we sync the lesson with the upstream, so I'd really appreciate it if you could undo them. You can leave the link change in (the fix is in the carpentries/styles already but we can get this fix in before we sync the lesson).

leriomaggio:
Perfect, I thought so that's why I pointed that out - and tbh I noticed of those changes only while I was submitting the PR. 
Will do immediately.

maxim-belkin:
Thank you for your contribution, @leriomaggio! 👍🏻

PR #946, opened 2021-07-15
Update the codes of readings_01, _02, _03 in the .zip

TRUONG3:
I have noticed that the codes in the python-novice-inflammation-code.zip are not up-to-date. In readings_01.py, readings_02.py, readings_03.py, the name of the variable is "row_mean", instead of just "m" as it is in the codes in the .zip.


ldko:
Hi @TRUONG3 , thank you for pointing out the code in some of the files included in the zip file aren't current with what is in this repo's /code directory. It looks like there are some files in addition to the ones you pointed out that are not up-to-date, so I think we should probably make a new zip file from the .py files via the Makefile that is in the repo's /code directory using `make zip`.

PR #947, opened 2021-07-15
Modify the second objective of the episode 03-matplotlib

TRUONG3:
I think it's a bit better to change the second objective of this episode from "Group several graphs ..." to "Plot a group several graphs ..."

### DIFF
--- a/_episodes/03-matplotlib.md
+++ b/_episodes/03-matplotlib.md
@@ -7,7 +7,7 @@ questions:
 - "How can I group several plots together?"
 objectives:
 - "Plot simple graphs from data."
-- "Group several graphs in a single figure."
+- "Plot a group several graphs in a single figure."
 keypoints:
 - "Use the `pyplot` module from the `matplotlib` library for creating simple visualizations."
 ---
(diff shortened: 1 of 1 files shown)


ldko:
Hi @TRUONG3 ,
Thank you for taking the time to suggest a change to the objective.  To help us know what you would like to see clarified, could you please share your reasoning for changing the text?

TRUONG3:
Hi, I think the second objective should be changed in this way to better parallel with the 1st objective, which is firstly doing the coding to plot a single graph and then secondly to plot a group of graphs. The coding is indeed for plotting as well, not just grouping the graphs.

_episodes/03-matplotlib.md
```diff
@@ -7,7 +7,7 @@ questions:
 - "How can I group several plots together?"
 objectives:
 - "Plot simple graphs from data."
-- "Group several graphs in a single figure."
+- "Plot a group several graphs in a single figure."
```
ldko, line 10: Perhaps we could cut this down to:
```suggestion
- "Plot multiple graphs in a single figure."
```

ldko (commented):
Hi @TRUONG3 , Thanks for giving us the reasoning behind your change. What you said makes sense to me about emphasizing plotting over grouping only.

_episodes/03-matplotlib.md
TRUONG3: yes, better so

ldko:
Thank you for contributing an improvement to the lesson @TRUONG3 !

PR #958, opened 2021-09-22
Solution should have the answer not options

ytakemon:
Some minor changes. I noticed the solutions dropdown had various options for solutions, which included several common incorrect answers. It wasn't clear what the correct solution was. I can understand if the intent was to have a multiple-choice question, but it was not formatted in that way.

### DIFF
--- a/_episodes/01-intro.md
+++ b/_episodes/01-intro.md
@@ -253,9 +253,6 @@ stuck on it](../fig/python-sticky-note-variables-03.svg)
 >
 > > ## Solution
 > > ~~~
-> > `mass` holds a value of 47.5, `age` does not exist
-> > `mass` still holds a value of 47.5, `age` holds a value of 122
-> > `mass` now has a value of 95.0, `age`'s value is still 122
 > > `mass` still has a value of 95.0, `age` now holds 102
 > > ~~~
 > > {: .output}


maxim-belkin:
Hi @ytakemon. Each line describes the evolution of variables/values and not the final state. Note that there are 4 lines of code and 4 lines in the "Solutions" section.

ytakemon:
Oh, I see! Sorry for the misunderstanding.

maxim-belkin:
No worries. Do you think there is a way to improve this part to avoid this confusion?

ytakemon:
Perhaps if the question was written out as: 

For each of the following statements, what values do the variables `mass` and or `age` hold? Test your answer by executing the lines.

1. `mass = 47.5`
2. `age = 122`
3. `mass = mass * 2.0`
4. `age = age - 20`

then the solutions as: 

1. `mass` holds a value of 47.5, `age` does not exist
2. `mass` still holds a value of 47.5, `age` holds a value of 122
3. `mass` now has a value of 95.0, `age`'s value is still 122
4. `mass` still has a value of 95.0, `age` now holds 102

I initially misunderstood the question to be: after executing all 4 lines of code what is the final value in `mass` and `age`. I think this way it might be clearer for the learner to understand we want to know what the answer is for each and which solution corresponds to what line of code.

PR #971, opened 2022-01-05
Add explicit parameters for numpy.delete

ndporter:
The alternative solution of using numpy.delete() to remove a column from an array is helpful, but the meaning of the arguments is not obvious without explicitly including the names. Even then, "obj" is non-obvious, so I also added a note in the solution description reminding them to use help files if necessary.

<details>

\- The Carpentries
</details>

### DIFF
--- a/_episodes/02-numpy.md
+++ b/_episodes/02-numpy.md
@@ -662,10 +662,11 @@ which is the average inflammation per patient across all days.
 > > ## Solution
 > >
 > > An alternative way to achieve the same result is to use Numpy's
-> > delete function to remove the second column of A.
+> > delete function to remove the second column of A. If you're not
+> > sure what the parameters of numpy.delete mean, use the help files.
 > >
 > > ~~~
-> > D = numpy.delete(A, 1, 1)
+> > D = numpy.delete(arr=A, obj=1, axis=1)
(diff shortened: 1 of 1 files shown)


chillenzer:
Could you expand a little on your first statement "The alternative solution of using `numpy.delete()` to remove a column from an array is helpful"? From my experience, I have not a single examples where the usage of `numpy.delete` is justified because it creates a new array instead of returning a view.

In the best case, the performance penalty does not matter and you never look at your code again and think "Hey, that element was deleted from the original array. So, it won't be there any more." Well, it is! You are bound to get surprised at least once by the fact that your original array didn't change at all and only if you are very lucky, that will not translate into a hard to track bug in your code where the number are slightly off all the time.

In the worst case, an inexperienced python user writes a `for`-loop to delete all the elements they don't want instead of using a mask or proper slicing to return a view. If you do that on any reasonably sized data, you will immediately abandon `python` for being almost as slow as doing it by hand. And then again, they might forget to assign the *created copy* (as opposed to an in-place change) back to the original variable and all that.

So, from my perspective `numpy.delete` does more harm than good. Even if you definitely need a copy instead of a view, you should do yourself a favour and state that explicitly by use of `numpy.copy`.

PR #893, opened 2020-12-10
edited code, variable used before instantiated

GeorgeSavvidis:
<details>
<summary><strong>Instructions</strong></summary>

Thanks for contributing! :heart:

If this contribution is for instructor training, please email the link to this contribution to
checkout@carpentries.org so we can record your progress. You've completed your contribution
step for instructor checkout by submitting this contribution!

Keep in mind that **lesson maintainers are volunteers** and it may take them some time to
respond to your contribution. Although not all contributions can be incorporated into the lesson
materials, we appreciate your time and effort to improve the curriculum. If you have any questions
about the lesson maintenance process or would like to volunteer your time as a contribution
reviewer, please contact The Carpentries Team at team@carpentries.org.

You may delete these instructions from your comment.

\- The Carpentries
</details>

PR #876, opened 2020-10-07
Revision to Code Sample

mdsimpson42:
I tried to do a live code demo of this exercise during my Carpentries Instructor Training and noticed that this first piece of code will not run as it is currently written (as the variable "data" is not initialised before it is used). I have added the two additional lines of code necessary to make this code sample work. 

(This is part of my Instructor Training checkout process)

<details>
<summary><strong>Instructions</strong></summary>

Thanks for contributing! :heart:

If this contribution is for instructor training, please email the link to this contribution to
checkout@carpentries.org so we can record your progress. You've completed your contribution
step for instructor checkout by submitting this contribution!

Keep in mind that **lesson maintainers are volunteers** and it may take them some time to
respond to your contribution. Although not all contributions can be incorporated into the lesson
materials, we appreciate your time and effort to improve the curriculum. If you have any questions
about the lesson maintenance process or would like to volunteer your time as a contribution
reviewer, please contact The Carpentries Team at team@carpentries.org.

You may delete these instructions from your comment.

\- The Carpentries
</details>

PR #986, opened 2022-04-01
re-import data at start of plotting

brownsarahm:
This makes the individual episodes a little more self-containted and makes it easier for instructors who may swap out between episdoes.  

It also will mean that we do not need to dis-allow this episode for teaching demos [see PR](https://github.com/carpentries/instructor-training/pull/1346/commits/5ecd1baf934afa0a386c74ee82e29027f7ed8d0c)

### DIFF
--- a/_episodes/03-matplotlib.md
+++ b/_episodes/03-matplotlib.md
@@ -22,6 +22,7 @@ import the `pyplot` module from `matplotlib` and use two of its functions to cre
 
 ~~~
 import matplotlib.pyplot
+import numpy
 data = numpy.loadtxt(fname='inflammation-01.csv', delimiter=',')
 image = matplotlib.pyplot.imshow(data)
 matplotlib.pyplot.show()
--- a/_episodes/03-matplotlib.md
+++ b/_episodes/03-matplotlib.md
@@ -20,10 +20,22 @@ there is no official plotting library, `matplotlib` is the _de facto_ standard. 
 import the `pyplot` module from `matplotlib` and use two of its functions to create and display a
 [heat map]({{ page.root }}/reference.html#heat-map) of our data:
 
+> ## Making sure we have the data
+>
+> If you are continuing inthe same notebook from the previous episode you already
+> have a `data` variable and have imported `numpy`.  If you are starting a new 
+> notebook at this point you need the following two lines. 
+>
+> ~~~
+> import numpy
+> data = numpy.loadtxt(fname='inflammation-01.csv', delimiter=',')
+> ~~~
+> {: .source}
+{: .callout}
+
+
 ~~~
 import matplotlib.pyplot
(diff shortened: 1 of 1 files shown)


ldko:
Note, this overlaps with #893 and #876. I think it would be good to get some sort of fix out for this as it clearly has come up for multiple people, but I am no longer a maintainer here. ping @maxim-belkin

maxim-belkin:
Making episodes more self-contained is a good thing, so I'm +1 here. Following my comment from 2020 (https://github.com/swcarpentry/python-novice-inflammation/pull/876#issuecomment-706258785), I would suggest two things:
1. Put this in a prereq (or callout) box at the beginning of the episode.
2. Provide a brief explanation on when the proposed steps (importing numpy + loading data) are necessary.

brownsarahm:
@maxim-belkin does that meet your request?

_episodes/03-matplotlib.md
```diff
@@ -20,6 +20,20 @@ there is no official plotting library, `matplotlib` is the _de facto_ standard.
 import the `pyplot` module from `matplotlib` and use two of its functions to create and display a
 [heat map]({{ page.root }}/reference.html#heat-map) of our data:
 
+> ## Making sure we have the data
+>
+> If you are continuing inthe same notebook from the previous episode you already
+> have a `data` variable and have imported `numpy`.  If you are starting a new 
+> notebook at this point you need the following two lines.
```
maxim-belkin, line 27:
```suggestion
> notebook at this point, you need the following two lines:
```

_episodes/03-matplotlib.md
```diff
@@ -20,6 +20,20 @@ there is no official plotting library, `matplotlib` is the _de facto_ standard.
...
+>
+> If you are continuing inthe same notebook from the previous episode you already
+> have a `data` variable and have imported `numpy`.  If you are starting a new 
+> notebook at this point you need the following two lines. 
+>
+> ~~~
+> import numpy
+> data = numpy.loadtxt(fname='inflammation-01.csv', delimiter=',')
+> ~~~
+> {: .source}
```
maxim-belkin, line 33:
```suggestion
> {: .language-python}
```

maxim-belkin (changes requested):
Thanks, @brownsarahm. Looks good! A few minor fixes and it will be good to go.
I also wonder if this note should be a `.prereq` instead of a `.callout` -- in that case we could standardize the title to "Episode prerequisites" and later use this title in other episodes (if and when necessary).

brownsarahm:
I do like episode prereqs. Espeically if we can promote that as a broader trend across lesson. Will change.

maxim-belkin (approved):
Thank you so much for this contribution, Sarah! Looks great!

PR #987, opened 2022-04-09
02-numpy.md: fix broken URL

amangoel185:
Fix the URL for `floating-point number`


maxim-belkin (approved):
Thank you for fixing the link, Aman!

PR #991, opened 2022-06-10
fix: use normal print instead of percent formatting

henryiii:
This was using the old percent formatting instead of the more modern `.format` (Python 2.6+) or f-strings (Python 3.6+). What's more, string formatting is not introduced or discussed anywhere else in the lesson, and it is not elaborated here. I think the simplest thing for this is to simply not introduce it here, it's not needed, and multi-argument print calls have been introduced.

### DIFF
--- a/_episodes/11-debugging.md
+++ b/_episodes/11-debugging.md
@@ -268,7 +268,7 @@ not more.
 > for patient in patients:
 >     weight, height = patients[0]
 >     bmi = calculate_bmi(height, weight)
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Hi, @henryiii. The change you propose here makes sense to me and IMHO can be merged as is. Unfortunately, I no longer have write access to this repository and I don't know who or when can look at / merge this PR.

CC @fmichonneau

PR #994, opened 2022-06-10
chore: clarify cli lesson

henryiii:
IMO, this lesson is teaching the wrong thing - CLI arguments should not be handled by hand! I've at least updated the main issue with this - it's critical to know about (and provide) `--help` for all CLI interfaces!


noatgnu:
Hi @henryiii ,

Thank you for submitting the pull request. We have discuss about it however we do not think that any changes are required. However if you feel strongly about the request, please open an issue and rephrase the needed changes.

henryiii:
The key issue is that this lesson is teaching a bad practice (manually parsing command line options) without listing the most important drawback of this approach - `--help` should _always_ be present in any app you can run. This is an important feature users should be able to rely on always being present - no app should run arbitrary code if `--help` is passed to it! This PR was just including the mention, it was not rewriting the lesson to do something different.

PR #1005, opened 2022-08-08
Remove dictionary references from 09-errors.md

### LINKED ISSUE #783: Dictionaries not introduced but used in exercise in "Error and Exceptions" - & suggestiong

patrick-austin:
Based on the feedback on #809, rather than introducing the concept of a `dict` have instead modified the exercise to use a `list` and an `IndexError`. The solution to the exercise has been modified to reflect this, and have changed `print_friday_message` to `print_sunday_message` to make the mistake of providing `7` as the index more plausible.

The original issue also made points about moving the exercise earlier. Given this wasn't addressed in the PR comments I assume that this is fine.

closes #783


knarrff:
I stumbled over the same issue in my own course yesterday and just came here with the intent of opening a very similar pull request myself only to find that there is already one open (thanks @patrick-austin). It would be nice to get this merged or some reaction from a maintainer? I does not create an inclusive feeling to leave pull requests open for so long without any reaction.

zkamvar:
> It would be nice to get this merged or some reaction from a maintainer? I does not create an inclusive feeling to leave pull requests open for so long without any reaction.

Hello @knarrff, 

I am a member of @swcarpentry/staff-curriculum and I hope I can provide some context: the maintainers for these lessons are working on a volunteer basis and the current maintainers inherited a very large backlog of pull requests last year. They are indeed trying to get through as many pull requests as they can while keeping the lesson up-to-date. The maintainer community has only recently organised and we on the core team are working as hard as we can to support them. If you have any questions or concerns, feel free to email us at curriculum@carpentries.org.

All the best,
Zhian

PR #1007, opened 2022-09-07
Removing use of gendered term layman

### LINKED ISSUE #966: Episode 01: avoid gendered term "layman"
https://swcarpentry.github.io/python-novice-inflammation/01-intro/index.html

In Episode 01 we have "in layman's terms..." which could be stated in a more neutral way. I'm sure it's rarely or never found to be offensive, but we're striving for a high standard.

We could say "in a layperson's terms..." or completely revise since arguably "a name for storing a value" may not be considered lay-speech any more than "variable" is.  

Perhaps:
 "In essence, a variable is a named location for storing, modifying, and recalling a value."

matt-graham:
Fixes #966 

Changes gendered term layman to non-gendered layperson. 

I think @matthewabrown's suggestion in #966 to instead avoid the use of lay[man/person] altogether (which I guess could be considered demotivational language in the sense implying the description should be easy to understood which might not be the case for everyone) by replacing the sentence with

> In essence, a variable is a named location for storing, modifying, and recalling a value

or some variant of this has merit, but as this would be a more significant change I've just made a minimal change to swap layman to layperson for now as this seems unlikely to be controversial.

### DIFF
--- a/_episodes/01-intro.md
+++ b/_episodes/01-intro.md
@@ -42,7 +42,7 @@ weight_kg = 60
 {: .language-python}
 
 From now on, whenever we use `weight_kg`, Python will substitute the value we assigned to
-it. In layman's terms, **a variable is a name for a value**.
+it. In layperson's terms, **a variable is a name for a value**.
 
 In Python, variable names:
 


ineelhere:
Hello @matt-graham . Thank you for the suggestion and raising the PR. 
Merging the same to make changes reflect on the lesson page.

Keep contributing and have a great day ahead!

Issue #1012, opened 2022-09-22
Episode 10 Solution incorrectly explains second assertion


patrick-austin:
https://github.com/swcarpentry/python-novice-inflammation/blob/gh-pages/_episodes/10-defensive.md#testing-assertions
Regarding this code:
> `assert int(element)`

The solution states:
> The second assertion checks that each value in the list can be turned into an integer. Input such as `[1, 2,'c', 3]` will make it fail.

While an input of `'c'` would make this "fail", the failure is due to a `ValueError` being raised by `int()`, not an `AssertionError`. What this line of code actually does is assert that `bool(int(element))` is `True`. For example, an input of `[True]` is fine because it is cast to `1` which is thruthy, but an input of `[0]` will fail this assertion as it evaluates to `False`. Either the explanation in the solution or the example code should change. 

If we do not want to allow `0`, the solution should be explicit that this is what we are testing for (and the assert should probably be changed to `assert int(element) != 0`  in order to be easily understood by learners). Alternatively, and as #692 suggests, the assertion could be `assert int(element) > 0` given a number of cars cannot be negative either.

If we only want to check we can cast to an integer, then this code isn't needed as:
> values = [int(element) for element in values]

would raise the `ValueError` regardless.

JamesAllsopp:
Propose that;
```
assert int(element)
```
be changed to:
```
assert str(element).isdigit()
```

This works if the element is not empty, handles floating point, and doesn't use exceptions. I'll add a MR for this later if people think this is better,

This shows the testing and demonstrates the original error;
```
Python 3.10.8 (main, Oct 13 2022, 21:13:48) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> test = [1, 2,'c', 3]

>>> for e  in test:
...     assert str(e).isdigit()
...
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
AssertionError
>>> test1 = [1, 2, 3]
>>> for e  in test1:
...     assert str(e).isdigit()
...
>>> test2 = [1,2,'',3]
>>> for e  in test2:
...     assert str(e).isdigit()
...
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
AssertionError
>>> for e  in test:
...     assert int(e)
...
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
ValueError: invalid literal for int() with base 10: 'c'
```

PR #1016, opened 2022-10-19
Change in response to the problem identified in Issue \#1012

JamesAllsopp:
This fixes the issue [1012](https://github.com/swcarpentry/python-novice-inflammation/issues/1012) by changing one piece of example code and providing additional examples. The issue includes output showing that this works, and is robust against other errors, e.g. floating point values and empty values in the list.


zkamvar:
Regarding the transition: this will happen in May and here are instructions what to do with your fork after the transition: https://carpentries.github.io/workbench/faq.html#github-forks

PR #1025, opened 2023-02-28
Do not mention "exceptions"

knarrff:
The section currently is called 'Errors and Exceptions', but the section itself only mentions 'errors' and just once 'errors and exceptions'. There might be an argument here about what the difference is between both, but neither is it currently made in the material, nor do I think it should in an introductory course like this.

Because of this, I propose to remove the mention of 'exceptions' entirely from this course (only 2 places), and especially from a section heading, to avoid unnecessary confusion by new learners.

<details>

\- The Carpentries
</details>


noatgnu:
Hi @knarrff,

Thank you for this submission. While we did not utilize exceptions explicitly, the official document does mention the usage of exceptions so by mentioning exceptions in the lesson, the learner will also have some ideas of what they might see in the official documentation. Also, the built-in error messages are also exceptions that have been raised within Python itself so I don't think we should remove the mention of exceptions from the lesson.

PR #1026, opened 2023-03-03
use list instead of tuple in example

knarrff:
Lists have been introduced in their own episode, while tupes are at this point new to learners and do not receive a proper introduction. I do not think the fine points of the differences between lists and tuples should be explained to beginners at this point, nor do I feel there would be time to do that within the regular course time. In addition, this example works just as well with lists, including reaching all learning objectives of defensive programming. Also, tuples are not used anywhere else in the course as far as I can see.

I this propose to let this example use a list as input and output of the function instead of a tuple. This removes the time needed to explain that this is yet another datatype and that it is really similar to lists, but a little different, but that this difference does not matter for this example. I did not change the time needed though because it is my feeling that the suggested time is not too long in any case.

<details>

\- The Carpentries
</details>

### DIFF
--- a/_episodes/10-defensive.md
+++ b/_episodes/10-defensive.md
@@ -107,8 +107,8 @@ assertions fall into three categories:
     is something that is always true at a particular point inside a piece of code.
 
 For example,
-suppose we are representing rectangles using a [tuple]({{ page.root }}/reference.html#tuple)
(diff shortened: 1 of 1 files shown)


zkamvar:
Regarding the transition: this will happen in May and here are instructions what to do with your fork after the transition: https://carpentries.github.io/workbench/faq.html#github-forks

PR #1031, opened 2023-03-30
[workbench transition] Update setup.md

zkamvar:
I'm updating these links because the commonmark parser does not recognise them and my intervention did not account for this situation. This is the only place in this or any of the other lessons I have modified where this happens, so it's easier for me to update this than to update my intervention and add tests.

FWIW, the link syntax in The Workbench is _much_ simpler:

https://carpentries.github.io/workbench/transition-guide.html#internal-links


ineelhere (approved):

ineelhere:
Thank you @zkamvar. PR #1031 has been merged!

PR #1035, opened 2023-05-02
Fix formatting issues on multi-line string

VictorGambarini:
Fix formatting issues on multi-line string on readings_08.py. The multi-line string should not be indented, as it causes extra whitespace to be printed when the message is displayed. Also updated the whole string to look better.

---

---

### DIFF
--- a/episodes/12-cmdline.md
+++ b/episodes/12-cmdline.md
@@ -816,9 +816,13 @@ def main():
     script = sys.argv[0]
     if len(sys.argv) == 1: # no arguments, so print help message
         print("""Usage: python readings_08.py action filenames
-              action must be one of --min --mean --max
-              if filenames is blank, input is taken from stdin;
-              otherwise, each filename in the list of arguments is processed in turn""")
+
+action:
+    Must be one of --min, --mean, or --max.
+
+filenames:
+    If blank, input is taken from standard input (stdin).
+    Otherwise, each filename in the list of arguments is processed in turn.""")
         return
 
     action = sys.argv[1]


vinisalazar (changes requested):
Hi @VictorGambarini,

Thank you for your contribution. I agree that unnecessary whitespace should be removed from multiline strings like that one.

However, it seems weird to me to remove all indentation from the string. Although this is not a docstring, [PEP-8](https://peps.python.org/pep-0008/#string-quotes) states that "For triple-quoted strings, always use double quote characters to be consistent with the docstring convention in [PEP 257](https://peps.python.org/pep-0257).". That PEP states that ["The entire docstring is indented the same as the quotes at its first line [...]."](https://peps.python.org/pep-0257/#multi-line-docstrings).

Do you think you could modify your PR to account for that? I believe that PEP 257 offers scripts to format these strings, or you could use a code formatting tool such as [black](https://github.com/psf/black).

Let me know if you need any assistance!

Thank you.
Vini

episodes/12-cmdline.md
```diff
@@ -816,9 +816,13 @@ def main():
...
-              action must be one of --min --mean --max
-              if filenames is blank, input is taken from stdin;
-              otherwise, each filename in the list of arguments is processed in turn""")
+
+action:
+    Must be one of --min, --mean, or --max.
+
+filenames:
+    If blank, input is taken from standard input (stdin).
+    Otherwise, each filename in the list of arguments is processed in turn.""")
```
ineelhere, line 825: Thanks for your contribution 😃 
Keeping code readability as a priority, it would be great if you could please shift these codes to the right for a consistent indentation.

ineelhere (changes requested):
Thanks for your comment and suggestion @vinisalazar. I too agree with you.

Hi @VictorGambarini. Thanks for the formatting suggestions. Keeping code readability as a priority, it would be great if you could please modify your PR with consistent indentation. 😃

VictorGambarini:
Hi @vinisalazar and @ineelhere,

Thank you for the very fast reply. :)

I discovered that the same file is in two places: "episodes/12-cmdline.md" and "episodes/files/code/readings_08.py."

The files were a little different in both places, so I updated both to have the exact same code.

I'm happy to contribute.

Cheers

ineelhere (approved):

PR #1045, opened 2023-06-22
Update 10-defensive.md

minyoungJYi:
I didn't know how to do the fancier mark up.  But I thought decomposing the assert function was part of the SWC ethos.

### DIFF
--- a/episodes/10-defensive.md
+++ b/episodes/10-defensive.md
@@ -55,7 +55,8 @@ and the most common way to do it is to add
 [assertions](../learners/reference.md#assertion) to our code
 so that it checks itself as it runs.
 An assertion is simply a statement that something must be true at a certain point in a program.
-When Python sees one,
+The structure of an assertion is "assert [test condition] [helpful error message].
+ When Python sees one,
 it evaluates the assertion's condition.
 If it's true,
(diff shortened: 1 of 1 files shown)


episodes/10-defensive.md
```diff
@@ -55,7 +55,8 @@ and the most common way to do it is to add
 [assertions](../learners/reference.md#assertion) to our code
 so that it checks itself as it runs.
 An assertion is simply a statement that something must be true at a certain point in a program.
-When Python sees one,
+The structure of an assertion is "assert [test condition] [helpful error message].
```
ineelhere, line 58: Thanks for the contribution. 
A trailing `"` after `message]` could be introduced to close the unclosed opening quotation mark before `assert`. 😃 

Please feel free to use the following statement as a fix
```
"assert [test condition] [helpful error message]"
```

ineelhere (changes requested):

episodes/10-defensive.md
edbennett: The "fancier formatting" suggested would be to use a backtick `` ` `` before and after the code text instead of the double quote `"`.

PR #1046, opened 2023-06-23
Update 11-debugging.md

minyoungJYi:
The solution isn't consistent with the code in the question.

Doing this as a part of instructor checkout.

### DIFF
--- a/episodes/11-debugging.md
+++ b/episodes/11-debugging.md
@@ -280,7 +280,7 @@ def calculate_bmi(weight, height):
 for patient in patients:
     weight, height = patients[0]
     bmi = calculate_bmi(height, weight)
-    print("Patient's BMI is:", bmi)
+    print("Patient's BMI is: %f", %f bmi)
 ```
 
 ```output


episodes/11-debugging.md
```diff
@@ -280,7 +280,7 @@ def calculate_bmi(weight, height):
 for patient in patients:
     weight, height = patients[0]
     bmi = calculate_bmi(height, weight)
-    print("Patient's BMI is:", bmi)
+    print("Patient's BMI is: %f", %f bmi)
```
ineelhere, line 283: the formatting placeholder `%f` is placed incorrectly. To format the `bmi` variable correctly, it should be placed inside the string using the `%` operator.
```
print("Patient's BMI is: %f" % bmi)
```

ineelhere (changes requested):
Thanks for the contribution and the PR @minyoungJYi 😃 

There is a syntax error in the code, please check my comments for the same.

edbennett:
It's no longer obvious in the "blame" view because of the lesson transformation, but this change would effectively revert #991.

Since the lesson doesn't teach the `%` syntax for string interpolation at any point, I'd suggest that the solution not being changed was an oversight in #991, and it would be preferable to adjust the solution to match the exercise rather than the other way around.

PR #1048, opened 2023-07-07
Update 08-func.md

minyoungJYi:
This is a really wordy/confusing way to present the same piece of code.

### DIFF
--- a/episodes/08-func.md
+++ b/episodes/08-func.md
@@ -48,7 +48,7 @@ Let's start by defining a function `fahr_to_celsius` that converts temperatures
 from Fahrenheit to Celsius:
 
 ```python
-def explicit_fahr_to_celsius(temp):
+def fahr_to_celsius(temp):
     # Assign the converted value to a variable
     converted = ((temp - 32) * (5/9))
     # Return the value of the new variable


ineelhere (commented):
Thank you @minyoungJYi for the PR. 

There are 2 different functions mentioned in the chapter. Changing the function `explicit_fahr_to_celsius` to `fahr_to_celsius` will result in 2 functions with the same name `fahr_to_celsius`. 

This shall confuse/impact the graphic


The two functions, `explicit_fahr_to_celsius(temp)` and `fahr_to_celsius(temp)`, both convert a temperature from Fahrenheit to Celsius. However, there are some differences in how they are implemented.

In the `explicit_fahr_to_celsius(temp)` function:

It assigns the converted value to a variable named converted.
It returns the value of the converted variable.
In the `fahr_to_celsius(temp)` function:

It directly returns the converted value without assigning it to a variable.
The main difference lies in the handling of the converted value and the return statement. 😃 

```
def explicit_fahr_to_celsius(temp):
    # Assign the converted value to a variable
    converted = ((temp - 32) * (5/9))
    # Return the value of the new variable
    return converted
    
def fahr_to_celsius(temp):
    # Return converted value more efficiently using the return
    # function without creating a new variable. This code does
    # the same thing as the previous function but it is more explicit
    # in explaining how the return command works.
    return ((temp - 32) * (5/9))
```

ineelhere (changes requested):
Thank you for your contribution 😃 
Please refer to the [comments here](https://github.com/swcarpentry/python-novice-inflammation/pull/1048#pullrequestreview-1526985493).

PR #1057, opened 2023-07-27
use four spaces for indentation, not two or three

edbennett:
I spotted while delivering this lesson that while the convention of using four spaces for indentation is used almost everywhere, one code block in one episode uses two spaces (without any other comment).

This convention is universally adopted enough that I think it's reasonable to use it exclusively. (If we were going to mix conventions then we would want to have a lot more explanation than is currently present.)

Based on this observation, I've systematically checked the full lesson and made all spacing consistent. There is one remaining place where spacing is inconsistent, but this is deliberate to cause an `IndentationError` in episode 09.


ineelhere (approved):
Thank you, @edbennett, for the contribution with uniformity in indentations.

PR #1058, opened 2023-07-27
adjust wording of range exercise; fix #1053

### LINKED ISSUE #1053: Strange wording in episode 05 challenge
A [challenge in the episode on loops](https://swcarpentry.github.io/python-novice-inflammation/instructor/05-loop.html#from-1-to-n) has the sentence

> Using `range`, write a loop that uses `range` to print the first 3 natural numbers:

Could this sentence be rephrased to remove either "Using `range`" or "that uses `range`"?

edbennett:
Adjusted the wording of the exercise using `range` in episode 5, to fix the confusion raised in #1053 

I chose to keep the "Using `range`", since the other wording might imply that `range` should do the printing. (I recently encountered a learner trying to call `range` directly for this challenge rather than using a loop with `print` in the body.)

### DIFF
--- a/episodes/05-loop.md
+++ b/episodes/05-loop.md
@@ -268,7 +268,7 @@ accept 1, 2, or 3 parameters.
   For example, `range(3, 10, 2)` produces `3, 5, 7, 9`.
 
 Using `range`,
-write a loop that uses `range` to print the first 3 natural numbers:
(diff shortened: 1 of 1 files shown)


ineelhere (approved):
Thank you for contributing to this lesson, @edbennett . Best wishes 🙂

PR #1060, opened 2023-07-27
New python branch

falavigit:
With defining shorter name for python libraries in the import line, we can save time in coding, as they are called frequently in codes. 
import numpy as np 
import matplotlib as plt 
---

---


ineelhere (commented):
Hello @falavigit, thank you for the suggestion and PR. Highly appreciate your efforts.

Although it is great to use abbreviations for the imported libraries, for this particular lesson, it would be a bit more helpful for new/first time learners to easily understand the usage of the library functions if the full name is used. 🙂

Thank you again for your efforts, happy coding!

PR #1066, opened 2023-10-13
Add callout explaining function names amax, amin, and mean

fiveop:
I was about to create a pull request changing all references to amax and amin to max and min (as they are equivalent according to the code and documentation), but then I found a merged PR that did the opposite: #995.
It explains the reasoning behind it.

To prevent someone else from starting to look into it again and to provide an explanation for teachers and learners why there is `amax` and `amin` but `mean` without `a`, I propose to add an explanatory callout section.

If you think an instructor note would be better/sufficient, I would change the PR accordingly, since I am not sure myself.

### DIFF
--- a/episodes/02-numpy.md
+++ b/episodes/02-numpy.md
@@ -376,6 +376,18 @@ for example: `help(numpy.cumprod)`.
 
 ::::::::::::::::::::::::::::::::::::::::::::::::::
 
+:::::::::::::::::::::::::::::::::::::::::  callout
+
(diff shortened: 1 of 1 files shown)


tmorrell (approved):
The change to amean/amin caught me by surprise in a workshop, and having the explanation in the workshop text would be very helpful. This text sounds good to me.

ineelhere (approved):
This added explanation should be very helpful to learners. Thanks for the contribution @fiveop! 🚀

PR #1067, opened 2023-10-13
Do not refer to NumPy array using the term list

fiveop:
The description of the exercise used the term list, while the code in the solution used functions that we have so far only used with NumPy arrays (`amax`, `amin`). Even though these functions do take arguments of type `list`, learners might get confused (and teachers do get confused, because I thought up until I tested it right now, that surely `numpy.amin` etc. do not work on `list` type objects… :)).

### DIFF
--- a/episodes/10-defensive.md
+++ b/episodes/10-defensive.md
@@ -467,7 +467,7 @@ can you think of a function that will pass your tests but not his/hers or vice v
 
 ```python
 # a possible pre-condition:
-assert len(input_array) > 0, 'List length must be non-zero'
+assert len(input_array) > 0, 'Array length must be non-zero'
 # a possible post-condition:
 assert numpy.amin(input_array) <= average <= numpy.amax(input_array),
 'Average should be between min and max of input values (inclusive)'
--- a/episodes/10-defensive.md
+++ b/episodes/10-defensive.md
@@ -456,7 +456,7 @@ This violates another important rule of programming:
 ## Pre- and Post-Conditions
 
 Suppose you are writing a function called `average` that calculates
-the average of the numbers in a list.
+the average of the numbers in a NumPy array.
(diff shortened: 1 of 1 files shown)


episodes/10-defensive.md
```diff
@@ -467,9 +467,9 @@ can you think of a function that will pass your tests but not his/hers or vice v
 
 ```python
 # a possible pre-condition:
-assert len(input_list) > 0, 'List length must be non-zero'
+assert len(input_array) > 0, 'List length must be non-zero'
```
ineelhere, line 470: Thanks @fiveop for the suggestion and contribution!
Please note that there is one more `list` term to be replaced with `numpy array` on this line.

While reviewing the code, I'd like to suggest considering the data type of the variable `input_list`. If it happens to be a standard Python list instead of a NumPy array, it might be beneficial to consistently refer to it as a 'list' in our comments to maintain clarity. The decision between 'list' and 'array' should align with the actual type of the variable at hand. 

To enhance comprehension for both teachers and students, it might be helpful to include a brief comment or explanation at certain points in the code. As you've initiated the pull request, I wonder if it might be appropriate to insert comments or descriptions where you think they would be most effective. Please feel free to do so, and if there's any need for assistance from a maintainer's end, please don't hesitate to reach out. 🙂

ineelhere (changes requested):

episodes/10-defensive.md
fiveop: I changed the assertion message to use the term array instead of list.

I am not sure whether I understand correctly, where you suggest to add further comments. If it is in the solution of this exercise, I would suggest to not do that, because it would distract from what the exercise is about: thinking about pre- and post-conditions.

ineelhere (approved):

PR #1068, opened 2023-10-13
Use correct function name

fiveop:
The function is first called `get_total_cars`, which fits to the problem description, then `get_total`. This pull request fixes that.


ineelhere (approved):
Thanks for the fix @fiveop 🚀 
Best wishes! 🙂

PR #1069, opened 2023-10-13
Inform instructors about potential syntax error

fiveop:
We ran into the problem described in the proposed instructor note when we taught based on the course material recently.

It should be in front of the first example that potentially causes the error.


ineelhere (approved):
Thanks for the contribution, @fiveop 🚀 
Best wishes!

PR #1076, opened 2024-01-03
Fix "FIXME" in Glossary page title

maneesha:
The `reference.md` file that contains the glossary has "FIXME" as a title, and has the "Glossary" heading twice.  This PR fixes both of those issues.


ineelhere (approved):
LGTM! Thanks for the update @maneesha 👍🏼 
This is the first PR of 2024 🥇 

Happy New Year, community! 🚀

Issue #979, opened 2022-03-04
A concrete example of moving from assertions to exceptions


DamienIrving:
Hi! I'm the lead maintainer of the Data Carpentry [Python for Atmosphere and Ocean Science](https://carpentries-lab.github.io/python-aos-lesson/) (PyAOS) lessons.

When I first wrote the PyAOS lessons a few years ago, I borrowed heavily from the Software Carpentry lessons. My defensive programming lesson was almost a straight copy of the corresponding python-novice-inflammation lesson. However, last year I received feedback (https://github.com/carpentries-lab/python-aos-lesson/issues/36) similar to #622, #773 and #874, suggesting that in most cases raising exceptions is a better approach to using assertions. I therefore just went through the process of re-writing our defensive programming lesson to address that feedback:  

https://carpentries-lab.github.io/python-aos-lesson/08-defensive/index.html

Instead of teaching assertions and then test-driven development, we now teach the following (quoted directly from the lesson):

_The first step toward getting the right answers from our programs is to assume that mistakes will happen and to guard against them. This is called defensive programming, and there are a number of tools and approaches at our disposal for doing this. Broadly speaking, we can:_
- _raise and handle errors to check and respond to program inputs_
- _use assertions to make sure nothing crazy or unexpected has happened_
- _write unit tests to make sure each component of our program produces expected outputs_
- _use a logging framework to report on program activity_

_In this lesson, we’ll look at how error handling, assertions and logging can make the unit conversion in our program more reliable, and we’ll provide links to further information on unit testing._

If people think our new PyAOS defensive programming lesson looks like a good approach to addressing the concerns in #622, #773 and #874, I'd be happy to submit a PR that adapts the content in our PyAOS lesson to fit with the Software Carpentry defensive programming lesson?

alee:
👍 as a lurker from the gapminder lesson I think this is a solid contribution, but I would recommend including conditionals as well to check for invalid inputs as a first line of defense before exceptions. In grad school I was taught to avoid using exceptions for flow control, i.e., as guards to dictate which branch a piece of code should flow into, but this was primarily for performance reasons.

But I think it's a useful distinction to leave exceptions for exceptional cases (i.e., if the filesystem cannot be written to, the network is down or unreachable, etc) vs invalid user input or other recoverable errors.

 These issues are also well articulated in https://stackoverflow.com/a/1835844/93370

maxim-belkin:
Hey Damien!

[...]

PR #1085, opened 2024-07-31
Update 10-defensive.md

### LINKED ISSUE #1084: Using Assert as input validation is wrong and needs to be changed
### What is the problem?

The problem has been reported several times already, so I'm late to the party (here fore example https://github.com/swcarpentry/python-novice-inflammation/issues/979).

To recap, `assert` keyword is not to be used as proposed in the example of the episode of Defensive Programming. The reason has been pointed out several times with various degrees of detail, but it boils down to that is basically a debug and testing phase tool, that can be bypassed at runtime by the interpreter. Python documentation states that basically evaluates to:

```python
if __debug__:
    if not expression: raise AssertionError
```
The episode of Defensive Programming focuses mainly on input validation. Input validation needs to stop the program every time an incorrect input is passed. `if raise` blocks are the standard and canonical way of doing this. For an example see this function in the scipy repository:

https://github.com/scipy/scipy/blob/v1.14.0/scipy/optimize/_minimize.py#L53-L775

I do not propose to also add the idea of `exceptions` neither `try except` blocks, since a novice user will not need this, I simple propose to replace `assert` blocks with `if raise` blocks. I will provide a pull request of this in the coming days.

### Location of problem (optional)

https://github.com/swcarpentry/python-novice-inflammation/blob/main/episodes/10-defensive.md

dhvalden:
complete change of the lesson to cover exceptions and try except pattern. Closes #1084

_If this pull request addresses an open issue on the repository, please add 'Closes #NN' below, where NN is the issue number._

_Please briefly summarise the changes made in the pull request, and the reason(s) for making these changes._

Changes completely the lesson to cover input validation using exception and try except patterns. It addresses #1084 adn #979 

_If any relevant discussions have taken place elsewhere, please provide links to these._

#979 #622, #773 and #874,

<details>

For more guidance on how to contribute changes to a Carpentries project, please review [the Contributing Guide](CONTRIBUTING.md) and [Code of Conduct](https://docs.carpentries.org/topic_folders/policies/code-of-conduct.html).

Please keep in mind that lesson Maintainers are volunteers and it may be some time before they can respond to your contribution. Although not all contributions can be incorporated into the lesson materials, we appreciate your time and effort to improve the curriculum. If you have any questions about the lesson maintenance process or would like to volunteer your time as a contribution reviewer, please contact The Carpentries Team at team@carpentries.org.

</details>

### DIFF
--- a/episodes/10-defensive.md
+++ b/episodes/10-defensive.md
@@ -6,11 +6,9 @@ exercises: 10
 
 ::::::::::::::::::::::::::::::::::::::: objectives
 
-- Explain what an assertion is.
-- Add assertions that check the program's state is correct.
-- Correctly add precondition and postcondition assertions to functions.
-- Explain what test-driven development is, and use it when creating new functions.
-- Explain why variables should be initialized using actual data values rather than arbitrary constants.
+- Understand the importance of input validation.
+- Learn how to use exceptions and raise errors in Python.
+- Practice writing defensive code to validate inputs.
 
 ::::::::::::::::::::::::::::::::::::::::::::::::::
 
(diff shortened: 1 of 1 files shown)


martinosorb:
Answering as a member of the CAC, but not representing the rest of the CAC.

I can agree that one should not use assert as input validation (although not all examples of assertions in the code of this episode are input validation). However these changes are much, much wider than that, and I think they would all need to be discussed. Just a few examples:
- Do we really want to silently remove test-driven development? I think it's a nice concept to mention
- Do we really want to teach `try: / except:`? In my experience, beginners abuse it, making messy code. In ten years programming in Python, I can't think of a single case where I really needed a `try` statement. To be honest, I have a personal beef against `try` and tend to think that the code is not very well written as soon as I see one.
- Do we really not want to teach `assert` at all? Asserting things is good practice to catch bugs. This episode is called 'defensive programming'. Error handling is not what I call defensive programming. It's also important, but it's a different thing.

In summary: one thing is removing the use of assert as input validation, one thing is completely rewriting the episode.

quirksahern (commented):
As a member of the SWC LPGC (now incorporating the CAC) I broadly support the recommended changes to this episode.

The focus is largely on focusing on introducing the concept that things can go wrong, and we need to check for those errors. At this stage I think that is sufficient. We can still maintain some content on TDD but that isn't the focus of this lesson.

For input validation assertions aren't appropriate, and can cause issues later on. 
For these instances try ... except is the better option. 

It would still be could to include assert, I recommend the following:

1. an example of the assert syntax be included, then 
2. example of using assert to check input and outputs to variables. 
3. Then add a comment that this kind of thing is done as part of the Test Driven Development approach, link out to further information on TDD

I don't think that there needs to be a discussion on TDD in this episode as it is beyond what learners need to know at this point in time.

martinosorb:
The first part can be read as using assert for input validation, which we have all agreed is not right, but could also be interpreted as testing, if we assume the input comes from other parts of our code. However, this is not sufficiently clear, and I agree to change this part to raising errors, since `raise` is not taught anywhere else.

The main problem is that we're throwing away the baby with the bathwater, as this becomes an episode on input validation only, removing the rest. I'd say we could keep a part on assert, and on the benefits of adding little testing devices to our code, which was the main point of this episode.
We can also still briefly talk about TDD, but without all the current detail.

In summary, I would
- keep the current introduction that motivates testing and checks
- when talking about pre-conditions
   - introduce `raise` for input validation
   - add some of @dhvalden's new material motivating input testing.
   - the discussion on the difference between TypeError and ValueError should be moved to ep 9, where all the error types are discussed, but can be referred to in the example
   - alternatively, we can keep the current example, which both shows pre- and post-conditions, so I think is better suited
- when talking about post-conditions, introduce `assert` for quick testing and explain its benefits
- shorten the part on TDD
- introduce `try...except`, but with a real use case where there is no solution other than using it.

quirksahern:
@dhvalden the SWC Lesson Program Governance Committee is keen to resolve the issue with this episode.

Are you willing to take on board @martinosorb 's comments and submit a revised pull request.
For info both myself and Martino are members of the LPGC.

dhvalden:
Hi! 

Yes! I think I can address the comments and upload a revised version.

ineelhere:
Please refer to PR https://github.com/swcarpentry/python-novice-inflammation/pull/1096 for an added suggestion on this topic. If this is not a potential conflict, PR https://github.com/swcarpentry/python-novice-inflammation/pull/1096 can be merged to lesson material.

unode:
Thanks @ineelhere for the cross ref.

Just adding my two cents after skimming through the lesson proposed and the comments in this thread. Particularly in light of the recent experience teaching this chapter.

I agree with the sentiment above that the proposed changes shift the focus entirely to input validation. I believe it also opens up into aspects that are considered "philosophical" namely the [correct or incorrect use of `isinstance`](https://www.reddit.com/r/learnpython/comments/2rhehb/if_type_checking_with_isinstance_is_bad_what/) and when create a custom exception versus raising or handling a built-in one. In the proposed example, should we 1. assume that it is a number and handle the error when it isn't or 2. ask if it is a number of a certain type (which excludes all numbers of other types).

The other philosophical aspect is the overall approach of "asking for forgiveness" vs "asking for permission". The best example I have in mind is "can I create a file?, please create" versus "create a file, oh sorry didn't know I couldn't". Arguably the latter is the better implementation because permissions could still change between asking and doing (or other racy situations) which would still lead to failure. So from a defensive programming point of view I believe we should focus on handling and anticipating the failure scenarios rather than double check out input assumptions. Our code would still be good code (better code?) if it can handle unexpected inputs but protects the user from nonsensical outputs.

PR #1096, opened 2025-02-27
Fix logical error in scaling code

unode:
When teaching this section the result was unexpected.
I believe there is a bug in the logic of the code that went undetected so far.

Inverting the scaling factor "fixed the issue" and made the code easier to understand for learners.

### DIFF
--- a/episodes/10-defensive.md
+++ b/episodes/10-defensive.md
@@ -126,7 +126,7 @@ def normalize_rectangle(rect):
     dx = x1 - x0
     dy = y1 - y0
     if dx > dy:
-        scaled = dx / dy
+        scaled = dy / dx
         upper_x, upper_y = 1.0, scaled
     else:
         scaled = dx / dy


ineelhere (approved):
Thank you for the suggestion. Please join the conversation at https://github.com/swcarpentry/python-novice-inflammation/pull/1085 for more context on upcoming changes discussed on this same part of the lesson and better collaboration and contributions. If this is not a potential conflict to PR https://github.com/swcarpentry/python-novice-inflammation/pull/1085, PR #1096 can be merged to lesson material.

### Adding some more context for reference - 

[The issue occurs when the rectangle is wider than it is tall.](https://github.com/swcarpentry/python-novice-inflammation/blob/ebe1c6d0ccf40223751f548e1e20a7eb8ae1a5b9/episodes/10-defensive.md?plain=1#L129) This fix ensures the aspect ratio is preserved by scaling `dy` relative to `dx` when `dx > dy`. The original code mistakenly used `dx/dy`, which inverted the proportions. Now, the longest side is always 1.0, and the shorter side is correctly scaled as `dy/dx` when needed.

tmorrell:
This is an intended error in the code to trigger the assertion, which is explained in https://github.com/swcarpentry/python-novice-inflammation/blob/dcb33beeb1763c70507073f6fc0fa5ef1049144a/episodes/10-defensive.md?plain=1#L220. Changing this would break the rest of the lesson.

unode:
Closing this old PR as not to be fixed.

PR #1097, opened 2025-02-27
Add illustration to help understand the code that follows

unode:
When trying to explain the code flow of the rectangle normalization, in a course I quickly put together this illustration which helped visualizing the two execution paths of the code.

Hereby contributing it to the lesson.

### DIFF
--- a/episodes/10-defensive.md
+++ b/episodes/10-defensive.md
@@ -109,6 +109,9 @@ representing the lower left and upper right corners of the rectangle.
 In order to do some calculations,
 we need to normalize the rectangle so that the lower left corner is at the origin
 and the longest side is 1.0 units long.
+
+![](fig/rectangle_normalization.svg){alt='Rectangle normalization follows one of two paths depending on which side is longer.'}
+
 This function does that,
 but checks that its input is correctly formatted and that its result makes sense:
 
--- a/episodes/fig/rectangle_normalization.svg
+++ b/episodes/fig/rectangle_normalization.svg
@@ -1,0 +1,255 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   width="100mm"
+   height="100mm"
+   viewBox="0 0 100 100"
+   version="1.1"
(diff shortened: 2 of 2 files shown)


unode:
This helped also to identify the logical error reported in #1096

unode:
Checking if this contribution is still pending consideration or if it has been rejected.

noatgnu:
Hello @unode , sorry for taking so long to review this. I agree that this part needs an illustration to better demonstrate the function. However, since it is a normalizing function, I would suggest making the two different rectangles becoming the same type of rectangle after normalization. As current, it is slightly confusing to have two arrows pointing from the same rectangle into two different rectangles. Another suggestion is adding a coordinate system to each rectangle to make the x0, x1 and y0, and y1 clearer for the learners.

Thank you for your contributions.

unode:
Hi @noatgnu, the intention with the two rectangles is to give a visual indication which of the "sides" needs to be scaled based on the condition. Notice that the condition in the center of the rectangles is different between the two.

Either way, from the discussion in #1096 and the https://github.com/swcarpentry/python-novice-inflammation/pull/1096#pullrequestreview-2648891438 comment, it seems there are other issues with this section.

From personal experience delivering this lesson, the defensive programming section has frequently been found to be difficult for learners to understand. Part of it, I think, comes from the not so simple example and the somewhat unexpected intentional error left in the assertions section.
For the learner it is meant to be a ha-ha! moment but from my experience it doesn't work due to the complexity of the example.

Either way, the suggestion was merely to try to help with this section that has always been difficult to teach. I've been more successful skipping the "Assertions" section altogether and directly starting with "Test-Driven Development".

---

Going a little off-topic, [others](https://github.com/swcarpentry/python-novice-inflammation/issues/874) have also raised concerns due to the fact that `assert` statements can be disabled at runtime. The risks [are well understood](https://snyk.io/blog/the-dangers-of-assert-in-python/) and try/except or if/.../else are generally preferred, even if they are more verbose.
This is especially true for libraries where exceptions should be documented. Handling `AssertionError` is generally difficult from other code. You know an error happened but to tell why or to know if you can recover from it, you need to parse the assertion message. Fine for humans but non-trivial in code.

PR #1117, opened 2026-04-23
Make use of amax and amin more consistent

edbennett:
The change in #995 left some instances of `max` and `min` present, which might cause learner confusion. This PR changes all of these that I could find to consistently use `amax` and `amin`. This required redrawing one diagram in SVG, and I made some related changes to using the full words "maximum" or "maxima" rather than "max" to be maximally explicit.

That said, I'm not sure that this is the correct course of action. The original intent of #995 was to make the lesson more consistent with the Numpy documentation, which discouraged use of `min` and `max. However, this seems to have been slightly premature:

- Shortly after that PR was opened, Numpy switched the aliases such that `amin` and `amax` are now aliases for `max` and `min`, and the latter have the full documentation.
- The discussion around this change indicates that the majority of code using Numpy uses these forms
- There is currently no encouragement to use the `amin`, `amax` forms in the documentation.
- `numpy.min` and `numpy.max` are not deprecated nor scheduled to be removed, so `from numpy import *` will overwrite our builtin namespace regardless of whether we use the `amin`, `amax` forms or not.
- Part of the reasoning for the change was to make `min` and `max` work correctly with Dask, whereas `amin` and `amax` now will not (if I'm reading the discussion correctly)
- Other functions that will clash with the builtin namespace, such as `numpy.sum` (which we use later in the lesson) do not have non-clashing aliases.

Based on this, it might be better to revert the `numpy.amin` and `numpy.amax` to `numpy.min` and `numpy.max`. If that is the preferred option, I can close this and open a fresh PR that does this.


tmorrell:
I'd lean toward going back to max and min given what you've seen in numpy. Explaining amax vs max is always confusing for learners. 

In any event we need to be 100% consistent with the usage across the lesson, as I've gotten tripped up by the change in the past.

Shorla:
Hi @edbennett, Thank you for fixing the issue! I am going to close this PR without merging, then merge the PR #1119

PR #1118, opened 2026-04-28
Add illustration for visual intuition of indexing and slicing

unode:
This illustration has been used to help beginners understand why the last digit of a slice doesn't imply including another element.

The intuition makes a distinction between indexing, the number of the position, and slicing, the number of the gap between elements.

For step based slices, the alternative intuition is to create rows with the values and keeping only the first row.

@tobyhodges 

python_indexing_slicing

### DIFF
--- a/episodes/04-lists.md
+++ b/episodes/04-lists.md
@@ -534,6 +534,9 @@ counts + counts
 
 ::::::::::::::::::::::::::::::::::::::::::::::::::
 
+## Visual intuition for indexing versus slicing
+
+![Indexing and slicing with steps](fig/python_indexing_slicing.svg){alt="Visual illustration of indexing, slicing and slicing with a step"}
 
 
 :::::::::::::::::::::::::::::::::::::::: keypoints
--- a/episodes/fig/python_indexing_slicing.svg
+++ b/episodes/fig/python_indexing_slicing.svg
@@ -1,0 +1,1092 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   width="820"
+   height="620"
+   viewBox="0 0 820 620"
+   role="img"
+   font-family="ui-sans-serif, system-ui, sans-serif"
+   version="1.1"
+   id="svg67"
+   sodipodi:docname="python_indexing_slicing.svg"
+   inkscape:version="1.4.2 (ebf0e940d0, 2025-05-08)"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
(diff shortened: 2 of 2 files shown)


Shorla:
Hi @unode, Thank you for taking the time to add the illustrations. 
I'm looking to merge this pull request. Can I get another pair of eyes on this before merging? @ineelhere @noatgnu

noatgnu:
There is potential problem with rendering this on different theme settings like dark or light mode because I am not able to see the text in many of the boxes when i open the image on its own.

image

unode:
The illustration is an SVG with transparent background and has only been tested on light backgrounds.

Dark themes have not been tested. I wasn't aware that this is now part of the lesson pipeline.

It should be possible to extend the SVG [to include support for dark themes](https://getpublii.com/docs/prepare-svg-for-light-dark-mode.html) but perhaps the easiest would be to force a light background rather than having it see-through.

noatgnu:
It is not part of the pipeline for dark and light theme for any of the lesson I think. It is just that browser on a dark mode desktop device. This is just me opening the image independently in another tab on the browser and all the transparent text boxes become dark. I think all the text boxes should have solid color background to avoid this native browser background.

unode:
I can't replicate your experiences with dark mode settings.
On my system it looks fine:

screenshot_2026-06-22_20-20-14_596624359

The [raw SVG "page"](https://raw.githubusercontent.com/unode/python-novice-inflammation/73ef9359c4fccd6f7876691d78d830c0d7d34aa4/episodes/fig/python_indexing_slicing.svg) also looks fine to me regardless of which settings I change.

@noatgnu could there be something else causing the visual artifacts you showed in your screenshot above?
And did you open the SVG directly or rendered it with The Carpentries Workbench?

I'm on Firefox, Linux, and to me it looks like the screenshot above. Dark mode takes the soul out of it and makes it suffer from contrast issues but is still visually consistent.

noatgnu:
I am in windows and opening the image on the post on both chrome and firefox but not the raw svg. If I open the raw svg it works. However, if I open the image right in message of the pull request. github probably mess up svg somehow and the transparency make it dark. So I think it is a github issue. This is what I see under the message at the top without clicking into a separate tab. As raw svg it should display fine without any problem only the github asset conversion that is causing problem above.

image

unode:
Ah wait, the screenshot you pasted above is from the PNG that I included for preview.

Try this instead for the direct SVG rendering:

unode:
I see now here that the 2 in the first line has very little contrast. I'll update the image and darken the yellow slightly.

unode:
Small change for better contrast on the yellow numbers.

unode:
Looks a little better in dark mode as well

screenshot_2026-06-22_21-40-37_437462657

Shorla:
@unode This is so much better! I would merge this. Thank You for your contribution!

